PR #2, opened 2021-04-07
Added Pageant authentication during Job submission

gpeb2:
The code now tries to log into the requested server to submit a job using Pageant to perform the authentication so the user does not have to provide their password.

### DIFF
--- a/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
+++ b/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
@@ -1077,7 +1077,7 @@ public class JobSubmission{
 	 * @return
 	 */
 	public boolean hostAvailabilityCheck(String server, int port) {
-		boolean available = true;server = "login-cpu.hpc.cam.ac.uk";
+		boolean available = true;
 		try (final Socket dummy = new Socket(server, port)){
 		} catch (IOException | NullPointerException e) {
 			available = false;
--- a/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
+++ b/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
@@ -12,7 +12,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.apache.commons.io.FileUtils;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -20,10 +19,15 @@ import org.slf4j.LoggerFactory;
 import com.jcraft.jsch.Channel;
 import com.jcraft.jsch.ChannelExec;
 import com.jcraft.jsch.ChannelSftp;
+import com.jcraft.jsch.IdentityRepository;
 import com.jcraft.jsch.JSch;
 import com.jcraft.jsch.JSchException;
 import com.jcraft.jsch.Session;
 import com.jcraft.jsch.SftpException;
+import com.jcraft.jsch.agentproxy.AgentProxyException;
+import com.jcraft.jsch.agentproxy.Connector;
+import com.jcraft.jsch.agentproxy.RemoteIdentityRepository;
+import com.jcraft.jsch.agentproxy.connector.PageantConnector;
 
 import uk.ac.cam.cares.jps.base.slurm.job.configuration.SlurmJobProperty;
 
@@ -439,6 +443,17 @@ public class JobSubmission{
 				session = jsch.getSession(slurmJobProperty.getHpcServerLoginUserName(), getHpcAddress(), 22);
 				String pwd = slurmJobProperty.getHpcServerLoginUserPassword();
 				session.setPassword(pwd);
+
+				try {
+					Connector con = new PageantConnector();
+					IdentityRepository irepo = new RemoteIdentityRepository(con);
+					jsch.setIdentityRepository(irepo);
+					session.setConfig("PreferredAuthentications", "publickey,keyboard-interactive,password");
+				} catch (AgentProxyException e) {
+					session.setConfig("PreferredAuthentications", "password");
+					logger.info("Failed to load identities from Pageant", e);
+				}
+
 				session.setConfig("StrictHostKeyChecking", "no");
 				session.connect();
 				scheduledIteration = 0;
@@ -1062,14 +1077,9 @@ public class JobSubmission{
 	 * @return
 	 */
 	public boolean hostAvailabilityCheck(String server, int port) {
-		boolean available = true;
-		try {
-			(new Socket(server, port)).close();
-		} catch (UnknownHostException e) {
-			available = false;
-		} catch (IOException e) {
-			available = false;
-		} catch (NullPointerException e) {
+		boolean available = true;server = "login-cpu.hpc.cam.ac.uk";
+		try (final Socket dummy = new Socket(server, port)){
+		} catch (IOException | NullPointerException e) {
 			available = false;
 		}
 		return available;
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -121,6 +121,16 @@
 			<artifactId>jsch</artifactId>
 			<version>0.1.55</version>
 		</dependency>
+		<dependency>
+			<groupId>com.jcraft</groupId>
+			<artifactId>jsch.agentproxy.jsch</artifactId>
+			<version>0.0.9</version>
+		</dependency>
+		<dependency>
+			<groupId>com.jcraft</groupId>
+			<artifactId>jsch.agentproxy.pageant</artifactId>
+			<version>0.0.9</version>
+		</dependency>
 
 		<dependency>
 			<groupId>javax.servlet</groupId>


JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
```diff
@@ -439,6 +443,17 @@ public void monitorJobs() throws SlurmJobException{
 				session = jsch.getSession(slurmJobProperty.getHpcServerLoginUserName(), getHpcAddress(), 22);
 				String pwd = slurmJobProperty.getHpcServerLoginUserPassword();
 				session.setPassword(pwd);
+
+				try {
+					Connector con = new PageantConnector();
+					IdentityRepository irepo = new RemoteIdentityRepository(con);
+					jsch.setIdentityRepository(irepo);
+					session.setConfig("PreferredAuthentications", "publickey,keyboard-interactive,password");
```
achCARES, line 451: Hardcoded values need to be replaced with variables. In Java, a good practice is to store them in properties file.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
```diff
@@ -439,6 +443,17 @@ public void monitorJobs() throws SlurmJobException{
...
+				} catch (AgentProxyException e) {
+					session.setConfig("PreferredAuthentications", "password");
```
achCARES, line 453: Same here, wrt hardcoding. You can see hardcoded string repetition starting to occur pretty quickly.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
```diff
@@ -439,6 +443,17 @@ public void monitorJobs() throws SlurmJobException{
...
+					logger.info("Failed to load identities from Pageant", e);
```
achCARES, line 454: Avoiding excessive logging is one of the coding guidelines for JPS. https://trello.com/c/jIfJbIdc/760-create-a-list-of-peer-review-criteria-for-codes-before-merging-into-develop-branch (this will be available soon on github for the reference)

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
```diff
@@ -439,6 +443,17 @@ public void monitorJobs() throws SlurmJobException{
 				session = jsch.getSession(slurmJobProperty.getHpcServerLoginUserName(), getHpcAddress(), 22);
 				String pwd = slurmJobProperty.getHpcServerLoginUserPassword();
 				session.setPassword(pwd);
+
+				try {
```
achCARES, line 447: Logic for this block is quite unclear. It looks like restricting authentication to password based upon exception. What is the reason for that? Exceptions are usually thrown when something goes wrong. Maybe some comments would help.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
```diff
@@ -1063,13 +1078,8 @@ public void deleteFolderOrFile(String folderOrFileName) throws JSchException, Sf
...
 	public boolean hostAvailabilityCheck(String server, int port) {
 		boolean available = true;
-		try {
-			(new Socket(server, port)).close();
-		} catch (UnknownHostException e) {
-			available = false;
-		} catch (IOException e) {
-			available = false;
-		} catch (NullPointerException e) {
+		try (final Socket dummy = new Socket(server, port)){
```
achCARES, line 1081: I would suggest checking availability on something different than IO/NullPointer exceptions. It seems to be too generic and can easily  lead to false negatives.

achCARES (changes requested):
Please review my suggestions.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: Whilst I agree that hardcoding settings is bad in general I'm not sure making this setting available to users would be beneficial. The only advantage I could see for doing so would be to allow the user to force authentication using the password even when a valid public key is loading into Pageant. I will add some comments to make this clearer in the code.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: Similar to the previous comment, if connecting to Pagaent has failed then the only option left is to authenticate using the password as trying any other method would just fail. Therefore giving the user the option to modify this value seems pointless. I will add some comments to make this clearer in the code.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: I had not been made aware of this document so thank you for pointing it out to me. I will remove this logging statement.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: I'll add some comments.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: I was just trying to remove the code flow where an exception thrown during the creation of the Socket could lead to the Socket not being closed, by changing it to use try-with-resources. I was not trying to fix the existing logic of the function. Looking at the code in the Socket constructor, the possibility of an IOException must be handled although I can't see an obvious source of one other than the subclass UnknownHostException. Regarding the NullPointerException I agree that this is very general also again looking into the Socket constructor's code it doesn't seem very likely that one will be thrown although the person who originally wrote the hostAvailabilityCheck method might have encountered it in a way that made this appropriate, I don't know. I could also see that a possible IllegalArgumentException exception can be thrown if the port value is out of the valid range so adding that in would probably be more appropriate. I'll make some changes to the method for you to check.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
achCARES: I haven't mentioned exposing the functionality to users. Just not hardcoding strings.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
achCARES: Again, I haven't mentioned anything about user previously.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
achCARES: You're welcome.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: I've added a further commit that I think answers this comment, please review it.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: Sorry, when I say users, I'm thinking of the developers using this code. Hopefully the code comments I've suggested in GitHub comment below make it clearer why those strings are hardcoded and not specified in a properties file.

achCARES: If you insist of not using properties, please at least use final class member variables for them (aka constants). You'll be able to reuse them in the code. Otherwise you're repeating the hardcoded strings.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
```diff
@@ -439,6 +443,17 @@ public void monitorJobs() throws SlurmJobException{
...
+
+				try {
+					Connector con = new PageantConnector();
+					IdentityRepository irepo = new RemoteIdentityRepository(con);
+					jsch.setIdentityRepository(irepo);
+					session.setConfig("PreferredAuthentications", "publickey,keyboard-interactive,password");
+				} catch (AgentProxyException e) {
+					session.setConfig("PreferredAuthentications", "password");
+					logger.info("Failed to load identities from Pageant", e);
+				}
```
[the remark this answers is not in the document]
achCARES: Is there any other way of checking whether there is a valid key present rather than via catching AgentProxyException? Here I am also a little bit concerned about a case where an exception is thrown for a different reason than invalid key potentially. Those are security issues and it is good to be cautious with such things.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
achCARES: Please also .close() the socket.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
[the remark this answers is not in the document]
achCARES: Also, please confirm that this code is not Windows dependent and will work on *nix like operating systems. Pageant uses win32 APIs, as far as I am aware.

achCARES (changes requested):

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: Whilst I agree that in general string constants should be made static final member variables I think that in this instance using the raw strings is clearer and will make it easier to add new authentication types in the future. These strings are not intended for use outside of this method and are very context specific. I also believe that making them variables will lead to potential confusion as the names of those variables would be intrinsically tied to their content and if someone changed the content (for example by adding a new authentication option) but did not change the name of the variable it might become misleading.
There would be a minor benefit to creating constants for each of the options in that it would reduce the probability of mistyping the strings but I think the benefits would be greatly outweighed by having to write code like:
`PUBLICKEY + ',' + KEYBOARD_INTERACTIVE + ',' + PASSWORD`
or
`Stream.of(PUBLICKEY, KEYBOARD_INTERACTIVE, PASSWORD).collect(Collectors.joining(",")`
especially considering the small likelihood of these particular strings being reused.
Finally, in all of the online examples I have seen that discuss this setting they all write the string argument in the same form as I have and so I feel that keeping it like this will make it easier for any future developers to compare the current settings to those in the examples they see.
I accept that my code is not always correct, let alone perfect, and appreciate that you are only trying to improve the general standard of the World Avatar code, which I fully support and am more than happy to work with you on, but blindly applying simple rules is not the way to improve code quality.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
gpeb2: The whole reason I changed this method in the first place was to remove the need to explicitly call the "close" method.
Support for "try-with-resources" was added in Java 7 to reduce mistakes with regards to closing things like files and sockets. Please read this article about using "try-with-resources" statements https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html .
I would strongly recommend that some mention of using "try-with-resources" is added to the coding standards document as it should be used wherever possible.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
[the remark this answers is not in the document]
gpeb2: The initial call the the PageantConnector constructor will fail quickly if the native Pageant runtime libraries are not found so will default to the password only option if Pageant is not installed/supported on the current system.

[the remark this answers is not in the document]
gpeb2: Regarding your first question, this block of code doesn't actually check any keys, it just tries to connect to a running Pagaent instance and, if successful, registers it with the JSch object. The only exception that can be thrown by any of these methods is AgentProxyException and that is only thrown if the code is unable to load one of the native libraries it requires.

achCARES (commented):
Please find some other person to do this review for you. I have a lot of other workload at the moment and do not even work as a developer at CARES.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
achCARES: Using constants makes it easier to modify code. DRY also affects compilation in Java. Stay consistent and use finals please.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/slurm/job/JobSubmission.java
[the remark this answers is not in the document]
achCARES: You either design it to work differently on different operating systems and make this explicit in your code or find a way to make the functionality being the same on different systems. I'd prefer the second option.

PR #3, opened 2021-04-28
123 ontology for provenance information

jb2197:
This is the initial attempt of creating OntoProvenance - an ontology for provenance information. Please review if anything needs to be revised.


sm453:
Several sentences in the explanations contain a phrase "is used to describe the situation where the original source published its information in a tabular/textual way that the agent responsible for deriving". These are not correct sentences, and I don't find them comprehensible. Please rewrite.

jb2197:
@sm453 Please have another look and I will merge develop to this branch and merge it back to develop if everything is okay. Thanks in advance.

PR #6, opened 2021-06-23
130 create ontomessage ontology tbox

jb2197:


sm453:
Please regenerate the OWL.

PR #13, opened 2021-07-16
(1076-dev-add-unit-test-for-querybuilder)

yjytt59:
1. add QueryBuilderTest
2. add 4 tests whether output match to the SPARQL query

### DIFF
--- a/JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/query/sparql/test/QueryBuilderTest.java
+++ b/JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/query/sparql/test/QueryBuilderTest.java
@@ -1,0 +1,39 @@
+package uk.ac.cam.cares.jps.base.query.sparql.test;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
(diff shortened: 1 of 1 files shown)


JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/query/sparql/test/QueryBuilderTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import uk.ac.cam.cares.jps.base.query.sparql.QueryBuilder;
+
+import static org.junit.Assert.*;
+
+public class QueryBuilderTest {
+    private QueryBuilder qbuilder = new QueryBuilder();
+
+    @Test @Before
```
gpeb2, line 13: Hi, I don't think you want to have the @Before annotations here as they indicate that the function should be run before each test (as described in this article https://www.softwaretestinghelp.com/junit-annotations-tutorial/).
The result is that each "test" function is being run multiple times, like this:
testSelect
testA
testProp
testBuild

testSelect
testA
testProp
testProp

testSelect
testA
testProp
testA

testSelect
testA
testProp
testSelect

PR #22, opened 2021-07-28
Merge branch 1135-dev-quick-fix-of-keyvaluemap-test into develop

nkasenburg:
Fixes the test for KeyValueMap that originally cleaned the map of the singleton resulting in other test:
1. Refactor all existing tests to use JUnit4/5 syntax.
2. Refactor all tests to not clear the map.
3. Add Assert.fail() assertion for all tests that should result in an exception.
4. Refactor the KeyValueMap class to be thread consistent when returning the instance.
5. Refactor the KeyValueMap class to use try with resource when an input stream is used to ensure proper closure of the stream.

### DIFF
--- a/JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/config/test/KeyValueMapTest.java
+++ b/JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/config/test/KeyValueMapTest.java
@@ -1,17 +1,15 @@
 package uk.ac.cam.cares.jps.base.config.test;
 
-import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.net.URISyntaxException;
 
-import junit.framework.TestCase;
 import org.junit.Assert;
 import org.junit.Test;
 import uk.ac.cam.cares.jps.base.config.KeyValueMap;
 import uk.ac.cam.cares.jps.base.exception.JPSRuntimeException;
 
 public class KeyValueMapTest {
+
 
 	@Test
 	public void testGetInstance() {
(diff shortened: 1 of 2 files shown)


JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/config/test/KeyValueMapTest.java
```diff
@@ -1,250 +1,220 @@
...
 		kvm.put(key, val);
-		assertEquals(kvm.get(key), val);
+		Assert.assertEquals(kvm.get(key), val);
 
+		// Accessing a key with value null should result in an exception
 		try {
-			key = null;
-			kvm.get(key);
+			kvm.get(null);
+			Assert.fail();
```
achCARES, line 41: Assert.fail() seems to be redundant here because it will not be asserted when the previous like causes exception which is caught after that and the exception message is asserted instead.

achCARES (changes requested):
please see comment about the Assert.fail()

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/config/test/KeyValueMapTest.java
nkasenburg: The idea is that an exception is expected. Without the Assert.fail(), if no exception is thrown the test will pass, but it should not because we expect an exception. The catch is then to further assert the type of exception and optionally the message.

achCARES: OK, noted. It is like a "belt and suspenders" approach. ✅

achCARES (approved):

PR #25, opened 2021-07-30
1133 dev overarching timeseries client

markushofmeister:
All unit and integration tests included and methods, which modify underlying time series data have their public modifier removed to channel all usage through overaching time series client


nkasenburg (approved):
All  files reviewed and where necessary changed with new commits (see commit messages for details).

gpeb2 (commented):
I've made a few comments relating to issues with exception handling that already existed before you created your branch. I don't think fixing them is urgent but they should be fixed at some point.

PR #26, opened 2021-08-02
1133 dev overarching timeseries client

markushofmeister:
Finalised TimeSeriesClient class, which now acts as only contact point to users and agents who want to interact with time series in the knowledge graph
- declared all methods in Sparql and RDB client classes, which could alter data in KG or RDB, as protected to hide from the user
- extended functionality of TimeSeriesClient to also offer pure KG-related methods (to emphasize this class as single point of contact)
- updated several issues with regards to exception handling


JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesRDBClient.java
```diff
@@ -95,63 +95,38 @@ public void setRdbPassword(String password) {
...
+        	setRdbUser(prop.getProperty("db.user"));
+        } else {
+        	throw new JPSRuntimeException(exceptionPrefix + "Properties file is missing \"db.user=<rdb_username>\" ");
+        }
+        if (prop.containsKey("db.password")) {
+        	setRdbPassword(prop.getProperty("db.password"));
+        } else {
+        	throw new JPSRuntimeException(exceptionPrefix + "Properties file is missing \"db.password=<rdb_password>\" ");
+        }
+
```
gpeb2, line 129: The `input` file stream should be created in a try-with-resources block so that the `close` call is always made correctly even if an exception is thrown.

gpeb2 (changes requested):
Please modify the highlighted code and any similar blocks to use try-with-resources as it is the most robust way to handle closing things like file streams. If you have any questions just let me know.

kokfoongcmcl (changes requested):
Add a constructor with only the timeClass as its argument

public TimeSeriesClient(Class<T> timeClass)  {
    	this.rdbClient = new TimeSeriesRDBClient<>(timeClass);
 }

nkasenburg (approved):
No further comments. Changes discussed in Skype meeting.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesRDBClient.java
markushofmeister: Updated in commit 9f174e2cff729807ad6f6916d8a72bd2a3f58a42

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesRDBClient.java
```diff
@@ -101,31 +101,32 @@ protected void loadRdbConfigs(String filepath) throws IOException {
...
+	        	throw new JPSRuntimeException(exceptionPrefix + "Properties file is missing \"db.user=<rdb_username>\" ");
+	        }
+	        if (prop.containsKey("db.password")) {
+	        	setRdbPassword(prop.getProperty("db.password"));
+	        } else {
+	        	throw new JPSRuntimeException(exceptionPrefix + "Properties file is missing \"db.password=<rdb_password>\" ");
+	        }
+		} catch (JPSRuntimeException e) {
+			throw e;
+		}
```
gpeb2: You don't need to include the `catch` block here as any `JPSRuntimeException` (and any other type of exception) will just be passed up to the calling method by default if it isn't caught. The try-with-resources form effectively adds an implicit `finally` block that satisfies the requirement that something must happen if an exception is thrown in the `try` block.

gpeb2 (changes requested):

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesRDBClient.java
markushofmeister: Unnecessary catch blocks have been removed with commit 5d46a6e03662bb9f552db77dc6881d7270868f76

PR #35, opened 2021-08-18
Dev derivedquantities example

kokfoongcmcl:
Hello, the base classes for derivation are located [here](https://github.com/cambridge-cares/TheWorldAvatar/tree/dev-derivedquantities-example/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation).

The example is located [here](https://github.com/cambridge-cares/TheWorldAvatar/tree/dev-derivedquantities-example/Agents/DerivationExample).

If you do not have time to go through the code, just review the README. Both the base class and the example have their own documentation.


nkasenburg (approved):
I ran all the unit tests and followed the example (including starting starting the Docker container and running the tests). Everything worked without problems!

markushofmeister (commented):
**Base classes**: 
- I was able to run all the tests of the base classes
- I reviewed the readme document and it is all well described and documented
- two small comments:
          1. “It is assumed that each agent has a single URL/operation” – Why is that necessary 
              and why do we then need both <Derivation> <:isDerivedUsing> <Agent> and 
              <Agent> <msm:hasOperation> <Operation> if only one operation is provided per 
              agent?
          2. “Derivation with time series” – It could be made clearer that time series in that case 
              refers to the outputs and not inputs; btw: Do we currently keep track of the provided 
              input arguments to the Agent when creating the derivation?

**Derivation Example**:
I was able to spin up all required docker containers and build the example project; however, I was not able to run the integration tests as Eclipse claims that there are no test runners defined for that project (applies to all JUnit versions 3-5)

kokfoongcmcl:
> **Base classes**:
> 
> * I was able to run all the tests of the base classes
> * I reviewed the readme document and it is all well described and documented
> * two small comments:
>   1. “It is assumed that each agent has a single URL/operation” – Why is that necessary
>   and why do we then need both  <:isDerivedUsing>  and
>    msm:hasOperation  if only one operation is provided per
>   agent?
>   2. “Derivation with time series” – It could be made clearer that time series in that case
>   refers to the outputs and not inputs; btw: Do we currently keep track of the provided
>   input arguments to the Agent when creating the derivation?
> 
> **Derivation Example**:
> I was able to spin up all required docker containers and build the example project; however, I was not able to run the integration tests as Eclipse claims that there are no test runners defined for that project (applies to all JUnit versions 3-5)

Regarding point 1:
I can definitely simplify it by just using the agent URL, i.e. <derivation> <isDerivedUsing> <agentURL>. I started off with the msm ontology where an agent is a "Service" and each service can contain multiple operations with different URLs. It is not clear to me whether we are going to use the msm ontology extensively in the future.

Point 2:
At the point of creating the derivations there are logs written. Other than that, every time the derivation is updated, the inputs are queried using the <isDerivedFrom> property. Also the input IRI may not stay the same if they are part of another derivation

Unit test:
You might have to configure your build path for a new project? Let me know if you are still having issues.

sm453:
>   1. “It is assumed that each agent has a single URL/operation” – Why is that necessary
>   and why do we then need both  <:isDerivedUsing>  and
>    msm:hasOperation  if only one operation is provided per
>   agent?

The purpose of hasOperation is to mark up a "function" of an agent (or multiple functions, if it can do more than one thing). This has nothing to do with derived information. The purpose of isDerivedUsing is to state that a piece of information was created using a particular agent. That is, however, only sufficient if the agent can only do one thing. If the agent has multiple hasOperation, then one cannot know (in general, without additional information) which function was used to carry out the derivation. This could be addressed, of course, but we chose not to in the first instance, to keep it simple.

>   2. “Derivation with time series” – It could be made clearer that time series in that case
>   refers to the outputs and not inputs; btw: Do we currently keep track of the provided
>   input arguments to the Agent when creating the derivation?

That is the purpose of isDerivedFrom statements. (Maybe I'm misunderstanding the question...)

sm453 (approved):
Looks ok to me.

markushofmeister:
> > 1. “Derivation with time series” – It could be made clearer that time series in that case
> >    refers to the outputs and not inputs; btw: Do we currently keep track of the provided
> >    input arguments to the Agent when creating the derivation?

> That is the purpose of isDerivedFrom statements. (Maybe I'm misunderstanding the question...)

I had in mind cases where the Derivation is, for example, derived from a particular part of a time series as input. In those cases it would be necessary to know what part of the time series history (e.g. last month, last year, full length, ...) has been used to derive the respective instance; however, in those cases one might "simply" have to define two input IRIs: the time series itself and the duration used as second input not to lose this information

PR #36, opened 2021-08-25
1160 dev aqmesh input agent

nkasenburg:
This PR adds the AQMesh input agent to The World Avatar. The AQMesh input agent, when built, is an executable jar (also available as docker image), which purpose it is to retrieve and store data from the AQMesh API.

Please start with reading the README and try to do a test run following the descriptions. Write a comment if something is unclear or things are not working.

One known caveat is that ideally the time-series measures (of which there are around 95) are attached to existing instances (IRIs) in the knowledge graph. However, these do not exist for an AQMesh representation as of now. A future addition to the module could be the functionality to automatically instantiate an AQMesh.


kokfoongcmcl (approved):
I've tested it and it works fine on my machine

PR #41, opened 2021-09-07
dev-ts-tweak: Make domain of hasTimeSeries generic and changed the ti…

kokfoongcmcl:
Changed namespace and made hasTimeSeries generic


markushofmeister (approved):
Suggested namespace change makes perfect sense.

PR #42, opened 2021-09-10
dev-python-agent-fix

mdhillmancmcl:
Very small change to the example Python agent's Dockerfile, adding a missing line to install git.


mdhillmancmcl:
As this change is small, not to source code itself, and has been tested locally, I'm going to approve it myself.

PR #45, opened 2021-09-14
Dev marie remove third party search

oparry-cmcl:
Removes Google and Wolfram results from the 'Marie' page of the website and the funtionality to run the searches in the Marie service itself.


oparry-cmcl:
Self-approving this PR, as the changes don't introduce any new code and have been tested locally.

PR #48, opened 2021-09-17
1199 dev pass namespace or iri to storerouter

csl37:
Changes to the StoreRouter to accept a target resource ID, which can be either an iri or just the namespace/filepath.
Only the namespace is matched against a label in the ontokgrouter triple store, so a full iri is not needed.


achCARES (approved):
✅

msff2 (commented):
We need to ensure that only HTTP or HTTPS IRIs (e.g. http://theworldavatar.com/kb/ontokin) and namespaces (e.g. ontokin) are accepted as IDs for repositories/namespaces. It is currently wide open; for example, strings including "test/ontokin" and "kb/ontokin" are considered valid inputs.

PR #49, opened 2021-09-17
dev-docker_stack_improvements

oparry-cmcl:
This branch includes a number of changes to the configuration files and docker-compose wrapper scripts which we use to deploy container stacks locally and on the servers.


mdhillmancmcl:
@oparry-cmcl has updated me on his changes to the build/deployment scripts, I have no issues with the new approach.

PR #50, opened 2021-09-20
135 dev weather input

kokfoongcmcl:
It will be sufficient to just confirm that you can build it (as we are both busy with the COP26)

You'll need to change the jps-base-lib-version to use 1.4.0-dev-weather-input-SNAPSHOT


markushofmeister:
I can successfully build and test the WeatherAgent project from within Eclipse after setting <jps.base.version>1.4.0-dev-weather-input-SNAPSHOT</jps.base.version> in the pom.xml

There are a couple of SLF4J and log4j warnings.

markushofmeister (approved):
I can successfully build and test the WeatherAgent project from within Eclipse after setting <jps.base.version>1.4.0-dev-weather-input-SNAPSHOT</jps.base.version> in the pom.xml

There are a couple of SLF4J and log4j warnings.

markushofmeister (approved):

PR #55, opened 2021-09-28
dev-status-agent

mdhillmancmcl:
New agent that regularly runs simple tests to check that KG endpoints are accessible.


mdhillmancmcl:
Status agent currently contains minimal functionality, but has been tested (manually and via automated GitHub actions).

PR #62, opened 2021-10-21
Added README.md to time series client

markushofmeister:


markushofmeister:
Added some information about how to access agent codes even if links break in commit 2184df8ff53f57b089d313f8503f573b25d30bab

PR #68, opened 2021-10-27
135 dev expand derivation framework

jb2197:
Expanded the derivation framework to support asynchronous operation mode. A working example is provided in branch 133-dev-design-of-experiment.


kokfoongcmcl (commented):
I've skimmed through most of the codes. I am hoping that there is some sort of skeleton agent code, e.g. a class called AsyncAgent, that contains all the behaviour expected of the agent monitoring the derivations (probably next to JPSAgent). Then developers can extend the class and add functions to it.

At the moment, it seems to me that everytime we create a new Asynchronous agent, people will have to manually code the behaviour, e.g. monitoring the statuses. 

The final approval is up to Seb.

sm453:
I agree with @kokfoongcmcl. @jb2197, as discussed, please implement an agent class that contains all the monitoring code, so that when people create new (asynchronous) agents they don't need to repeat all the monitoring code.

sm453:
Looks ok to me. Unless @kokfoongcmcl spots any show-stoppers...

sm453 (approved):

kokfoongcmcl (approved):

PR #70, opened 2021-10-29
Fixed issues with running testcontainers

markushofmeister:
- Updated the pom.xml of the JPS_BASE_LIB to include two additional dependencies needed by testcontainers (jackson-annotation, slf4j) and updated version numbers. All integration test outputs will go to console and not our regular log4j logging output file. 
- Replaced all inefficient context.meta() queries from RDBClient with more efficient ones and re-introduced some time series instantiation checks to allow for more descriptive exception messages. All re-introduced checks are performed with RDBClient due to performance reasons

### DIFF
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -341,7 +341,8 @@
 			<artifactId>junit</artifactId>
 			<version>4.13.2</version>
 		</dependency>
-        <!-- ??? -->
+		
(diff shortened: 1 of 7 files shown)


kokfoongcmcl (approved):
tested with the flood input agent and performance is unaffected

kokfoongcmcl:
might also be a good opportunity to remove the @Ignore tags

markushofmeister:
> might also be a good opportunity to remove the @ignore tags

That's a very good idea and will include these changes in a further commit before merging

PR #71, opened 2021-11-03
dev-floodagent-output: write outputs into folder structure required b…

kokfoongcmcl:
FYI, modified scripts to write files into the structure required by the visualisation framework


mdhillmancmcl:
I've not run @kokfoongcmcl's code (I do not have the time). I have seen the results and, looking at the code, all seems well.

PR #78, opened 2021-11-16
Minimum working example of how to use geospatial time series

markushofmeister:
Please test whether you can instantiate, retrieve, and visualise geospatial time series data using the provided instructions


kokfoongcmcl:
I personally prefer the API key to be written by the script, specified in the properties file. Also missing instruction to spin up Blazegraph with geospatial enabled, I know how to initialise a new namespace programatically, let me know if you wish to do that.

kokfoongcmcl:
better to provide the command to run the python scripts explicitly, e.g. <venv>\Scripts\python.exe instantiate_data
It's not super clear that the docker compose script will spin up a container named "docker", I had to work a bit to find the "Open in Browser" option

markushofmeister:
> better to provide the command to run the python scripts explicitly, e.g. \Scripts\python.exe instantiate_data. It's not super clear that the docker compose script will spin up a container named "docker", I had to work a bit to find the "Open in Browser" option

Included suggested changes in commit 3e7eaec17c2537fa70788eade520d15e0c77f3c1

PR #80, opened 2021-11-22
Dev timeseries json

kokfoongcmcl:
added two keys in the time series JSON object to specify the class of the time and values column. If it's an instance of "Number" (e.g. integers, doubles), it will output "Number", otherwise it will write whatever the original value's class is.

Michael will support a selection of classes and if his code detects any other classes, it should throw an error. For the time column, I assume the code will support either "Instant" or "Number". The list of supported classes will probably grow in the future.

@markushofmeister this is mostly FYI for you


markushofmeister:
> added two keys in the time series JSON object to specify the class of the time and values column. If it's an instance of "Number" (e.g. integers, doubles), it will output "Number", otherwise it will write whatever the original value's class is.
> 
> Michael will support a selection of classes and if his code detects any other classes, it should throw an error. For the time column, I assume the code will support either "Instant" or "Number". The list of supported classes will probably grow in the future.
> 
> @markushofmeister this is mostly FYI for you

I understand the approach; however, it is not fully clear to me how this information will be used by Michael's code (I assume that's the DTVF) and how we make sure that this implementation is "in sync" with the requirements of the updated DTVF?

PR #85, opened 2021-12-01
Dev update derivations

kokfoongcmcl:


kokfoongcmcl:
@jb2197 
After I merge these changes, please remove all the little queries, e.g. isRequested, isInProgress, ... etc. , and replace it with something like getDerivations that queries all the information in 1 go, and use/expand the Derivation and Entity class to cache the information in memory

PR #88, opened 2021-12-02
Dev add time series data

kokfoongcmcl:
addTimeSeriesData will now always look for an existing row with the equivalent time value before inserting a new row. If it exists, it will update the row.

@markushofmeister regarding the integration tests that I ignored, try not to stop the container after every single test when you have time to improve them, use the same container instead for a class


markushofmeister:
> addTimeSeriesData will now always look for an existing row with the equivalent time value before inserting a new row. If it exists, it will update the row.
> 
> @markushofmeister regarding the integration tests that I ignored, try not to stop the container after every single test when you have time to improve them, use the same container instead for a class

I disagree with disabling the integration tests. Some of the tests require clean slates of the RDB or Blazegraph, or require the container to actually shut down. Hence, frequent restarting of the containers is actually required for the tests. I just tried to build the JPS_BASE_LIB 5 times and all worked well with the integration tests.

PR #91, opened 2021-12-06
140 dev chain of asyn derivation

jb2197:
This PR further expends the asynchronous part of the derivation framework by enabling a cascade of asynchronous derivations, a minimum working example is provided in https://github.com/cambridge-cares/TheWorldAvatar/pull/92.


kokfoongcmcl:
I can see that you are still keeping the little queries, i.e.
line 591: List<String> inputsAndDerived = this.sparqlClient.getInputsAndDerived(instance);
line 621: List<String> inputs = this.sparqlClient.getInputs(instance);
line 624: if (isDerivedAsynchronous(instance))
line 626: if (isOutOfDate(instance, inputs)) 

I think hasStatus and markAsPendingUpdate can be combined into a single query. The rule of thumb is to make as little connection to the triple-store as possible.

Is it not possible to query all these in a single query before you start traversing the graph?

Also you need to remember to revert the base lib version number.

kokfoongcmcl:
The same applies to checkAtPendingUpdate, I believe you can query all the information you need in a single query, including the inputs, status and timestamps. Looping through each derivation and querying the information for each derivation separately in a loop will slow down really quickly as you have more derivations.

In cleanUpFinishedDerivationUpdate, I think it is possible to replace getNewDerivedIRI and getDerivedEntitiesAndUpstreamDerivation with a single query.

jb2197:
> I can see that you are still keeping the little queries, i.e. line 591: List inputsAndDerived = this.sparqlClient.getInputsAndDerived(instance); line 621: List inputs = this.sparqlClient.getInputs(instance); line 624: if (isDerivedAsynchronous(instance)) line 626: if (isOutOfDate(instance, inputs))
> 
> I think hasStatus and markAsPendingUpdate can be combined into a single query. The rule of thumb is to make as little connection to the triple-store as possible.
> 
> Is it not possible to query all these in a single query before you start traversing the graph?
> 
> Also you need to remember to revert the base lib version number.

The small queries got kept as we need to retrieve the most up-to-date information from the knowledge graph under the asynchronous operation mode - this is particularly important as there are more than one agent that is able to modify the status of the derivation: the agent who requested the update (updateDerivationAsyn), and the agent who monitors the derivation (monitorDerivation). Caching the whole graph before starting traversing has the risk of making decisions based on outdated information, i.e. the agent monitoring the derivation updated the knowledge graph right after the caching and before the traversing start. Also, the performance issue is less of a concern in asynchronous operation mode as the updateDerivationAsyn method is only marking status but not firing HTTP requests - the job is done automatically by the agent who monitors the derivation. 

hasStatus is kept separate from markAsPendingUpdate as hasStatus is also used within deleteStatus method.

Thanks for the reminder about the base lib version number, I will change that once the PR is approved and before merge it back to develop.

jb2197:
> The same applies to checkAtPendingUpdate, I believe you can query all the information you need in a single query, including the inputs, status and timestamps. Looping through each derivation and querying the information for each derivation separately in a loop will slow down really quickly as you have more derivations.
> 
> In cleanUpFinishedDerivationUpdate, I think it is possible to replace getNewDerivedIRI and getDerivedEntitiesAndUpstreamDerivation with a single query.

The small queries in checkAtPendingUpdate were kept due to the same reason - to ensure the decision is made based on the most updated information in the knowledge graph. As the asynchronous operation mode is designed for long jobs, the number of derivations in the chain is not considered to be a lot in this round of development. Also for long jobs, the time spent querying the knowledge graph is less of an issue - the time interval between issuing an update request and completing the job is long. However, the job can be completed at any time, so it's important to keep the latest information while making decisions. 

You are right, I agree it's possible to combine the getNewDerivedIRI and getDerivedEntitiesAndUpstreamDerivation into one single query. Only one agent has access to status IRI at this point in the lifecycle, so storing some information locally is helpful in reducing the operating time. In fact, this is already done partially as getDerivedEntitiesAndUpstreamDerivation stores its query results in a List of Entity instances.

kokfoongcmcl:
> > I can see that you are still keeping the little queries, i.e. line 591: List inputsAndDerived = this.sparqlClient.getInputsAndDerived(instance); line 621: List inputs = this.sparqlClient.getInputs(instance); line 624: if (isDerivedAsynchronous(instance)) line 626: if (isOutOfDate(instance, inputs))
> > I think hasStatus and markAsPendingUpdate can be combined into a single query. The rule of thumb is to make as little connection to the triple-store as possible.
> > Is it not possible to query all these in a single query before you start traversing the graph?
> > Also you need to remember to revert the base lib version number.
> 
> The small queries got kept as we need to retrieve the most up-to-date information from the knowledge graph under the asynchronous operation mode - this is particularly important as there are more than one agent that is able to modify the status of the derivation: the agent who requested the update (updateDerivationAsyn), and the agent who monitors the derivation (monitorDerivation). Caching the whole graph before starting traversing has the risk of making decisions based on outdated information, i.e. the agent monitoring the derivation updated the knowledge graph right after the caching and before the traversing start. Also, the performance issue is less of a concern in asynchronous operation mode as the updateDerivationAsyn method is only marking status but not firing HTTP requests - the job is done automatically by the agent who monitors the derivation.
> 
> hasStatus is kept separate from markAsPendingUpdate as hasStatus is also used within deleteStatus method.
> 
> Thanks for the reminder about the base lib version number, I will change that once the PR is approved and before merge it back to develop.

It really depends on which process is quicker (monitoring / updating). Is the monitoring agent so fast that the graph gets updated halfway through a update process? Traversing the graph in memory is super quick (after the query), I'd just execute the update process again if we are worried that something is out-of-date. In our experience firing the HTTP requests is not the bit that is slowing down, it is the bit that queries for the timestamps and inputs each time we traverse down the graph.

jb2197:
> > > I can see that you are still keeping the little queries, i.e. line 591: List inputsAndDerived = this.sparqlClient.getInputsAndDerived(instance); line 621: List inputs = this.sparqlClient.getInputs(instance); line 624: if (isDerivedAsynchronous(instance)) line 626: if (isOutOfDate(instance, inputs))
> > > I think hasStatus and markAsPendingUpdate can be combined into a single query. The rule of thumb is to make as little connection to the triple-store as possible.
> > > Is it not possible to query all these in a single query before you start traversing the graph?
> > > Also you need to remember to revert the base lib version number.
> > 
> > 
> > The small queries got kept as we need to retrieve the most up-to-date information from the knowledge graph under the asynchronous operation mode - this is particularly important as there are more than one agent that is able to modify the status of the derivation: the agent who requested the update (updateDerivationAsyn), and the agent who monitors the derivation (monitorDerivation). Caching the whole graph before starting traversing has the risk of making decisions based on outdated information, i.e. the agent monitoring the derivation updated the knowledge graph right after the caching and before the traversing start. Also, the performance issue is less of a concern in asynchronous operation mode as the updateDerivationAsyn method is only marking status but not firing HTTP requests - the job is done automatically by the agent who monitors the derivation.
> > hasStatus is kept separate from markAsPendingUpdate as hasStatus is also used within deleteStatus method.
> > Thanks for the reminder about the base lib version number, I will change that once the PR is approved and before merge it back to develop.
> 
> It really depends on which process is quicker (monitoring / updating). Is the monitoring agent so fast that the graph gets updated halfway through a update process? Traversing the graph in memory is super quick (after the query), I'd just execute the update process again if we are worried that something is out-of-date. In our experience firing the HTTP requests is not the bit that is slowing down, it is the bit that queries for the timestamps and inputs each time we traverse down the graph.

The monitorDerivation process and subsequently executing a job is (much) slower by design, but the thing is that the job can be done at any point in time. As it's not designed for the agent requesting the update (updateDerivationAsyn) to get the information it accessed immediately, the same agent will need to run updateDerivationAsyn multiple times (maybe periodically) before the information it requested is updated and can be used for other purposes. So it's possible that the knowledge graph got updated during the agent running the updateDerivationAsyn - if we cache the information in one query, there is a risk of the agent messing up the status of the derivation with outdated information, which will be difficult to resolve.

sm453:
In DerivationClient.java, method updateDerivationAsyn, you have removed some code and put instead a statement "if isDerivedAsynchronous" with no else clause... This makes me wonder, does this code still work for synchronous derivations? Remember, we need to get to a point where we can use derivations both in synchronous and asynchronous mode. Please make sure we have sufficient unit/regression/integration test coverage so that we can be sure the code works in both modes...

sm453:
Is there a difference between an "upstream" and a "previous" derivation? I'm not sure there is. If there is, then I think this should be clarified (at least in a comment, preferably in clearer variable names). If not, I'd prefer these to be consistently named "upstream" everywhere. Otherwise it's just confusing...

sm453:
> In DerivationClient.java, method updateDerivationAsyn, you have removed some code and put instead a statement "if isDerivedAsynchronous" with no else clause... This makes me wonder, does this code still work for synchronous derivations? Remember, we need to get to a point where we can use derivations both in synchronous and asynchronous mode. Please make sure we have sufficient unit/regression/integration test coverage so that we can be sure the code works in both modes...

Perhaps I should clarify: We need to get the code to a stage where it not only works in both modes, but ideally in a way that hides (from the calling code) whether a derivation is synchronous or asynchronous. It would be undesirable to have two sets of methods (one that should be called for synchronous operation, and a separate one for asynchronous operation). In either mode, the calling code should be calling the same methods (so that if one changes one's mind, one doesn't need to rewrite the calling code).

sm453:
> Perhaps I should clarify: We need to get the code to a stage where it not only works in both modes, but ideally in a way that hides (from the calling code) whether a derivation is synchronous or asynchronous. It would be undesirable to have two sets of methods (one that should be called for synchronous operation, and a separate one for asynchronous operation). In either mode, the calling code should be calling the same methods (so that if one changes one's mind, one doesn't need to rewrite the calling code).

I appreciate that the above may be a bigger task, which may not make sense to be addressed as part of this pull request. The minimum requirement for merging this branch is that the changes don't break synchronous operation. If that is the case, then I would say go ahead.

jb2197:
> > Perhaps I should clarify: We need to get the code to a stage where it not only works in both modes, but ideally in a way that hides (from the calling code) whether a derivation is synchronous or asynchronous. It would be undesirable to have two sets of methods (one that should be called for synchronous operation, and a separate one for asynchronous operation). In either mode, the calling code should be calling the same methods (so that if one changes one's mind, one doesn't need to rewrite the calling code).
> 
> I appreciate that the above may be a bigger task, which may not make sense to be addressed as part of this pull request. The minimum requirement for merging this branch is that the changes don't break synchronous operation. If that is the case, then I would say go ahead.

The current pull request as it is at the moment supports both synchronous and asynchronous operations (it doesn't break anything). I agree that the current framework is not ideal, as you pointed out, it requests developers to use different methods (updateDerivationAsyn and updateDerivations), it also misses a method that is able to deal with a chain of mixed types derivations - all of which I agree should be addressed. Given the naming of this pull request ("140-dev-chain-of-**asyn-derivation**"), it would be really appreciated if we could merge this PR and I am happy to open a new one addressing the above issues.

sm453 (approved):

jb2197:
@kokfoongcmcl The small queries in the method monitorDerivations were replaced with one single query with an enum to determine the type of status. The IntegrationTest was also updated correspondingly in https://github.com/cambridge-cares/TheWorldAvatar/pull/92 and passed on my machine.

kokfoongcmcl:
> @kokfoongcmcl The small queries in the method monitorDerivations were replaced with one single query with an enum to determine the type of status. The IntegrationTest was also updated correspondingly in #92 and passed on my machine.

You are still sending queries in loops in different places,
I believe the first query you need is: 
select ?derivation ?statusType
where{
?derivation a AsyncDerivation. // not really necessary in my view
?derivation hasStatus/rdf:type ?statusType
?derivation isDerivedUsing <agent>}

Here you can get all your derivations along with its status type.

Queries such as isDerivedAsynchronous and hasStatus should be scraped, you can just specify the triples in your queries.

I believe the query you need for checkAtPendingUpdate is
insert{<derivation> hasStatus <status>; <status> a Requested}
where{
	<derivation> isDerivedFrom/belongsTo ?upstreamDerivation
	?upstreamDerivation hasStatus ?status.
	?upstreamDerivation hasTime/inTimePosition/numericPosition ?upstreamDerivationTimestamp
	optional{?upstreamDerivation isDerivedFrom/hasTime/inTimePosition/numericPosition ?pureInputTimestamp}
	optional{?upstreamDerivation isDerivedFrom/belongsTo/hasTime/inTimePosition/numericPosition ?inputsBelongingToDerivationTimestamp}

	filter{?upstreamDerivationTimestamp > ?pureInputTimestamp || ?upstreamDerivationTimestamp > ?inputsBelongingToDerivationTimestamp}
}

instead of checking the inputs and timestamps of your upstream derivation one by one. If this does not work, I think it can be done in 2 queries.

You can easily combine markAsFinished and addNewDerivedIRIToFinishedStatus, since they are just adding triples.

I hope you get the idea.

If you are generating the IRIs using UUIDs, the chances of getting the same string is near zero, you don't need to check that it exists.

jb2197:
> > @kokfoongcmcl The small queries in the method monitorDerivations were replaced with one single query with an enum to determine the type of status. The IntegrationTest was also updated correspondingly in #92 and passed on my machine.
> 
> You are still sending queries in loops in different places, I believe the first query you need is: select ?derivation ?statusType where{ ?derivation a AsyncDerivation. // not really necessary in my view ?derivation hasStatus/rdf:type ?statusType ?derivation isDerivedUsing agent}
> 
> Here you can get all your derivations along with its status type.

Thanks for the suggestion, I have treaked the SPARQL you suggested a bit and replaced the queries in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/91/commits/1c43086548f149900d862b23f7e216690f0cee13.

> Queries such as isDerivedAsynchronous and hasStatus should be scraped, you can just specify the triples in your queries.

Usage of isDerivedAsynchronous at method monitorDerivations is now removed, however, the one used in updateDerivationAsyn is kept as it is necessary to ensure the type of derivation before entering the if clause - this is due to the fact that method updateDerivationAsyn doesn't support synchronous operation in this PR but will be addressed in the future.

Usage of hasStatus in method deleteStatus is now removed, the public method hasStatus is also removed in DerivationClient - these are done in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/91/commits/bedd3be3ad83dcbabc72b1d2b31d5b93394349a8. The usage of hasStatus in method updateDerivationAsyn is kept to separate the logic between checking if there exists status and marking it as PendingUpdate.

> I believe the query you need for checkAtPendingUpdate is insert{derivation hasStatus status; status a Requested} where{ derivation isDerivedFrom/belongsTo ?upstreamDerivation ?upstreamDerivation hasStatus ?status. ?upstreamDerivation hasTime/inTimePosition/numericPosition ?upstreamDerivationTimestamp optional{?upstreamDerivation isDerivedFrom/hasTime/inTimePosition/numericPosition ?pureInputTimestamp} optional{?upstreamDerivation isDerivedFrom/belongsTo/hasTime/inTimePosition/numericPosition ?inputsBelongingToDerivationTimestamp}
> 
> ```
> filter{?upstreamDerivationTimestamp > ?pureInputTimestamp || ?upstreamDerivationTimestamp > ?inputsBelongingToDerivationTimestamp}
> ```
> 
> }
> 
> instead of checking the inputs and timestamps of your upstream derivation one by one. If this does not work, I think it can be done in 2 queries.

I agree the loop of checking each upstream derivation in the method checkAtPendingUpdate can be replaced with one single query, but I don't think the insert should be done within the same query. The query suggested is really helpful and I treaked it a bit, you can see it in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/91/commits/81db2e8799bd0af702ea350fc9428908db4eb123 and https://github.com/cambridge-cares/TheWorldAvatar/pull/91/commits/818b1c9597b5255f53ca22536ad49178bcd3f31e.

> You can easily combine markAsFinished and addNewDerivedIRIToFinishedStatus, since they are just adding triples.

Thanks for the suggestion, this is now updated in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/91/commits/fb2c9f5ad6693828f7882618e3275b75ab0b606e.

> I hope you get the idea.
> 
> If you are generating the IRIs using UUIDs, the chances of getting the same string is near zero, you don't need to check that it exists.

This checking is removed throughout DerivationSparql as in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/91/commits/df5aeef528eff8748c1b8a2372ec5e8ae29ca5a3. The method checkInstanceExists is also marked as deprecated.

kokfoongcmcl (approved):

kokfoongcmcl:
> Usage of isDerivedAsynchronous at method monitorDerivations is now removed, however, the one used in updateDerivationAsyn is kept as it is necessary to ensure the type of derivation before entering the if clause - this is due to the fact that method updateDerivationAsyn doesn't support synchronous operation in this PR but will be addressed in the future.

You don't have to do it in this PR but you should query everything you need regarding the derivation including its rdf:type with a single query at each stage of the recursive call if you cannot query everything. So that function is not necessary.

jb2197:
> > Usage of isDerivedAsynchronous at method monitorDerivations is now removed, however, the one used in updateDerivationAsyn is kept as it is necessary to ensure the type of derivation before entering the if clause - this is due to the fact that method updateDerivationAsyn doesn't support synchronous operation in this PR but will be addressed in the future.
> 
> You don't have to do it in this PR but you should query everything you need regarding the derivation including its rdf:type with a single query at each stage of the recursive call if you cannot query everything. So that function is not necessary.

That makes sense, we can also move away from calling isOutOfDate. I shall modify the codes in the next round of development.

PR #92, opened 2021-12-06
139 dev minimum working example of asyn agent

jb2197:
This PR is a minimum working example of cascade asynchronous derivation enabled by https://github.com/cambridge-cares/TheWorldAvatar/pull/91.


sm453:
In the dockerfile, there is a comment:
# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
Shouldn't it be the other way round? Surely, if the encrypted password contained any "|", then it would be a bad idea to use that as a delimiter... (I assume the code is correct, it's just the comment that may need fixing.)

sm453 (approved):

kokfoongcmcl:
there's an extra curly bracket in the query provided

jb2197:
> there's an extra curly bracket in the query provided

Thanks for pointing it out, it's now fixed in the commit https://github.com/cambridge-cares/TheWorldAvatar/pull/92/commits/21cd44d5fb46823a3c3ada4f37f27e28329bc850

jb2197:
> In the dockerfile, there is a comment: "(Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter" Shouldn't it be the other way round? Surely, if the encrypted password contained any "|", then it would be a bad idea to use that as a delimiter... (I assume the code is correct, it's just the comment that may need fixing.)

Thanks for pointing it out, it is now fixed in the commit https://github.com/cambridge-cares/TheWorldAvatar/pull/92/commits/a4ab06f3f9f1e4f21b160de2a4c476e4792df0ea.

kokfoongcmcl (approved):

PR #94, opened 2021-12-09
dev-improveFileUpload: used FileEntity instead of StringEntity in pos…

kokfoongcmcl:
@dln22 I'm improving the uploadRDFFile method in the RemoteStoreClient, previously the method will fail for large files. I've renamed it to uploadFileToBlazegraph, let me know if it's an issue.

### DIFF
--- a/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/RemoteStoreClient.java
+++ b/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/RemoteStoreClient.java
@@ -800,13 +800,15 @@ public class RemoteStoreClient implements StoreClientInterface {
     }
 
     /**
-     * upload a file to a Blazegraph using its REST API
-     * check Blazegraph's documentation for the supported file types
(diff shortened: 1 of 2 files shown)


sm453:
I'm not saying this needs to be addressed right now, but it is highly problematic that the RemoteStoreClient, which is meant to be triple-store-agnostic, contains any functionality that is specific to one particular technology (blazegraph). That defeats the whole point of the RemoteStoreClient...

kokfoongcmcl:
> I'm not saying this needs to be addressed right now, but it is highly problematic that the RemoteStoreClient, which is meant to be triple-store-agnostic, contains any functionality that is specific to one particular technology (blazegraph). That defeats the whole point of the RemoteStoreClient...

It’s done through the REST API, so it might not be Blazegraph specific, it might work with other triple-stores if they accept the same content types, just that I’ve only tested it with Blazegraph so far

kokfoongcmcl:
> I'm not saying this needs to be addressed right now, but it is highly problematic that the RemoteStoreClient, which is meant to be triple-store-agnostic, contains any functionality that is specific to one particular technology (blazegraph). That defeats the whole point of the RemoteStoreClient...

It seems that it is a requirement for SPARQL endpoints to support REST API, but it's not very clear with the file formats that it needs to support. I've tested this with RDF4J and it works fine.

PR #95, opened 2021-12-16
Dev timeseries units

kokfoongcmcl:
changed the way units and table headers are provided to the method, they need to be maps with the IRIs as the key. The getDataIRIs method returns a list from a map, therefore the order is not guaranteed to be the same each time it is called


kokfoongcmcl:
Consequence of this is all codes that are using this need to be updated (they were incorrect anyway)

markushofmeister (approved):
Those changes make perfect sense.

As my code mainly uses this convertToJSON method via the py4jps wrapper maintained by Daniel, we would need to package and upload the JPS_BASE_LIB after merging those changes to develop and a new version of the py4jps package to be released which includes those changes.

I will make sure that the "Time Series Example" agent gets updated.  @msff2: Are you updating the "Gas Grid Agent" or shall I take care of it?

PR #98, opened 2021-12-22
Cesium ion support and bug fix for hover highlighting

jeffechua:
Transitioned from subscribing to mouse events on every layer to a single listener which hits again all layers, and reduced to only listening to "mousemove". This fixes a bug I was experiencing where if there were multiple layers in the line of view, the hover state would glitch and cause features to be stuck highlighted.

Also moved some code from Manager.initialiseMap() to index.html, which is a prerequisite for smooth Cesium ion implementation.

There is also some Cesium ion implementation work in this pull request but only strictly additively apart from that described above.


jeffechua:
This branch currently contains working Cesium ion support by wrapping the Cesium ion viewer object in a class providing Mapbox Map methods. There is a demo in cesium.html, in the same directory as the original index.html demo. It consumes the exact same data as the existing Mapbox implementation.

**Mainly I want to get the mouse interaction change (interaction_handler.js), outlined in the OP of this request, merged so the branches do not diverge.** This is necessary for Cesium ion compatibility and also fixes some odd behavior with overlapping features. **On the other hand, the Cesium ion support, specifically the new js/cesium_wrapper.js and cesium.html files, do not need to be reviewed**. They are WIP and do not really belong in this branch or pull request, but there is no harm in merging at this time: they live outside the existing framework and do not interact with the Mapbox flow at all.

sm453:
@jeffechua, you have added a number of huge json and geojson files, e.g. in commit 0f5460287d2b9752f925551928b60ac9d1951d28, e.g. in the folders
web/UKDigitalTwin/example-dt-vis/data/set-0/2D/
and
web/UKDigitalTwin/example-dt-vis/data/set-0/3D/.

Such (data) files should never be added to any source code repository (because they dramatically inflate the size of the repository). Please remove them ASAP. (And note to yourself for future reference that deleting the files does not reduce the size of the repository...)

mdhillmancmcl:
To me it looks like something funky has gone on with some of the commits here; when I attempt to merge from “dev-combined-vis” into your “dev-cesium-vis” branch (which would be the final step before approving the merge), it appears that a number of older changes to the code are overwriting newer ones. I’m not quite sure what these “Quicksave” and “Quick commit…” commits hoped to achieve, could these be source of the issues?

My understanding of this PR was to bring in a fix to the highlighting of buildings (which is great), but I also see a number of additions concerning Cesium Ion support. I’m not sure what Cesium offers let alone whether we want to add this to the central vis framework at this point (or at all).

For the time being I think it’s worthwhile creating a new branch from “dev-combined-vis”, re-adding your bug fix, then requesting a new PR **solely** for that. 

_I will follow up with @jeffechua and @sm453 (and/or @jwja2) to get this bug fix in and discuss the possibility of adding Cesium support in future._

jeffechua:
There were not merge conflicts at the time of the creation of this pull request, and I fixed conflicts a few times since then, but further divergence between your branch and mine has led to the current conflicts. I will fix the conflicts when the new behaviour and content of the pull request are approved.

The key change in interaction_handler.js is that instead of adding a new mouse event listener for every layer, only one listener is added, and it handles events for all layers.

I agree that we should schedule some time to have a proper discussion about this.

Issue #99, opened 2022-01-04
Derivation time stamps updated to incorrect values


sm453:
The value of the time stamp at which a derivation was last updated must be the instant in time at which the inputs to the derivation (i.e. the entities which the derivation is derived from) were read. The value of the time stamp at which a derivation was last updated MUST NOT be the instant in time the entities belonging to the derivation (i.e. the entities whose values are derived) were written.

Unfortunately, the current implementation appears to do the latter, not the former. This must be rectified.

Please note that the difference between the two implementations is critical. In the latter one, we are susceptible to the nasty bug that an input value was updated AFTER it was read but BEFORE the derivation was updated, but the time stamp associated with the derivation will be AFTER the update of the input, thus erroneously suggesting the derivation is up to date while in reality it is not!

One may argue that, for synchronous derivations, this is not too serious because the wall time delay between reading an input and updating a derivation may be 'small'. For asynchronous derivations, this is much more serious, because by definition the time delay between reading the inputs and writing the outputs is 'large'.

I am therefore convinced that, for a robust implementation, both modes (synchronous and asynchronous) must update a derivation time stamp to the instant in time at which the inputs were READ.

kokfoongcmcl:
The values are read by the agent itself, and the derivation client only provides the IRIs to the agent. For the synchronous side, the easiest thing I can do is to change the timestamp to the timestamp at the point when it sends the HTTP request to the agent.

I think what you really want is for the agent to update the timestamp at the point it reads the input. To do this really properly and efficiently on the synchronous side, the agent needs to include the timestamp in its HTTP response. This is because the timestamps are updated in 1 go only after all the derivations are updated (updating the timestamps 1 by 1 after each agent call is inefficient). Do you prefer this option? This enforces another kind of manual requirement on the agent that is difficult to enforce with a superclass.

There is a public method in the derivation client that updates the timestamp for a given derivation IRI or the instances belong to a derivation, I guess the asynchronous agents can use this method.

kokfoongcmcl:
Actually we can create three classes: 1) derivation agent, 2) agent inputs, 3) agent outputs. The "derivation agent" class will have "agent inputs" as its argument and must return an "agent outputs" object. This "agent outputs" object will have timestamp as the compulsory argument in the constructor.

[...]

PR #103, opened 2022-01-07
143 dev fix async derivation timestamp

jb2197:
Fixes (asynchronous derivation mode) for the issue raised in https://github.com/cambridge-cares/TheWorldAvatar/issues/99


sm453 (approved):
There is a typo in "retireveInputReadTimestamp". Please correct.
Apart from that it looks ok to me.

PR #109, opened 2022-01-19
144 dev set return at async derivation pendingupdate

jb2197:


jb2197:
A very small change to the return data type of a method, no changes to the design of the DerivationClient nor AsynAgent. The IntegrationTest as in DerivationAsynExample has been tested locally, so I will just approve this myself.

PR #111, opened 2022-01-21
142 dev python async agent

jb2197:
A python wrapper to standardise the development of asynchronous agents in python. Can be installed via pip install asyncagent.


sm453:
In async_agent.py, line 64, it should be "if self.kgUser is None:"?

sm453:
In async_agent.py, member function add_url_pattern, I think it would be good to add to the documentation that use of this is (strongly) discouraged. The whole point of an asynchronous agent is to communicate via the KG, and NOT via HTTP/web requests.

sm453:
In data_model/utils.py, please change the docstring of trimIRI to:
"This method deletes the '<' and '>' around the given IRI (or lists of IRIs)."

sm453:
In kg_operations/sparql_client.py, getNumberOfTriples should really be a method of the RemoteStoreClient (or ideally even of its parent class), so that you don't need to (re-)implement it here (like every other developer in every project...).

jb2197:
> In async_agent.py, line 64, it should be "if self.kgUser is None:"?

This is fixed in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/111/commits/7ea4b79fb995417dab9d24f030ae9acddb5b0346.



> In async_agent.py, member function add_url_pattern, I think it would be good to add to the documentation that use of this is (strongly) discouraged. The whole point of an asynchronous agent is to communicate via the KG, and NOT via HTTP/web requests.

This is fixed in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/111/commits/56816402bdb352e9680004ad57f86efcfed26c2e.



> In data_model/iris.py, it is questionable to define large numbers of hard-coded IRI strings to project-specific concepts like Vapourtech and HPLC. Remember that this is meant to be a generic agent template. I would keep only the most generic definitions in that file, and base URLs like theworldavatar.com or the location of TWA GitHub repository should be defined explicitly in one place only (and all strings assembled from them should do that using variables).

The base URLs are provided as variables and adopted for other strings, as done in https://github.com/cambridge-cares/TheWorldAvatar/pull/111/commits/db3341ce1fb1bd553afa4c267985c38674055a8f. The project-specific concepts are left un-removed to avoid code duplication in different agents that require access to the same concept IRIs. A better way of doing this might exist and will be provided in future development.



> In data_model/utils.py, please change the docstring of trimIRI to: "This method deletes the '<' and '>' around the given IRI (or list of IRIs)."

This is fixed in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/111/commits/768faf43264258d093458876a1e062e7cdddda04.



> In kg_operations/sparql_client.py, getNumberOfTriples should really be a method of the RemoteStoreClient (or ideally even of its parent class), so that you don't need to (re-)implement it here (like every other developer in every project...).

The suggested function was implemented in https://github.com/cambridge-cares/TheWorldAvatar/pull/114, it is now accessed as method getAmountOfTriples() done in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/111/commits/7204eb65a36eeb129a2daaabdc789a26c05bd6bf.

sm453 (approved):

Issue #127, opened 2022-02-08
The FileServer.writeFilePart method should return the full URL of the uploaded file


gpeb2:
It would be helpful to have the FileServer.writeFilePart methdto return a full path/URL to the uploaded file (e.g. http://address/download/subdir1/subdir2/file1.txt) so that it can be used to retrieve the file again.

PR #128, opened 2022-02-08
Dev kg file server

gpeb2:
Closes issue #127


Deploy/stacks/db/fileserver/fileserver/src/main/java/com/cmclinnovations/fileserver/FileServer.java
```diff
@@ -0,0 +1,198 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ *
+ * @author Owen Parry {@literal <oparry@cmclinnovations.com>}
+ */
+@WebServlet(name = "FileServer", urlPatterns = {FileServer.DELETE_URL_PATTERN, FileServer.DOWNLOAD_URL_PATTERN, FileServer.UPLOAD_URL_PATTERN, FileServer.UPLOAD_URL_PATTERN + "/"})
+@MultipartConfig(
+fileSizeThreshold = FileServer.ONE_MB_IN_B,
+location = "/app/fs_root/",
```
gpeb2, line 27: Make it possible to set the file size settings via environment variables.

Deploy/stacks/db/fileserver/fileserver/src/main/java/com/cmclinnovations/fileserver/FileServer.java
```diff
@@ -0,0 +1,198 @@
...
+        /*
+         * Wait for a millisecond to ensure the next call of this function never generates the same
+         * fallback suffix.
+         */
+        try {
+            wait(1);
+        } catch (InterruptedException ex) {
+            LOGGER.error("writeFilePart: (Thread)InterruptedException whilst pausing at the end of the method", ex);
+        }
+        return destPath.toString();
```
gpeb2, line 95: Return the full file path/URL so that it can be used as an argument of the get or delete methods. See issue #127.

Deploy/stacks/db/fileserver/fileserver/src/main/java/com/cmclinnovations/fileserver/FileServer.java
gpeb2: These settings have been moved to the web.xml file and are populated during the Maven build stage.

PR #131, opened 2022-02-16
151-fix-update-ontochemexp-dependency: updated import and dependency …

jb2197:
…of KnowledgeRepository (it was moved from JPS_BLAZEGRAPH to be part of JPS_BASE_LIB) to make project compilable.


jb2197:
@sganguly-cares The project is now compilable on my machine, please let me know if there's still an issue in compiling from your side.

PR #138, opened 2022-02-24
184 mops csv to json scripts

Kondinski:
Hi, could you please look at the code? Thanks!


am2145 (commented):
I have had a look. All of the proposed commits are self-contained and should be fine except for one. There is a change in the CompChemParser to do with getting elements' molecular weight by their symbol. I remember an issue with this arising from multiple letter symbols. I believe Aleks and I implemented a temporary fix several months back, but Daniel subsequently fixed it on develop. Daniel, could you confirm this? If so, I would not change the parser. Otherwise, we will need to double check if we want to accept the changes from Aleks' branch.

dln22:
> I have had a look. All of the proposed commits are self-contained and should be fine except for one. There is a change in the CompChemParser to do with getting elements' molecular weight by their symbol. I remember an issue with this arising from multiple letter symbols. I believe Aleks and I implemented a temporary fix several months back, but Daniel subsequently fixed it on develop. Daniel, could you confirm this? If so, I would not change the parser. Otherwise, we will need to double check if we want to accept the changes from Aleks' branch.

I have just had a look, and indeed that issue was fixed by you Angiras :) and is already on develop. I have suggested changes that Aleks needs to make to the "thermo/CoMoCompChemParser/compchemparser/helpers/elements_data.py" file. Once that is done I am happy for his code to be merged.

PR #144, opened 2022-03-03
153 dev update async derivation

jb2197:
In this update, I provided:

1. For those async derivations created requiring an update, mark the status of DerivationAsyn as Requested at its creation
2. For those async derivations only created to markup the information, mark timestamp of DerivationAsyn as current timestamp at its creation
3. Changed from oneOrMore(rdf:type) to zeroOrMore(rdf:type)/zeroOrMore(rdfs:subClassOf) to support arbitary length of subclass relationship, both at instance and class level
4. Add option for creating derivation given another derivation, the matching is done in a similar fashion as the matching between derivation inputs to agent inputs
5. Let AsyncAgent check if any immediate upstream derivations are yet to be updated at the status Requested
6. Removed the status PendingUpdate throughout
7. Add unit test cases and update the integration tests for asynchronous derivation


sm453 (approved):
Large binaries, and this includes pixel images, are generally not a good idea for source code repositories. A plain-text vector format, e.g. svg, is much preferable.

Please remember to remove SNAPSHOT versions from pom files before merging.

PR #150, opened 2022-03-08
Dev mouseover fix

jeffechua:
Fix for bug where multiple polygon layers break the highlighting feature if the user mouses over a point intersecting multiple layers' polygons, causing polygons to be stuck highlighted. The issue is not seen when mousing over multiple polygons in the same layer.

This is because how the interaction handler is structured, adding a mouse event listener for every layer. In my fix, I switch to only one listener which handles all layers.

I have committed a data sample to dev-combined-vis to highlight this feature. In the "3D" data set, if you mouse over some of the example house data, either in the 2D or 3D group, the issue is observed. In dev-mouseover-fix, mousing over the same data set behaves correctly.


mdhillmancmcl:
I am unfortunately still unable to approve this merge request, changes I have made to the dev-combined-vis branch (as recent as 01/03/2022) are being overriden by older versions of the code on this branch.

For reference, the changes to the handleMultipleFeatures() function within the InteractionHandler class (made on the 01/03/2022, commit #9b07a63d37967d47928437b290b74f0f305c3ba8) are overwritten with an older version of the code that somehow crept onto the dev-mouseover-fix branch.

The following is what I'll need to approve this bug fix being merged:
- A _brand new_ branch created from the HEAD of dev-combined-vis
- A single commit on that new branch containing the bug fix
- No changes to any other files, or functions, unrelated to the bug fix

I am closing this PR as I expect a new one (from a new branch) to be submitted, please contact me if there are any issues with this.

PR #167, opened 2022-03-28
1061 jpsbaselib unit tests

mehalaga:


csl37 (changes requested):
For a few tests, it would be better to use a temporary folder rather than writing to and deleting files in the home directory

csl37 (approved):

PR #175, opened 2022-04-06
Added a top level .gitattributes file

gpeb2:
Added a top level .gitattributes file so that we can force certain file types to have specific line endings when cloned. For instance shell scripts (.sh) should always have Linux line endings. Git defaults to checking out text files with the line endings of the local OS, which can cause issues particularly when the shell scripts need to be run in Docker containers.


sm453:
Forcing all xml files to check out with Linux line-endings seems heavy-handed to me. Can we really assume that there is no application where some xml file is required on Windows with the correct line-endings? Even if that isn't presently the case, can we really be sure that this isn't going to be required at some point in the future? I'm not so sure.

Am I guessing correctly that, treating xml files as default text (so that they are checked out with the appropriate line-endings for each OS) produces Windows line-endings on WSL? I understand that for shell scripts this is a real problem. What's the issue with xml files?

mdhillmancmcl:
> Forcing all xml files to check out with Linux line-endings seems heavy-handed to me. Can we really assume that there is no application where some xml file is required on Windows with the correct line-endings? Even if that isn't presently the case, can we really be sure that this isn't going to be required at some point in the future? I'm not so sure.
> 
> Am I guessing correctly that, treating xml files as default text (so that they are checked out with the appropriate line-endings for each OS) produces Windows line-endings on WSL? I understand that for shell scripts this is a real problem. What's the issue with xml files?

I am for making this change for the shell scripts, but like Seb, I don't quite see the need to change the XML files.

gpeb2:
I've removed the rule for XML files and reverted any changes to them. I've also ensured any recent changes made on the main branch have been merged correctly. Please approve the pull request if you're happy or add another comment if you're not.

In answer to your questions:
> Forcing all xml files to check out with Linux line-endings seems heavy-handed to me. Can we really assume that there is no application where some xml file is required on Windows with the correct line-endings? Even if that isn't presently the case, can we really be sure that this isn't going to be required at some point in the future? I'm not so sure.

This shouldn't be the case as using Linux line-endings for XML files is specified in the standards but I've also rarely, if ever, encountered a program that read XML and cared about line-endings either way so leaving them to have the OS default should be fine.
> Am I guessing correctly that, treating xml files as default text (so that they are checked out with the appropriate line-endings for each OS) produces Windows line-endings on WSL? I understand that for shell scripts this is a real problem. What's the issue with xml files?

To clarify, checking out in Windows, through Tortoise Git or git run in cmd, will cause the "text" files to have Windows line-endings but running "git clone" from a WSL terminal causes them to have Linux line-endings.
The issue is not just with WSL but also occurs if those scripts are copied into Docker containers, which are generally Linux based.

sm453 (approved):

PR #176, opened 2022-04-07
Included addition of time series data in bulk

markushofmeister:
Include new bulkaddTimeSeriesData for TimeSeriesClient to allow for addition of multiple time series data within one RDB connection


kokfoongcmcl (commented):
Looks ok, if I'm being nitpicky this does not warrant a version bump because it shouldn't break any existing code unless they're using the rdb client directly

sm453:
Fine with me, too.

sm453 (approved):

PR #181, opened 2022-04-11
Dev pyuploader

dln22:
Extended a list of possible file formats for the triple store uploads.


JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/RemoteStoreClient.java
```diff
@@ -892,4 +892,93 @@ public void uploadFile(File file) {
             throw new JPSRuntimeException("Upload RDF file failed.", ex);
```
gpeb2, line 892: Presumably this can be replaced by a call to the override of this function with the extension as an argument.

gpeb2 (changes requested):

gpeb2 (approved):

PR #182, opened 2022-04-13
Dev timeseries convert to json fix for missing values

markushofmeister:
Updated the convertToJSON method to derive the value class from the first not null value in the data series instead of the very first value element to avoid NullPointer exceptions in case of missing data

### DIFF
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -12,7 +12,7 @@
     details on how version numbers should be selected -->
 
 
-    <version>1.8.0</version>
+    <version>1.8.1</version>
 
     <!-- Project Properties -->
     <properties>
--- a/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeries.java
+++ b/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeries.java
@@ -62,11 +62,13 @@ public class TimeSeries<T> {
     	List<?> v = getValues(dataIRI);
     	if (v == null) {
     		return null;
-    	} else if (v.get(0) instanceof Number) {
-    		return v.stream().map(value -> ((Number) value).doubleValue()).collect(Collectors.toList());    		
     	} else {
-    		throw new JPSRuntimeException("TimeSeries: Values for provided dataIRI are not castable to \"Number\"");
-    	}   	
+            try {
+                return v.stream().map(value -> value == null ? null : ((Number) value).doubleValue()).collect(Collectors.toList());
+            } catch (Exception e) {
+                throw new JPSRuntimeException("TimeSeries: Values for provided dataIRI are not castable to \"Number\"");
+            }
+        }
     }    
     
     /**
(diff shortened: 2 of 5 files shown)


sm453 (changes requested):
Rather than using explicit integer indexing, please rewrite your loop using an iterator.

markushofmeister:
> Rather than using explicit integer indexing, please rewrite your loop using an iterator.

This comment has been addressed by commit d96bb290787b0fe57abc78fec3f03848eb4f43f7

markushofmeister:
> Careful - every time you call next() on an iterator, it moves the iterator to the next element! You've got three occurrences of next() in what is meant to be a single loop iteration...

This should be fixed by commit 71bc2b81720cc171507b544baed3d66bc9adb263

kokfoongcmcl:
I'm not sure what will the visualisation show if you set the class to "Unknown", as far as I know the current framework accepts either "Boolean" or "Number" (because of the CReDo project). This can certainly be improved and requires coordination with Michael

markushofmeister:
> I'm not sure what will the visualisation show if you set the class to "Unknown", as far as I know the current framework accepts either "Boolean" or "Number" (because of the CReDo project). This can certainly be improved and requires coordination with Michael

Just looping in @mdhillmancmcl on this topic.

The behavior of the DTVF is as follows:
1) For missing (i.e. null) values in a "Numbers" series, the data points will simply not been shown (figure 1)
2) Setting the class to "Unknown" does not seem to cause any issues (at least if all values as missing, i.e. null) (figure 2)

Hence, I believe there is no further amendment needed; however, we could change "Unknown" to "Number" with all values missing if we really want to push for explicitly supported value formats.

Let me elaborate why we need this in the first place: While developing the agent to instantiate and retrieve the MetOffive live weather data I noticed, that normally 6-9 parameters are reported at the same frequency; hence, are stored as one time series. It can happen that some values are not always reported, e.g. wind gust. Depending on the time frame for which to retrieve data, it can happen that all values for an individual time series (i.e. dataIRI) might be missing (i.e. null).

Handling of null values in Number class
Handling of Unknown class

kokfoongcmcl:
If it's all empty, I think it's better to just take out the column from the JSON?

kokfoongcmcl (approved):

mdhillmancmcl:
For an operation this simple, I'm happy using a for-loop. But it's worth noting that there are places (particularly in any code that _could_ be run by multiple threads) that better suit iterators, or Java 8s Streams API (which is faster over larger collections).

mdhillmancmcl (approved):

kokfoongcmcl:
also a note, the getValuesAs____ methods in the TimeSeries class need to be improved to deal with nulls

markushofmeister:
> also a note, the getValuesAs____ methods in the TimeSeries class need to be improved to deal with nulls

This should be addressed by commit https://github.com/cambridge-cares/TheWorldAvatar/pull/182/commits/48a8a336161661174cd08408268f0b5e48f90390

kokfoongcmcl (approved):

sm453 (approved):

Issue #184, opened 2022-04-19
Derivation Framework: Concurrent HTTP request handling


jb2197:
# Issue
During testing the mixed type of derivation DAGs (case 5 in DerivationAsynExample), an issue was found due to the way how tomcat handles the concurrent HTTP requests. In the derivation DAG, two `OntoDerivation:Requested` asynchronous derivations (MaxValueDerivation and MinValueDerivation) depend on the same synchronous derivation (RNGDerivation). As the derivation agents (MaxValue Agent and MinValue Agent) were initialised with the same periodical timescale for monitoring derivations, they fire the HTTP request for an update of RNGDerivation at the same time. Upon receiving the concurrent HTTP requests, RNG Agent handles them with different threads (i.e. http-nio-8080-exec-4 and http-nio-8080-exec-5 in the log file), therefore, the RNG Agent computes the derivation outputs with the same inputs twice and return the response **respectively** to MaxValue Agent and MinValue Agent. The two sets of outputs were then **both** written to the knowledge graph by the derivation client that ran by MaxValue/MinValue Agent, yielding multiple outputs `OntoDerivation:belongsTo` RNG Derivation, also MaxValue/MinValue Derivation both `OntoDerivation:isDerivedFrom` multiple inputs - as shown in below screenshot. Both MaxValue/MinValue Agent were coded that they assume there's only one IRI passed in so it operates on the first IRI in the list, although the derivation inputs retrieved by both agents will have two instances. The rest of the computation and the function of Difference Agent are thus still able to complete execution without errors (although the results are inconsistent with the multiple outputs/inputs derivation markup). The test case of case 5 (testMaxMinDiffAsync(), current commented out) in DockerIntegrationTest.java fails as it checks if the outputs generated by RNG Derivation contain the correct amount of points - which is expecting 7 instead of the actual 14 due to multiple calculations.

RNG Derivation marked with multiple outputs: 
![rng_derivation_multiple_outputs](https://user-images.githubusercontent.com/49490446/164014230-4997502b-9458-4406-b3ae-bf58ddefb48b.PNG)

MaxValue Derivation marked with multiple inputs: 
![max_derivation_multiple_inputs](https://user-images.githubusercontent.com/49490446/164014883-06d2b20a-1cf9-47c0-8505-336f38d1cbb9.PNG)

MinValue Derivation marked with multiple inputs: 
[...]

PR #185, opened 2022-04-19
154 dev unified derivation framework

jb2197:
This PR follows up on the discussion in https://github.com/cambridge-cares/TheWorldAvatar/issues/99 and provides a unified derivation framework that is able to update a directed acyclic graph (DAG) formulated by pure sync derivations/pure async derivations/async derivations depending on sync derivations.

This PR also provides additional support for two modes of asynchronous operation: (1) new information generation; and (2) existing information update. Both modes work fine with the mixed type DAG as aforementioned.

DerivationExample and DerivationAsynExample were updated using the latest codes. For more details, please refer to README.md of each minimum working example.

An issue was encountered during one of the tests (case 5, testMaxMinDiffAsync in DockerIntegrationTest.java of DerivationAsynExample) and discussed in detail in https://github.com/cambridge-cares/TheWorldAvatar/issues/184.


jb2197:
> I believe that all queries must not be looped, what you need to write are SPARQL updates with sub queries in it, see Example 6 [here](https://www.w3.org/TR/sparql11-update/).
> 
> I've only looked at the synchronous bit in detail because I'm familiar with it:
> 
> * updatePureSyncDerivation
>   I already mentioned this, please do not update timestamps in the loop, I purposely written the function to update all timestamps in 1 go. The whole point of `derivation.setUpdateStatus(true); ` is to know which timestamps to update, which is redundant now that you replaced it in the loop. The only queries that I left in loops are the reconnecting ones, because they are not actively used compared to the timeseries one, they can certainly be replaced as well.

The timestamp update in the loop is to make sure the world outside the memory/cache of the calling entity of `updatePureSyncDerivation` or `updateDerivations` functions also gets the most up-to-date information about the derivation instances - so that multiple entities can call the `updatePureSyncDerivation` function at the same time, like what `monitorDerivations` in the DerivationAgent is doing as you noticed. This is actually quite important to enable the unified derivation framework that is able to handle mixed DAGs.

You may also notice that the SPARQL query for reconnecting inputs to derivations, timestamp update and also delete any status for sync derivations (this will be applicable when updating mixed DAGs) are all included in one function `reconnectInputsUpdateTimestampDeleteStatus` https://github.com/cambridge-cares/TheWorldAvatar/blob/fbb895e8233b397e26f7e622111207985b7ba08d/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationSparql.java#L1785. 

I agree this increased the number of SPARQL queries in the update and might slow down the response time when there's a large number of sync derivations to be updated, but I think the amount of queries increased is not too bad if we compare the number of queries in below situation:

For simplicity, assume we are updating 1000 synchronous derivations in the middle of a DAG and among which 200 are `DerivationWithTimeSeries`, leaving 800 normal `Derivation` (assume 500 of them have downstream derivations, i.e. are inputs to another derivation, so `inputToAnotherDerivation.size() > 0`)
now we calculate the number of SPARQL queries within the `if (derivation.isOutOfDate())` clause:
| Individual SPARQL query/update | `updatePureSyncDerivation` | `updateDerivation` |
| --- | --- | --- |
| deleteBelongsTo | 800 | 800 |
| addNewEntitiesToDerived | 800 | 800 |
| initialiseNewEntities | 500 | 800 |
| reconnectInputsUpdateTimestampDeleteStatus | 1000 | N/A |
| reconnectInputToDerived | N/A | 500 |
| updateTimestamps | N/A | 1 |
| --- | --- | --- |
| **total** | 3100 | 2900 |

**PS:**
- initialiseNewEntities is moved to within the `if (inputToAnotherDerivation.size() > 0)` clause for updatePureSyncDerivation, we can also do the same for updateDerivation, but the difference is still okay-ish in the above situation (3100 vs. 2600).
- I suppose the number of SPARQL queries can be further reduced by combing `deleteBelongsTo` and `addNewEntitiesToDerived`, even `initialiseNewEntities` into one SPARQL query/update call


Given the advantage of enabling distributed requests for sync derivation update, which is the key to enabling mixed type DAG, I would strongly advocate keeping the timestamp update after each derivation update. It should be noted tho, that this advantage comes with a price that we also need to solve the issue with concurrent HTTP requests in the long run as detailed in https://github.com/cambridge-cares/TheWorldAvatar/issues/184.

Please check if you agree with the suggestions and I can make changes to combine those SPARQL queries.

> * I find the mixture of deprecated functions confusing.

The intention was to provide the updated methods but not break any of the existing codes - in theory, the newly provided functions should be able to cover the same functionalities as the codes that are marked as deprecated, but the developer can choose if and when to switch to newer functionalities. Would you mind elaborating on which exact function was confusing?

kokfoongcmcl:
It's going to scale very badly if the number of queries increases with the number of derivations. 

I always believe that there should be only one entity that can call the updateDerivation functions, having monitorDerivation to call it is a bad idea, why should this be triggerred by a derivation agent? I think this should be called by a higher up agent

kokfoongcmcl:
and FYI, all the queries above updateTimestamps are not called in the CReDo project and they can be replaced, so it's 1 query against 3100

kokfoongcmcl:
> I agree this increased the number of SPARQL queries in the update and might slow down the response time when there's a large number of sync derivations to be updated, but I think the amount of queries increased is not too bad if we compare the number of queries in below situation:

well more accurately the number of queries to update timestamps, so 800 against 1

kokfoongcmcl:
> 800

jb2197:
> For simplicity, assume we are updating 1000 synchronous derivations in the middle of a DAG and among which 200 are `DerivationWithTimeSeries`, leaving 800 normal `Derivation` (assume 500 of them have downstream derivations, i.e. are inputs to another derivation, so `inputToAnotherDerivation.size() > 0`) now we calculate the number of SPARQL queries within the `if (derivation.isOutOfDate())` clause:
> 
> | Individual SPARQL query/update | `updatePureSyncDerivation` | `updateDerivation` |
> | --- | --- | --- |
> | deleteBelongsTo | 800 | 800 |
> | addNewEntitiesToDerived | 800 | 800 |
> | initialiseNewEntities | 500 | 800 |
> | reconnectInputsUpdateTimestampDeleteStatus | 1000 | N/A |
> | reconnectInputToDerived | N/A | 500 |
> | updateTimestamps | N/A | 1 |
> | --- | --- | --- |
> | **total** | 3100 | 2900 |
> 

In commit https://github.com/cambridge-cares/TheWorldAvatar/pull/185/commits/104fd7ee3100e6285b44e61ad4528869abb8b6b8, method `deleteBelongsTo`, `addNewEntitiesToDerived`, and `reconnectInputsUpdateTimestampDeleteStatus` were combined into one method `reconnectNewDerivedIRIs`, which will only be executed if the derivation is still outdated (which partially solves the concurrent HTTP request issue), so that the amount of SPARQL operations for the hypothetical situation becomes:

| Individual SPARQL query/update | `updatePureSyncDerivation` |
| --- | --- |
| initialiseNewEntities | 500 |
| reconnectNewDerivedIRIs | 1000 |
| --- | --- | --- |
| **total** | 1500 |

> It's going to scale very badly if the number of queries increases with the number of derivations.

The number of queries now scales linearly with the number of derivations: the number of derivations to be updated + the number of derivations that have downstream derivations.

kokfoongcmcl:
The number of queries should not be a function of the number of derivations, there should be zero queries within the updateDerivation functions. The non-timeseries derivation is not actively used at the moment so it's not a big issue, I do plan on making the change when we start to use it.

jb2197:
> The number of queries should not be a function of the number of derivations, there should be zero queries within the updateDerivation functions. The non-timeseries derivation is not actively used at the moment so it's not a big issue, I do plan on making the change when we start to use it.

As discussed, we would like to achieve a decentralised design of agents operating over the knowledge graph. It is thus important to allow multiple agents to query and access the same piece of information at the same time, therefore, the updated information needs to be recorded in the knowledge graph after the update of each derivation, hence the SPARQL updates in the loop are inevitable. 

However, the scaling issue related to the pure DAG of `OntoDerivation:DerivationWithTimeSeries` is valid and should be addressed. As the focus of this PR is to provide a unified derivation framework to support derivation DAG with async derivations depending on sync derivations, the issue with timeseries derivations will be addressed in another PR on its own, https://github.com/cambridge-cares/TheWorldAvatar/issues/191.

kokfoongcmcl:
I'm not a huge fan of these sort of manual string manipulation unless it's absolutely necessary, I believe a better way to check that it's a valid IRI is to use the java.net.URI class. Under what circumstances do you expect IRI with the angled brackets?

void checkIfValidIri(String s, String p) {
		if (!trimIRI(s).matches(HTTP_HTTPS_PROTOCOL) || !trimIRI(p).matches(HTTP_HTTPS_PROTOCOL)) {
			throw new JPSRuntimeException(INVALID_IRI_ERROR + s + ", and " + p);
		}
	}

	String trimIRI(String iri) {
		if (iri.startsWith("<")) {
			iri = iri.substring(1);
		}
		if (iri.endsWith(">")) {
			iri = iri.substring(0, iri.length() - 1);
		}
		return iri;
	}

kokfoongcmcl:
also FYI http and https are not the only ones that can be IRIs - https://en.wikipedia.org/wiki/File_URI_scheme

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationSparql.java
```diff
@@ -1578,52 +2348,338 @@ boolean validatePureInputs() {
...
+						d.has(PropertyPaths.path(isDerivedFrom, zeroOrOne(belongsTo)), ups),
+						ups.has(PropertyPaths.path(hasTime, inTimePosition, numericPosition), upsTs))
+				.filter(tsFilter);
+		// this GraphPattern queries all the old outputs (entities)
+		GraphPattern entityPattern = GraphPatterns.and(e.has(belongsTo, d), e.has(p1, o), s.has(p2, e).optional());
+		// these patterns query the status of this derivation if it exist, thus optional
+		TriplePattern tp1 = d.has(hasStatus, status);
+		TriplePattern tp2 = status.isA(statusType);
+		TriplePattern tp3 = status.has(hasNewDerivedIRI, sndIRI);
+		GraphPattern statusQueryPattern = GraphPatterns.and(tp1, tp2, tp3.optional()).optional();
```
kokfoongcmcl, line 2480:
```suggestion
		GraphPattern statusQueryPattern = GraphPatterns.and(tp1, tp2, tp3).optional();
```

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationOutputs.java
```diff
@@ -0,0 +1,273 @@
...
+		if (iri.startsWith("<")) {
+			iri = iri.substring(1);
+		}
+		if (iri.endsWith(">")) {
+			iri = iri.substring(0, iri.length() - 1);
+		}
+		return iri;
+	}
+
+	String getClassName(String rdfType) {
```
kokfoongcmcl, line 257: I'm not sure if it's a good idea to extract a class name like this

kokfoongcmcl:
Design/function wise I don't major concerns, majority of the SPARQL updates are updated to be robust against duplicate requests except monitorAsyncDerivations (the one I managed to spot) which is optional to fix in this PR.

My main concern is that the code is getting more complicated, less user friendly and difficult to understand, e.g.
- Mixture of deprecated/non-deprecated functions
- When someone comes in and try to understand what the code is doing, it is not clear at first glance that DerivationAgent is also executing some important SPARQL updates, as it is outside the derivation package

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationOutputs.java
```diff
@@ -0,0 +1,273 @@
...
+			}
+		}
+		this.oldEntitiesDownstreamDerivationMap = map;
+	}
+
+	////////////////////////////////////////////////////////////////////
+	// Create new entities and add triples - to be used by developers //
+	////////////////////////////////////////////////////////////////////
+
+	public void createNewEntity(String iri, String rdfType) {
```
kokfoongcmcl, line 103: The derivation framework does not support multiple IRIs with the same rdf:type, I think an exception should be thrown here if there's a duplicate

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationSparql.java
jb2197: I think we should remove `TriplePattern tp3 = status.has(hasNewDerivedIRI, sndIRI);` as it is a triple that would appear for asynchronous derivation only, NOT for synchronous derivation. As here we are only calling reconnectNewDerivedIRI when updating synchronous derivations, we can just ignore that.

jb2197: This is addressed in https://github.com/cambridge-cares/TheWorldAvatar/pull/185/commits/b77eb21d7abc361942cf2d20bf4bbc233bb41ca3.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationSparql.java
```diff
@@ -1578,52 +2348,338 @@ boolean validatePureInputs() {
...
+		TriplePattern deleteOldTimestamp = unixtimeIRI.has(numericPosition, dTs);
+		TriplePattern insertNewTimestamp = unixtimeIRI.has(numericPosition, retrievedInputsAt);
+
+		// construct the complete SPARQL update, note that some of the insert triples
+		// have already been added at the beginning of this method
+		modify.prefix(p_time, p_derived)
+				.delete(deleteEntityAsSubject, deleteEntityAsObject, tp1, tp2, tp3, deleteOldTimestamp)
+				.insert(insertNewTimestamp).where(sub);
+
+		try {
```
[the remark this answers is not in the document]
jb2197: Thanks for the suggestion, it is addressed in https://github.com/cambridge-cares/TheWorldAvatar/pull/185/commits/b77eb21d7abc361942cf2d20bf4bbc233bb41ca3.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationOutputs.java
jb2197: As we discussed, the framework works fine as long as the outputs are not inputs to another derivation. So I will leave it as it is now.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationSparql.java
```diff
@@ -1578,52 +2348,356 @@ boolean validatePureInputs() {
...
+
+		// construct the complete SPARQL update, note that some of the insert triples
+		// have already been added at the beginning of this method
+		modify.prefix(p_time, p_derived)
+				.delete(deleteEntityAsSubject, deleteEntityAsObject, tp1, tp2, deleteOldTimestamp)
+				.insert(insertNewTimestamp).where(sub);
+
+		try {
+			if (storeClient.getClass() == RemoteStoreClient.class) {
+				if (((RemoteStoreClient) storeClient).isUpdateEndpointBlazegraphBackended()) {
```
kokfoongcmcl, line 2505:
```suggestion
				if (storeClient.isUpdateEndpointBlazegraphBackended()) {
```

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/derivation/DerivationSparql.java
```diff
@@ -1578,52 +2348,356 @@ boolean validatePureInputs() {
...
+					HttpResponse httpResponse = ((RemoteStoreClient) storeClient)
```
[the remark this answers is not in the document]
jb2197: As discussed, this will not compile. So I will leave it as it is now.

kokfoongcmcl (approved):

sm453 (approved):

PR #190, opened 2022-04-25
Dev chemaboxwriters

dln22:


sm453 (approved):
In ontology-tools/CMCLABoxManagement/chemaboxwriters/chemaboxwriters/common/abox_stages.py, line 14, you're defining a function that returns something, but you're declaring a return type hint of None...

gpeb2 (approved):

PR #205, opened 2022-05-12
Dev met office agent

sm453:


Agents/MetOfficeAgent/metoffice/flaskapp/inputtasks/routes.py
```diff
@@ -0,0 +1,105 @@
...
+
+# Initialise logger
+#logger = agentlogging.get_logger("prod")
+
+inputtasks_bp = Blueprint(
+    'inputtasks_bp', __name__
+)
+
+# Define route for API request to instantiate all stations
+@inputtasks_bp.route('/api/metofficeagent/instantiate/stations', methods=['GET'])
```
gpeb2, line 23: We need to have a wider discussion about the blanket use of the "GET" method for every route in every agent. For now I think it is fine to stick with it though as it is the current convention in the TWA.

Agents/MetOfficeAgent/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "3.8"
+
+services:
+  metoffice_agent_debug:
```
gpeb2, line 4: There are standard ways to removed the need for duplication in docker-compose files but I'm not sure it's worth changing things at this point.

Agents/MetOfficeAgent/resources/metoffice.properties
```diff
@@ -0,0 +1,34 @@
+###--- Properties for MetOffice Agent agent ---###
```
gpeb2, line 1: There should be a wider discussion about properties files as it should be possible to populate most of the information when starting the agent's container. What's here is fine for now though.

gpeb2 (approved):
I've added a few comments, none of them require action but I just wanted to add my thoughts. The only other comment would be that in the future the files for the visualisation will also need to be served through an endpoint but how to do this needs to be discussed.

PR #206, opened 2022-05-16
158-hotfix-agentlogging-isatty: hotfix for error "'StreamToLogger' ob…

jb2197:
…ject has no attribute 'isatty'" when running pytest with docker-compose option.


mdhillmancmcl (approved):
Assuming this change doesn't break anything in the codes using the logging library (I doubt it does), then I don't see any issues with this.

Note: in the future we will have to add unit tests to check changes in the logging library don't break anything, but for now the developer making the change will have to check manually. As a reviewer I would ideally like to test this too, but I do not currently have the time for that.

PR #209, opened 2022-05-19
hotfix-marie-frontend

mdhillmancmcl:
Trivial check in the frontend web page to see if requests to the Marie backend return a null response, showing an error label if they do.


kokfoongcmcl (approved):
I've tested this and it works

PR #211, opened 2022-05-22
1180 dev marie simplify google module

xiaochizhou:
the OntoSpecies mapping is added to Marie agent module, which provides the information about the update IRIs of the species

This commit is tested 20 mins before CLAUDIUS goes down. From the test we know that the backend is working for the example questions listed on the paper. The list of questions will be sent to the Skype group. However, what we don't know is whether the **front-end** is able to **visualize** the backend response, which might be a major issue on Monday.


sm453 (approved):
I can see code that has no hope of being sustainable, such as hard-coded species names and "old-to-new" IRI mappings, but I guess now is not the time to sort this out. I'm approving this in the interest of getting on with it...

PR #212, opened 2022-05-22
152 dev pyderivationagent

jb2197:
This PR follows up on the upgrade of the derivation framework in https://github.com/cambridge-cares/TheWorldAvatar/pull/185 and provides a python package for developing derivation agents (updated from pyasyncagent to pyderivationagent). 

Within the development, a few changes were made to the derivation framework on the Java side to make it compatible with py4jps (py4j).

A set of dockerised integration test examples was also provided using the `.env` files to configure the agents deployed within the same docker stack with the triple store. 

New version of py4jps and pyderivationagent will be released once the PR is approved.


sm453 (approved):
Looks ok to me, but it appears we may need to revisit our Java vs. Python philosophy at some point...

PR #218, opened 2022-05-27
Dev agent routing

csl37:


csl37:
CachedRouter

csl37:
Pending creation of ontoagent triple store 

**To do before merge**: update version number

sm453 (approved):
In AgentRouter.getQuery, getting the agent URL by filtering the agent name out of its IRI string is a bit of a disaster. Trying to extract semantics out of an IRI is kind of the cardinal sin of the semantic web, but I'm guessing this is simply a symptom of the inadequacy of OntoAgent, right? If so, then I wouldn't try to improve this here, because solving problems with OntoAgent is clearly out of scope of this PR.

csl37:
> In AgentRouter.getQuery, getting the agent URL by filtering the agent name out of its IRI string is a bit of a disaster. Trying to extract semantics out of an IRI is kind of the cardinal sin of the semantic web, but I'm guessing this is simply a symptom of the inadequacy of OntoAgent, right? If so, then I wouldn't try to improve this here, because solving problems with OntoAgent is clearly out of scope of this PR.

Correct. I think we would need to extend the ontology to include a label -- similar to ontokgrouter.

PR #219, opened 2022-05-30
Updated Derivation Readme.md

markushofmeister:
Incorporated updates:
1) Ensure resolvability of OntoDerivation.owl
2) Fixed a few minor typos

Pending question: Shall ontology IRIs in OntoDerivation.owl get updated as we no longer use a "develop" branch? Might become "main" or even moved to theworldavatar.com


sm453:
> Pending question: Shall ontology IRIs in OntoDerivation.owl get updated as we no longer use a "develop" branch?

Yes, please. I also agree with using raw.githubusercontent.com rather than github.com. However, not as part of this pull request, please, as it has wider implications.

jb2197 (approved):

PR #220, opened 2022-06-01
Incorporate ontology changes to OntoBuiltEnv

markushofmeister:
Updated OntoBuiltEnv TBox and also included instances as Abox as discussed. The latest MIRO board for comparison can be found here: https://miro.com/app/board/uXjVOIznYOU=/?share_link_id=626431683477 (I believe you still have the edit link)


markushofmeister:
After talking to Mei Qi and Srishti, I realized that the current `range` of `hasOntoCityGMLRepresentation` should actually not be a regular `string`. This relationship connects a (OntoBuiltEnv) building instance with its OntoCityGML "equivalent". The OntoCityGML "equivalent" is also an instance which is further connected to geometric information. As shown in the screenshot below, however, this instance has neither a `rdf:type` nor `datatype` associated with it. To still make it "resolvable" as IRI for any further queries in the KG, I changed `string` to `anyURI`. Please review whether that makes sense.

image

PR #224, opened 2022-06-10
dev-remove-KnowledgeRepository: removed dependency on com.blazegraph …

kokfoongcmcl:
I've removed the KnowledgeRepository class from base lib because it is causing errors when tomcat is started without a blazegraph instance in it. As far as I know, nobody is using this functionality from the base lib.


sm453:
In the removed code, there are a number of functions that could be quite useful. Is there equivalent code, or code that supersedes it elsewhere? If not, I'm not sure we should just throw this away...

sm453:
In the removed code, there are a number of functions that could be quite useful. Is there equivalent code, or code that supersedes it elsewhere? If not, I'm not sure we should just throw this away...

kokfoongcmcl:
There's a function called uploadFile in the RemoteStoreClient that uses HTTP POST instead. I believe this class is duplicated from somewhere else.

msff2 (commented):
Move the deleted code to the JPS_BLAZEGRAPH project so that people who use this service to upload data to any instance of Blazegraph still can do that.

msff2:
I have tested the data uploader by importing four OWL files, the size of each more than 100 MBs. It has shown good performance against Blazegraph and RDF4J.

PR #227, opened 2022-06-17
Dev air quality agent

markushofmeister:
Agent to continuously instantiate/update latest air quality data across the UK and create output files for DTVF. Current updating frequency once per day, but can easily be increased/amended by different frequency of job scheduler within Docker container. 
More frequent instantiation or output file creation also to be trigger by HTTP requests when needed.

One remaining observed issue is the correct display of the respective units in the time series panel, i.e. µg/m3.
When creating the output files outside of Docker (i.e. locally run script or locally deployed agent) the units get displayed correctly. In case the output files are created from within Docker, the units get messed up. 

*Created outside Docker*:
unit representation in timeseries json file: `"units": [ "\u03bcg/m3" ]`
image

*Created from within Docker*:
unit representation in timeseries json file: `"units": [ "\u00ce\u00bcg/m3" ]`
image


gpeb2:
> 

It looks like a difference in the encoding Unicode vs UTF-8, see https://en.wikipedia.org/wiki/Mu_(letter)#Character_encodings

markushofmeister:
> > 
> 
> It looks like a difference in the encoding Unicode vs UTF-8, see https://en.wikipedia.org/wiki/Mu_(letter)#Character_encodings

I tried several things now, and "simply" changing the encoding doesn't seem to help. I debugged a little and figured out that the error likely occurs when writing the unit symbol `μg/m3` into Blazegraph when running the agent and Blazegraph within Docker. I tried to replace the unit `μg/m3` with its UTF-8 or UTF-16 representation, but also then the populated unit within Blazegraph is not what we want. Could setting the locale/encoding for the Blazegraph image within the container help? And if so, how could that be done?

gpeb2:
> > > 
> > 
> > 
> > It looks like a difference in the encoding Unicode vs UTF-8, see https://en.wikipedia.org/wiki/Mu_(letter)#Character_encodings
> 
> I tried several things now, and "simply" changing the encoding doesn't seem to help. I debugged a little and figured out that the error likely occurs when writing the unit symbol `μg/m3` into Blazegraph when running the agent and Blazegraph within Docker. I tried to replace the unit `μg/m3` with its UTF-8 or UTF-16 representation, but also then the populated unit within Blazegraph is not what we want. Could setting the locale/encoding for the Blazegraph image within the container help? And if so, how could that be done?

Not sure how helpful this is but I think it's probably relevant https://github.com/blazegraph/database/wiki/Unicode.

Agents/AirQualityAgent/airquality/dataretrieval/stations.py
```diff
@@ -0,0 +1,267 @@
...
+    #                          observation_types=['PM10Concentration'])
+
+    # Data retrieval with geospatial search require a geospatially-enabled
+    # Blazegraph namespace to work successfully
+    # create_json_output_files('C:\TheWorldAvatar-git\Agents\AirQualityAgent\output',
+    #                          circle_center='52.75#0.4', circle_radius='100')
+
+    # create_json_output_files('C:\TheWorldAvatar-git\Agents\AirQualityAgent\output',
+    #                          circle_center='52.75#0.4', circle_radius='100',
+    #                          observation_types=['PM10Concentration'])
```
gpeb2, line 267: I assume this block is just for debugging/manual testing, is that correct?

gpeb2 (approved):
Assuming what I said in my comment is correct then this looks fine. If you don't already, I would suggest you open this in VSCode with the sonarlint linter enabled as there are a few minor suggestions mainly regarding exceptions that might make the code a little bit nicer.

Agents/AirQualityAgent/airquality/dataretrieval/stations.py
markushofmeister: Yes, this code is only reached when executed as main script (i.e. called manually)

markushofmeister:
As this agent in its current design is not working with/within the stack, will the FeatureInfoAgent still be able to visualise and extract the data? Or shall the agent be re-factored to work within the same stack as all other agents?

PR #228, opened 2022-06-17
dev-ssh-in-docker: removing session.setConfig("PreferredAuthenticatio…

kokfoongcmcl:
The code was failing to authenticate while executed from Docker previously, for some weird reasons, removing the lines mentioned in the commit seems to have solved this.

Tested that Pageant option still works


sm453:
If you do decide to keep this change, can you please also update the comments? That is, not only should the no-longer-accurate current comments be removed, but you may consider adding something helpful, e.g. "NB We do not set the preferred authentication methods for the session here, because this causes problem X when run inside a container.".

sm453 (approved):

PR #230, opened 2022-06-20
1348 dev refactor admsagent code and python scripts

gmmajalCARES:
1) The ADMSAgent has been refactored.
2) validateInput method has been added. Mock values have been added in the ADMSAgent class and the createEmissionInput method has been merged into one method.
3) Unit tests have been added for the ADMSAgent.
4) Mock values have been added in the python scripts associated with the ADMSAgent class. 
5) Both the 'ship' and 'plant' side of the python scripts have been tested with and without mocked values. The generated .apl files with and without mocks were checked to see if the values were identical. The .apl files were also verified within ADMS as well.


kokfoongcmcl:
I'll review this in detail after it is containerised. One thing I noticed in the folder structure (you probably copied it from the old code), the Java code folder should be something like src/main/java/..., if it compiles then I have no complains

PR #236, opened 2022-06-28
Merge branch dev-OntoChemPlant-update into main

sganguly-cares:


msff2 (commented):
For some reason, the OWL file does not include any Functional Properties; therefore, I have run the TBoxGenerator to convert the OntoChemPlant CSV file into OWL and committed and pushed the OWL file to the Git repository. You will now see all Functional properties are there.

Please fix the following issue;

- Convert the first letter of each word of hasyearofEstablishment into a capital letter except the first one, so it should be hasYearOfEstablishment after modification. Also, make sure that all Functional properties are regenerated.

msff2 (approved):

PR #241, opened 2022-07-08
164 dev create sync derivations for new info

jb2197:
This PR adds the function to create synchronous derivations for new information (new output instances) on the fly. DerivationAsynExample was updated to provide integration tests. All unit tests (including those added in this PR) and all integration tests passed.


kokfoongcmcl:
Or maybe it just needs some minor clarification/arrangement

kokfoongcmcl:
Also one more thing, now there are two places that sends the HTTP request with the "query" parameter, if there is no clean solution now, I'd at least put a comment in both places

sm453 (approved):

PR #244, opened 2022-07-13
Dev add timeseries postgis support

kokfoongcmcl:
Added ability to store geometries as time series, key changes:
1) the class to provide during initialisation will be one of the classes that extends org.postgis.Geometry, this will create a column with the geometry type
2) added getValuesAsPoint in the TimeSeries class, can easily add for the other polygon types if we ever need them
3) added integration test using the PostGIS test container
4) populateTimeSeriesTable no longer uses jOOQ's native execute() to upload data, updates are executed using java.sql instead because PostGIS is not supported in the open source edition

### DIFF
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -477,9 +477,14 @@
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
-            <version>2.6</version>
+            <version>2.11.0</version>
+        </dependency>
+
+        <!-- used in time series client to upload postgis geometries -->
+        <dependency>
+            <groupId>org.postgis</groupId>
+            <artifactId>postgis-jdbc</artifactId>
+            <version>1.3.3</version>
         </dependency>
     </dependencies>
-
-
 </project>
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -1,32 +1,30 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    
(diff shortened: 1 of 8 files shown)


gpeb2:
This looks good for now. Depending on what we what to do with geometry data in time series in the future two things that would be good to add are:

1. adding the specific data type and the SRID to the column type, this should prevent issues in the future (https://gis.stackexchange.com/a/153988);
2. adding indexing to improve performance, this is probably true of the time series tables in general.

markushofmeister (commented):
Looks good to me; however, I havent tested it. One final question (to which the answer likely is "yes"): Will the `org.postgis.Geometry` class also be available in Python via py4jps? This might be relevant as increasing amount of people are using Python for agent development

kokfoongcmcl:
It's an external class, does it need to be explicitly included in py4jps?

markushofmeister:
> It's an external class, does it need to be explicitly included in py4jps?

I know that this is an external class, but I do not know whether this requires special handling. Currently I can use the `Instant` class via `jpsBaseLibView.java.time.Instant`. If we expect the geospatial features to be available similarly, then I am fine with it.

kokfoongcmcl:
I think it should be included the next time someone builds it, as the pom downloads all the dependencies

gpeb2:
From my limited understanding of how py4j works I don't think you need to explicitly specify the available classes anywhere but I might be wrong.

kokfoongcmcl:
> This looks good for now. Depending on what we what to do with geometry data in time series in the future two things that would be good to add are:
> 
> 1. adding the specific data type and the SRID to the column type, this should prevent issues in the future (https://gis.stackexchange.com/a/153988);
> 2. adding indexing to improve performance, this is probably true of the time series tables in general.

Point 1 is should addressed by [abf5617](https://github.com/cambridge-cares/TheWorldAvatar/pull/244/commits/abf5617823ce9322487a74fc6b761c6c9782f095).

There is an additional argument to specify the srid in the initTimeSeriesTable method. Geometry columns are now created with restrictions on the provided geometry type and srid (if provided).

Point 2 can be addressed in a separate PR

gpeb2 (approved):
The addition of the specific geometry type and the SRID looks good.
I agree that the addition of indexing should be done in a separate pull request.

PR #247, opened 2022-07-15
dev-add-timeseries-index: create index for all columns in the central…

kokfoongcmcl:
added index for the central dbTable and time columns


kokfoongcmcl:
tested this with a database with 5000 tables, only managed to get 5% improvement, better than nothing. Given that the speed improvement is not huge, no code is provided to add indexes to an existing time series database

gpeb2 (approved):
The indices don't seem to make much difference but might do with larger databases. Changing from the default indexing method might improve things but this can be looked into in the future if performance becomes an issue.

PR #248, opened 2022-07-15
Dev wind and solar sensor agents

msff2:


sm453 (changes requested):
A few points:
1) The file Agents/RenewableEnergyAgents/dtvf_visualisation/README.md appears to be just the readme file of the example visualisation copy-pasted. Please make it specific to what _this_ visualisation is doing.
2) The two files Agents/RenewableEnergyAgents/dtvf_visualisation/index.html and Agents/RenewableEnergyAgents/resources/renewable_energy_agents.properties appear to contain Mapbox API keys. In both files, please replace the actual key with "<PUT_YOUR_KEY_HERE>", and add a reminder to the main readme file(s) that such a key needs to be added to both of these files in order for the visualisation to work.
3) It appears that a number of RDF4J dependencies have been added to JPS_BASE_LIB/pom.xml. What is the reason for that? We need to make every effort not to add unnecessary dependencies to the baselib, especially not ones that tie us to specific triple stores.

msff2:
I have edited the index and property files and the relevant readme files to improve the project and documentation further, as indicated in the three points.

sm453 (approved):
I note that after the merge from main the base lib pom file is no longer shown as modified. The RDF4J dependencies had already been added on the main branch. However, some of them appear unnecessary - they shall be removed on another branch/pull request.

PR #249, opened 2022-07-18
1385-dev-create-separate-library-for-ogm: inital commit for adding og…

mehalaga:
…m as a separate library


sm453 (changes requested):
The config property resource file appears to contain an Ordnance Survey API key. Can you please remove this?

PR #252, opened 2022-07-19
Dev postgres derivation example

kokfoongcmcl:


sm453 (changes requested):
While I'm fine with the other changes (assuming that the authentication is still secure), I strongly disagree with the changes regarding the password. It may not be externally facing, but it is an _example_. An important purpose of examples is to demonstrate good practice (how things should be done), because people do copy-paste from examples without bothering too much. Putting a plain-text password into a (version controlled) docker-compose file is most definitely not good practice. Please revert this change.

sm453:
> I _think_ the default is md5, when I explicitly specified it to use md5, the issue also disappear. As this is just a container within an example stack, I don't think security is an issue here?

md5 is insecure. That means we need to ensure that the container is not reachable from outside... Same comment as on the password issue above: For the example itself it's not important. Let's just hope that no-one copy-pastes this into a production container... At the very least you may want to add a warning comment to the docker-compose file that for production deployment, people may want to ask themselves whether this level of encryption is sufficient for their purposes...

jb2197 (commented):
I have run all integration tests of DerivationExample and DerivationAsynExample and nothing breaks. For the record: the "postgres_data" volume in docker must be deleted before composing up the stack to apply the changes in the authentication. Otherwise, errors will be thrown in the agent containers when TimeSeriesClient is trying to access the database.

kokfoongcmcl:
There's a small chance that this issue might be introduced by a docker update, the code was working when it's deployed on my windows tomcat, accessing postgres installed on windows

jb2197:
> There's a small chance that this issue might be introduced by a docker update, the code was working when it's deployed on my windows tomcat, accessing postgres installed on windows

I think this is possible, the error also only surfaces on my machine after updating docker-desktop to v4.10.1 (from v4.0.0).

sm453:
> For the record: the "postgres_data" volume in docker must be deleted before composing up the stack to apply the changes in the authentication. Otherwise, errors will be thrown in the agent containers when TimeSeriesClient is trying to access the database.

This may be worth including into the comment in the docker-compose file...

kokfoongcmcl:
I think this is solved by updating the postgres driver in base lib. Passwords are no longer committed and specified through a secret file

jb2197 (approved):
Tested from my side, and all integration tests working fine.

sm453:
Please also revert the changes to the readme file (at least the ones relating to the blazegraph password.

PR #261, opened 2022-08-02
1397 remove impexp client dependency from ogm

mehalaga:


sm453 (approved):

sm453 (approved):
Just to note for the record: The geo models are not an essential part of the core OGM library, so we have decided to remove them. This resolves a CitiesKG dependency issue.

PR #262, opened 2022-08-03
Dev add restriction, domain and range together

msff2:


sm453 (changes requested):
The quantifier argument of the addDomain method now appears unused. Can/should this be removed?

sm453 (approved):

PR #287, opened 2022-08-15
286 make it easier to add generated partially random iris to the postgres tables

gpeb2:


kokfoongcmcl:
Use the sql string directly if there's no "@"?

kokfoongcmcl (approved):
tested the functionality

PR #293, opened 2022-08-19
281 make it easier to spin up multiple stacks

gpeb2:


kokfoongcmcl:
Also justify the recommended extensions to install?

gpeb2:
This pull request adds the following improvements:

1. The external port through which the Nginx container in each stack can now be specified as an additional argument after the stack name when running "./stack.sh start ..." for the stack-manager project.
2. Datasets can now include other Datasets by listing their names under a "externalDatasets" node.
3. If there is a Dataset with the same name as the the stack being started then only that Dataset and any of its "externalDatasets" will be loaded, otherwise all datasets will be loaded.
4. When debugging via VSCode the debug port is now generated automatically to reduce the probability that the port is already in use. This requires the "augustocdias.tasks-shell-input" VSCode extension so it has been added to the dynamic-stack workspace's recommended extensions list.
5. Several other required or highly recommended VSCode extension are now listed in the dynamic-stack workspace's config file. These include extensions for building and testing Java as well as a Java linting extension that highlights and explains potential issues with the code.

kokfoongcmcl (approved):
tested

PR #294, opened 2022-08-19
Dev weather agent in stack

kokfoongcmcl:
Integration to the stack is now completed. When creating stations, the coordinates are created in PostGIS using the GDAL client from stack-clients, an ontop mapping is created for the coordinates.

However writing tests is no longer straightforward and disabled for now, we'll need some sort of "mock stack" in order to write proper tests.

Updates to the APIs are reflected in the README


sm453:
It appears you have replaced all JPSAgents with HttpServlets. Why?

kokfoongcmcl:
It's because I think PUT and DELETE are more appropriate, JPSAgent does everything as GET and also forces inputs to be JSON objects

kokfoongcmcl:
There is an issue with underscore in the host name when the container is spun up using docker compose (through the stack.sh script), George is enabling spinning containers via the stack manager which should solve this issue.

sm453 (approved):
Can you post a new issue to upgrade JPSAgent to support PUT and DELETE requests? (I have a suspicion that this may already exist, so please check.)

kokfoongcmcl:
PUT exists, but there are a number of issues. 
- A route should only accept a specific list of functions, throws the correct exception when it receives a request it does not support. 
- The way it's currently designed, submitting a PUT/POST request to an agent will result in the exact same behaviour. 
- To me it only makes sense to have a framework when there are accompanying AgentInput and AgentOutput classes for each agent class, like the DerivationAgent. At the moment by just using JSON objects it's not too different from a generic API.
- If we develop a proper framework we might need to ban python agents, or maybe it's not too bad to just duplicate the base classes.

I think this needs careful thought and planning instead of extending the current classes.

PR #303, opened 2022-08-24
1319 dev implement caching for store router

csl37:
1. Implements cache for StoreRouter by extending AbstractCachedRouter class
2. Improves method of getting endpoints by reducing this to a single sparql query


JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/StoreRouter.java
```diff
@@ -163,6 +197,53 @@ public static StoreClientInterface getStoreClient(String targetResourceID, boole
...
+	/**
+	 * Get store client for ontokgrouter
+	 */
+	@Override
+	public StoreClientInterface getRouterStoreClient() {
+		return new RemoteStoreClient(storeRouterEndpoint);
+	}
+	
+	@Override
+	public List<String> getFromStore(String targetResourceLabel, StoreClientInterface storeClient){
```
msff2, line 209: What does it get from the store? I think a comment is required here to let users of this public method know what they will get if they call it.

msff2 (approved):

PR #305, opened 2022-08-26
297 dev flood agent in stack

kokfoongcmcl:
Key changes:
1) Endpoints for databases are obtained using clients from stack-clients
2) Coordinates are stored in PostGIS with an ontop mapping
3) Geoserver layer created for visualisation
4) Test disabled until we have a mock stack


gpeb2 (approved):

markushofmeister:
I believe the agent works; however, I have a few comments / clarification questions:

1) It could be made clearer that the first `stack.sh start flood-stack` should be run from `Deploy/stacks/dynamic/stack-manager/`
2) It could be made clearer that the java agent commands, e.g. `java -cp FloodAgent-2.0.0-SNAPSHOT.jar uk.ac.cam.cares.jps.agent.flood.InitialiseStations` need to be run within the agent's Docker console
3) Spinning up the agent using `stack.sh start flood-stack` seems to work without providing the mentioned `datum.json` (and `river_stations.csv`) files. Although folders with those names seem to get created when spinning up the agent
4) Spinning up the agent using `stack.sh start flood-stack` already creates 178,361 triples. Is that expected? If so, it might be work clarifying what they are
5) Running `java -cp FloodAgent-2.0.0-SNAPSHOT.jar uk.ac.cam.cares.jps.agent.flood.InitialiseStations` just after spinning up the agent causes a JPSRuntimeException for me. Is that expected? Might be as it is not clear to me what the initialisation vs. the just spinning up actually does

kokfoongcmcl:
> I believe the agent works; however, I have a few comments / clarification questions:
> 
> 1. It could be made clearer that the first `stack.sh start flood-stack` should be run from `Deploy/stacks/dynamic/stack-manager/`
> 2. It could be made clearer that the java agent commands, e.g. `java -cp FloodAgent-2.0.0-SNAPSHOT.jar uk.ac.cam.cares.jps.agent.flood.InitialiseStations` need to be run within the agent's Docker console
> 3. Spinning up the agent using `stack.sh start flood-stack` seems to work without providing the mentioned `datum.json` (and `river_stations.csv`) files. Although folders with those names seem to get created when spinning up the agent
> 4. Spinning up the agent using `stack.sh start flood-stack` already creates 178,361 triples. Is that expected? If so, it might be work clarifying what they are
> 5. Running `java -cp FloodAgent-2.0.0-SNAPSHOT.jar uk.ac.cam.cares.jps.agent.flood.InitialiseStations` just after spinning up the agent causes a JPSRuntimeException for me. Is that expected? Might be as it is not clear to me what the initialisation vs. the just spinning up actually does

1. It's already there? I added quotes around them to highlight it
2. Added in readme, though this is not required
3. The folders are created because they are defined in the docker volumes and the files don't exist
4. The ./stack.sh start command is all that is needed, it instantiates everything.
5. Standard usage does need require this, so not important

markushofmeister:
> 4. The ./stack.sh start command is all that is needed, it instantiates everything.
> 5. Standard usage does need require this, so not important
I am not saying that this is wrong. Just saying that it wasn't clear to me what each call / method does and some explanation along them in the README might have clarified that.

markushofmeister (approved):

PR #311, opened 2022-09-01
Added vscode launch.json for remote debugging

markushofmeister:
Added launch.json configuration for remote debugging within docker container for non agent not running within the stack


gpeb2 (approved):
I'm not familiar with the Python options but this look fine.

PR #313, opened 2022-09-01
Dev add met office agent run config on stack

gpeb2:
This should allow you to use the VSCode Debug configurations with the MetOfficeAgent.
The debug image still has the "debug" prefix but the release one doesn't any more for consistency with the other agents.
The hotpatch functionality works for me.
The debug version will run when built and launched through VSCode and the production version can be built by running `./stack.sh build` in a terminal, and then run by running `./stack.sh start <STACK NAME>`, where `<STACK NAME>` is the name an already running stack.
I've also added references to the Docker Swarm "confgs" and "secrets" to the docker-compose.yml file so you should be able to call the "readEndpointConfig" method. I should be able to remove the Java Class argument from this method to simplify things in the future.


markushofmeister (changes requested):
I would recommend that we close PR #311 first where we merge the run configs for remote debugging outside the stack to the agent first.
Afterwards, I would recommend to merge these changes into branch `dev-MetOfficeAgent-withinStack`, where I already started on the stack-compliant version of the agent. This feels cleaner to me and would leave us with a properly working agent snapshot with all supporting files for running it outside the stack before adding everything stack related.

gpeb2:
This branch has been merged into the "dev-MetOfficeAgent-withinStack" branch.

PR #315, opened 2022-09-02
1386 investigate remotestore client for rdb access

mehalaga:


mehalaga:
Summary of changes to Timeseries classes:
1. RDBStoreClient: this class accepts the credentials and the url required to create a connection to RDB. It has a get connection method which can be used to obtain a connection object by classes that are intended to work with relational databases. It has getter and setter methods for the credentials.
2. TimeSeriesRDBClient has been modified to accept the rdbstoreClient object in its constructor which contains the RDB credentials. In the connect method, the instance of the RDBStoreClient is used to get the connection object from the getConnection method in RDBStoreClient and hence create the connection to postgresql required by the TimeSeriesRDBClient.
3. TimeSeriesClient: the constructors of this class accept either (1) properties file path from which the credentials are loaded using loadRDBConfigs method or (2) the credentials for connecting to the RDB .. using these credentials an RDBStoreClient object is created which is used to create an instance of the TimeSeriesRDBClient. The loadRDBConfigs method has been shifted from TimeSeriesRDBClient to TimeSeriesClient.

kokfoongcmcl:
We've had a bit of discussion here and to address the issue of closing connections properly, we want to change the way we use the TimeSeriesClient slightly, the idea is to not store the connection object in the client, and to make use of Java's try-with-resources https://www.baeldung.com/java-try-with-resources. Java will automatically close the connection this way.

Something like the following:
```
RDBStoreClient rdbStoreClient = new RDBStoreClient(url, user, password)
try (Connection conn = rdbStoreClient.getConnection())
    TimeSeries ts = TimeSeriesClient.getTimeSeriesData(lowerbound, upperbound, conn)
    TimeSeriesClient.addTimeSeriesData(ts, conn)
    // and do other things
}
```
Here, the getConnection method should always create a new object. Another thing you need to do is to add an extra argument to all the methods currently in TimeSeriesClient/TimeSeriesRDBClient to take an extra connection argument. If we want to keep backwards compability, you can mark all the old methods that stores the connection object as deprecated for a few months before removing them. 

You'll need to update the Javadocs as well as the unit tests to show proper usage of the TimeSeriesClient. If anything is unclear, get in touch with me on Skype.

kokfoongcmcl:
Thanks for the update. Please also add some comments in the classes on how they're supposed to be used, also update the readme on how to use this

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/RemoteRDBStoreClient.java
```diff
@@ -0,0 +1,114 @@
...
+    public Connection getConnection(){
+        try {
+            Class.forName("org.postgresql.Driver");
+            conn = DriverManager.getConnection(this.rdbURL, this.rdbUser, this.rdbPassword);
+            stmt = conn.createStatement(java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+        } catch (SQLException | ClassNotFoundException e) {
+            throw new JPSRuntimeException(e.getMessage(), e);
+        }
+        return conn;
+    }
```
kokfoongcmcl, line 55: A better practice here is to catch the exceptions separately and give a different message, e.g.
` catch (ClassNotFoundException e) {
    throw new JPSRuntimeException ("Failed to load postgresql driver",e)"`
    
The information in e.getMessage() is already in "e", so you're probably duplicating information here

mehalaga: sure, will make this change.
Would if make sense to replace it with this catch block:
`catch (Exception e) {
                System.out.println("Connecting failed: " + this.rdbURL);
		throw new JPSRuntimeException(exceptionPrefix + "Establishing database connection failed");
}`
or to handle each exception seperately?

mehalaga: where exceptionPrefix = this.getClass().getSimpleName() + ": "

kokfoongcmcl: It's better to handle each separately, SQLException for connection failure, ClassNotFoundException for Failing to load class

markushofmeister (changes requested):
I had a look at the changes and they look alright to me, although, I cannot judge everything in detail. However, I can run all tests without issues and believe the changes are fine to be merged.

Just a few comments:
1) Are the examples in the README still up to date? I believe they still work, as old methods are only deprecated, but it might be made clearer that they do not follow the updated design and/or we should include examples of the updated design (if we have them already)
2) The design of a connection object as parameter to all RDB Client methods makes sense; however, I am not sure how easy this will be to replicate in Python. Using py4jps to wrap the BASE_LIB is easiest with standard datatype. So, getting this new design working in Python for the first time might need some fiddling with Java Datatypes. That doesn't require any changes, but we should be aware of it.
3) The pom version number of this branch is `1.11.7` while `main` is already `1.16`. So, please increment as necessary before merging and/or merge main into this branch before merging back
4) I am currently unable to build the BASE_LIB as a whole due to failing tests in `testProcessRequestParameters(uk.ac.cam.cares.jps.base.agent.test.JPSAgentTest`. Hopefully that gets solved after merging in main.

kokfoongcmcl:
The error you see with that test is probably from matching the stack trace?

mehalaga:
> 4\. I am currently unable to build the BASE_LIB as a whole due to failing tests in `testProcessRequestParameters(uk.ac.cam.cares.jps.base.agent.test.JPSAgentTest`. Hopefully that gets solved after merging in main.

-> This test has been fixed on a seperate branch and I think it occurred due to the chnage in Java version from 8 to 11. I excluded it and was able to build JPS_BASE_lIB

markushofmeister:
> The error you see with that test is probably from matching the stack trace?
> 
> I believe the way to do it in python is to use try catch finally, where you close the connection in the finally block.

The failing test is unrelated to the TimeSeries changes. I just mentioned it, as it feels wrong to publish new versions of the entire JPS_BASE_LIB, which does not build properly.

Try-catch-finally is always good practice; however, I was referring to the fact that you would now need to properly create/retrieve a Java connection object in Python to call any TimeSeries method. Prior one only needed basic Data types, i.e. strings. Getting such Java Data Types can be tricky initially, but we can do it.

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesRDBClientIntegrationTest.java
```diff
@@ -32,15 +28,14 @@ public class TimeSeriesRDBClientIntegrationTest {
 	// Define RDB database setup (analogous to a triple-store endpoint)
 	// Using special testcontainers URL that will spin up a Postgres Docker container when accessed by a driver
 	// (see: https://www.testcontainers.org/modules/databases/jdbc/). Note: requires Docker to be installed!
-	private static final String dbURL = "jdbc:tc:postgresql:13.3:///timeseries";
+	private static final String dbURL = "jdbc:tc:postgresql:9.6.8:///timeseries";
```
kokfoongcmcl, line 31: Is there a reason for using an older version here?

mehalaga: The reason for changing the version is to make it unique and different from the ones used in the existing tests. This also solves the problem of failing tests and solves the error of a data iri having an existing timeseries instance.

kokfoongcmcl: This is not a good solution. 
It looks by default the testcontainer class tries to reuse an existing container to increase performance. You should stop the container at the end of the test class, there should be more than 1 way to do this.

kokfoongcmcl: There should be a method to clear all time series data, if you use that at in an @AfterClass method that should work too

mehalaga: done

kokfoongcmcl (approved):

kokfoongcmcl:
Please remember to merge from main and update the version number, and deploy it

markushofmeister (approved):

markushofmeister:
Please note that main got recent pushes yesterday, including increasing JPS_BASE_LIB version number. So, please re-merge main before merging and increasing the version number again

PR #317, opened 2022-09-02
Dev 1285 upgrade from java 8

gpeb2:
The intention of this pull request is to make it possible to build and run the TWA codes and agents with a Java 11 JDK/JRE.
This has been made necessary due to at least one dependency now being compiled to Java 11 bytecode.
Actually updating the TWA code to compile to Java 11 bytecode will require updating some of the key dependencies such as Spring.
Updating to Java 11 or ideally Java 17 bytecode would allow us to use newly added Java functionality as well as greatly improved security.

### DIFF
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -10,7 +10,7 @@
     details on how version numbers should be selected -->
 
 
-    <version>1.15.0</version>
+    <version>1.16.0-java11-SNAPSHOT</version>
 
     <!-- Project Properties -->
     <properties>
@@ -47,6 +47,7 @@
                 <configuration>
                     <source>1.8</source>
                     <target>1.8</target>
+                    <release>8</release>
                 </configuration>
             </plugin>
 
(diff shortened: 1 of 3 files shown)


gpeb2:
So far I've just looked at fixing the JPS base lib (all the tests now run for me). I will look at the agents in the "Agents" folder next. @sganguly-cares or @csl37 can you add a list of agents you think should be able to run in this here and I'll make sure their tests run?

sganguly-cares:
> So far I've just looked at fixing the JPS base lib (all the tests now run for me). I will look at the agents in the "Agents" folder next. @sganguly-cares or @csl37 can you add a list of agents you think should be able to run in this here and I'll make sure their tests run?

All of the agents in that folder should run, given that all projects in the 'main' branch have to be deployable. I have already tested the agents which are deployed on the CARES side, and all of them build fine. I believe the rest are maintained by CMCL/CoMo.
Added a list of agents still left to be tested here (It's under the "Agents" sheet): https://www.dropbox.com/s/8czq6roufkkwvry/Projects%20status%20with%20Java%2011.xlsx?dl=0

csl37 (approved):

gpeb2:
Before merging can you please go through and update the pom.xml files of the agents we've modified with the following changes:

1. Update the version of the jps-base-lib dependencies to 1.17.0
2. Increment the minor version number for each modified agent

PR #320, opened 2022-09-06
281-make-it-easier-to-spin-up-multiple-stacks: follow up

gpeb2:
I forgot to remove the SNAPSHOT qualifier from the parent-pom version number.


gpeb2:
I've already built the artefacts so this just needs merging back in.

kokfoongcmcl (approved):

PR #322, opened 2022-09-07
321-postgisclient-driver-not-found: load postgres driver explicitly a…

kokfoongcmcl:
When PostGISClient is used externally, it does not load the driver automatically
Key changes:
1) load driver explicitly
2) renamed executeQuery to executeUpdate

### DIFF
--- a/Deploy/stacks/dynamic/stack-clients/src/main/java/com/cmclinnovations/stack/clients/postgis/PostGISClient.java
+++ b/Deploy/stacks/dynamic/stack-clients/src/main/java/com/cmclinnovations/stack/clients/postgis/PostGISClient.java
@@ -21,6 +21,11 @@ public class PostGISClient extends ContainerClient {
     }
 
     private Connection getConnection(String database) throws SQLException {
+        try {
+            Class.forName("org.postgresql.Driver");
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
         return DriverManager.getConnection(
                 postgreSQLEndpoint.getJdbcURL(database),
                 postgreSQLEndpoint.getUsername(),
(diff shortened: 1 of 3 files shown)


Deploy/stacks/dynamic/stack-clients/src/main/java/com/cmclinnovations/stack/clients/postgis/PostGISClient.java
```diff
@@ -21,6 +21,11 @@ private Connection getDefaultConnection() throws SQLException {
     }
 
     private Connection getConnection(String database) throws SQLException {
+        try {
+            Class.forName("org.postgresql.Driver");
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
```
gpeb2, line 27: It's good practice to add a message to the new exception explaining what has failed from the point of view of the current function.

Deploy/stacks/dynamic/stack-clients/src/main/java/com/cmclinnovations/stack/clients/postgis/PostGISClient.java
```diff
@@ -57,7 +62,7 @@ public void removeDatabase(String databaseName) {
         }
     }
 
-    public void executeQuery(String databaseName, String sql) {
+    public void executeUpdate(String databaseName, String sql) {
         try (Connection conn = getConnection(databaseName);
                 Statement stmt = conn.createStatement()) {
             stmt.executeUpdate(sql);
```
gpeb2, line 68: Please replace "query" with "update" in the exception message.

gpeb2 (changes requested):

gpeb2 (approved):

PR #331, opened 2022-09-12
Updated geospatial representation of ReportingStations

markushofmeister:
Updated geospatial representation of ReportingStations to be consistent with Ontop mappings. Subproperty classifications and labels as required by FetaureInfoAgent are neglected (for now) as the required data format can be achieved with tweaked SPARQL queries (see example in MetOfficeAgent)


kokfoongcmcl (approved):
Looks ok, just that Seb does not like owl files being committed

PR #338, opened 2022-09-21
22-dev-solar-radiation-extension-in-ontoubemmp

YiKaiTsai:


sm453 (changes requested):
The domain/range definitions of the properties hasPeakPower, hasSpecificYield, and receivesRadiation seem to be incorrect (they should be unions rather than intersections), and unnecessarily complex. These could be much simplified by introducing suitable superclasses, e.g. a SolarEnergyGenerator, and then this could be used as the domain for receivesRadiation.

Several quantities seem to have hourly, daily, monthly, and annual versions. It would be much more elegant to only have single quantities with appropriate physical units attached.

image

sm453:
I'm afraid in your latest commits you have reintroduced a mistake: You have changed the OntoUBEMMP TBox-URL from the previously correct https://www.theworldavatar.com/kg/ontoubemmp/ to an outdated one. Can you please undo this?

sm453:
Regarding hourly, daily, etc. versions of quantities, I guess there is a need for clarification: I believe we must not make the mistake of interpreting semantics into a physical unit. For example, "kWh/day" is a unit of power that is just as good as any other - one must not conclude that it is somehow averaged or taken cumulatively over a day, just because "day" appears in the unit. If we want to encode this quantity as averaged or accumulated over a certain time period, then we should do so using semantic mark-up.

Note the four different, completely independent timescales in this example: a rolling daily average of a power quantity, reported once per minute, in units of kWh/a. Note in particular that the period of averaging is not necessarily the same as the gap between two entries in a timeseries.

sm453 (approved):

PR #339, opened 2022-09-21
15-dev-district-heating-network-instantiation-agent

xuehansong:
…and OWL file for district heating network


markushofmeister (changes requested):
I only checked the ontology in WebVOWL and here are a few comments:
1) I think the Wear Cost (fixed vs. variable) is not represented properly
2) Some of the descriptions of the concepts seem very tailored (maybe they can be kept a little more general, but no priority)
3) The base URI of the ontology and concepts currently is `http://theworldavatar.com/ontology/ontoheatnetwork/OntoHeatNetwork.owl#`. I think we have aligned at some point that out ontologies should follow another syntax, 
 in that case `http://theworldavatar.com/kg/ontoheatnetwork/`. @msff2 can you please confirm? If that is the case then please keep the old name in the .csv file (as the TBoxGenerator will face issues otherwise) and simply string replace the base URIs in the .owl file

markushofmeister:
@xuehansong Does this version already include all my paper based comments (apologies again for that)? What about Mei Qi's comments?

xuehansong:
> @xuehansong Does this version already include all my paper based comments (apologies again for that)? What about Mei Qi's comments?

yes, all yours and Meiqi's are incorporated, except three issues need to clarify with you

markushofmeister (changes requested):
Unfortunately, WebVOWL is down currently and I would like to have a look at the ontology in its visualised form before finally approving it.
I can already suggest a couple of improvements:
1) Something went wrong when creating the `UnitRate` IRI, which is now `UnitRate%20`
2) The base IRI is back to the old format

markushofmeister:
I had another look at the ontology in WebVowl (which is back online again) and have the following comments:
1) Some comments are still off, e.g. comment for `CostInTimeInterval`
2) Relationship for `hasCost` seems weird. It should be `HeatGenerator` - ... - `Cost...`, but it is `Thing` - ...
3) Maybe we should change the naming of `hasObservableProperties` to singular as the subclasses will be instantiated individually (I assume)
4) The definition of `Conctract` seems to have gone wrong as you specify `HeatingNetwork UNION MunicipalUtility UNION HeatBoiler UNION GasTurbine UNION Contract` as range for labels, but never introduce a Class `Contract` as we use an external definition for it 

I suggest you have a look at the ontology in WebVowl in detail yourself and update it accordingly before re-requesting a final review. Thank you!

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
```diff
@@ -0,0 +1,132 @@
+Source,Type,Target,Relation,Domain,Range,Quantifier,Comment,Defined By,Label
+OntoHeatNetwork,TBox,http://theworldavatar.com/ontology/ontoheatnetwork/OntoHeatNetwork.owl,https://www.w3.org/2007/05/powder-s#hasIRI,,,,,,
+OntoHeatNetwork,TBox,1,http://www.w3.org/2002/07/owl#versionInfo,,,,,,
+OntoHeatNetwork,TBox,An ontology defined for representing the district heating network ,http://www.w3.org/2000/01/rdf-schema#comment,,,,,,
+OntoHeatNetwork,TBox,,http://www.w3.org/2002/07/owl#imports,,,,,,
+HeatingNetwork,Class,,,,,,Represents a grid-bound district heating system that transfer heat to residential or commercial buildings,http://www.theworldavatar.com/kg/ontoheatnetwork/,Heating network
+GridConnection,Class,,,,,,"Grid point at which (thermodynamic) properties are observed, i.e., flow and return temperature for heat feed in",http://www.theworldavatar.com/kg/ontoheatnetwork/,Grid connection
```
markushofmeister, line 7: remove "for heat feed in"

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
```diff
@@ -0,0 +1,132 @@
...
+hasMaxAnnualPurchaseVolume,Object Property,,,https://spec.edmcouncil.org/fibo/ontology/FND/Agreements/Contracts/Contract,http://www.ontology-of-units-of-measure.org/resource/om-2/Energy,,A relationship between a contract and maximum quantity to be purchased annually,http://www.theworldavatar.com/kg/ontoheatnetwork/,has maximum annual purchase volume
+hasMaxAnnualPurchaseVolume,Object Property,,,https://spec.edmcouncil.org/fibo/ontology/FND/Agreements/Contracts/Contract,http://www.ontology-of-units-of-measure.org/resource/om-2/Energy,exactly 1,,http://www.theworldavatar.com/kg/ontoheatnetwork/,has maximum annual purchase volume
+isFulfilledBy,Object Property,,,https://spec.edmcouncil.org/fibo/ontology/FND/Agreements/Contracts/Contract,IncinerationPlant UNION HeatProvider,,"A relationship between a contract and the fulfilling party, e.g.., party which provides heat",http://www.theworldavatar.com/kg/ontoheatnetwork/,is fulfilled by
```
markushofmeister, line 112: HeatProvider should be sufficient as Incineration plant is subclass of it

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
```diff
@@ -0,0 +1,132 @@
...
+http://www.ontology-of-units-of-measure.org/resource/om-2/hasUnit,Object Property,,,http://www.theworldavatar.com/ontology/ontoems/OntoEMS.owl#Forecast,http://www.ontology-of-units-of-measure.org/resource/om-2/Unit,exactly 1,,https://www.theworldavatar.com/kg/ontoems/,has unit
+http://www.ontology-of-units-of-measure.org/resource/om-2/hasNumericalValue,Data Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,float,,,http://www.ontology-of-units-of-measure.org/resource/om-2/,has numerical value
+http://www.ontology-of-units-of-measure.org/resource/om-2/symbol,Data Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Unit,string,,,http://www.ontology-of-units-of-measure.org/resource/om-2/,symbol
+http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_performance/economic_performance.owl#hasCost,Object Property,,,HeatGenerator ,CostInTimeInterval UNION http://www.ontology-of-units-of-measure.org/resource/om-2/Cost,,A relationship between heat generator and various cost associated with it,http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_performance/economic_performance.owl,has cost
```
markushofmeister, line 82: Shouldn't cost be sufficient here? CostInTimeInterval is a subclass

markushofmeister (changes requested):

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
markushofmeister: Thanks for the update, we are almost there ;-)

After incorporating the above points, please
1) Check the ontology once more using https://service.tib.eu/webvowl/ to ensure all connections are as expected
2) Ensure that the MIRO board and the ontology are in sync (As Magnus is now also using the ontology, there shouldn't be different versions)

msff2:
I think Lower and Higher CalorificValues are relative. A CalorificValue is lower in a given context, and the context is provided by a fuel through the link hasLowerCalorificValue. Defining such quantities as CalorificValue would also increase their reusability though we might decide not to reuse them.

msff2:
> > I think Lower and Higher CalorificValues are relative. A CalorificValue is lower in a given context, and the context is provided by a fuel through the link hasLowerCalorificValue. Defining such quantities as CalorificValue would also increase their reusability though we might decide not to reuse them.
> 
> I believe your understanding is wrong. Calorific Value aka heating value aka heat of combustion is a very well defined quantity for each fuel type and not relative in any sense.

I agree with you that Calorific Value is a well-defined quantity. It seems you didn't understand my message. I said lower and higher calorific values are relative.

sm453:
> I think Lower and Higher CalorificValues are relative. A CalorificValue is lower in a given context, and the context is provided by a fuel through the link hasLowerCalorificValue. Defining such quantities as CalorificValue would also increase their reusability though we might decide not to reuse them.

Sorry, I don't agree. A calorific value does not 'acquire' whether it is lower or higher through context - it does not depend on the context. Sure, the context of which fuel it belongs to is obviously key. But whether it is lower or higher is solely determined by whether or not it includes the heat of vaporisation of water. It is important to include into the representation whether or not a calorific value includes the vaporisation heat - just calling them all CalorificValue is not sufficient. Ok, one could use a property to capture this, but subclasses are preferable here, I believe.

sm453:
Looks like @markushofmeister and I agree. :-)

markushofmeister:
> > I think Lower and Higher CalorificValues are relative. A CalorificValue is lower in a given context, and the context is provided by a fuel through the link hasLowerCalorificValue. Defining such quantities as CalorificValue would also increase their reusability though we might decide not to reuse them.
> 
> Sorry, I don't agree. A calorific value does not 'acquire' whether it is lower or higher through context - it does not depend on the context. Sure, the context of which fuel it belongs to is obviously key. But whether it is lower or higher is solely determined by whether or not it includes the heat of vaporisation of water. It is important to include into the representation whether or not a calorific value includes the vaporisation heat - just calling them all CalorificValue is not sufficient. Ok, one could use a property to capture this, but subclasses are preferable here, I believe.

Isn't the solution to that simpoly

> Looks like @markushofmeister and I agree. :-)

So, isn't the solution simply:
`LowerCalorificValue rdfs:subClassOf CalorificValue`
`HigherCalorificValue rdfs:subClassOf CalorificValue`
and
`FuelType hasLowerCalorificValue LowerCalorificValue` 
`FuelType hasHigherCalorificValue HigherCalorificValue` 
or would you just keep
`FuelType hasCalorificValue CalorificValue`
instead of the previous 2 statements and query the type of calorific value simply from the rdf:type?

sm453:
> So, isn't the solution simply: `LowerCalorificValue rdfs:subClassOf CalorificValue` `HigherCalorificValue rdfs:subClassOf CalorificValue` and `FuelType hasLowerCalorificValue LowerCalorificValue` `FuelType hasHigherCalorificValue HigherCalorificValue` or would you just keep `FuelType hasCalorificValue CalorificValue` instead of the previous 2 statements and query the type of calorific value simply from the rdf:type?

I would go for the latter, like I suggested originally (https://github.com/cambridge-cares/TheWorldAvatar/pull/339#pullrequestreview-1139099726). I don't see the point in defining two relationships.

msff2:
> 



> 

By looking at the data in the HTML file made it clearer that we need to instantiate the LowerCalorificValue and HigherCalorificValue separately.

msff2:
> 

To reduce the complexity of queries, I will go for Seb's first option because it will not require us to discriminate between the lower and higher values with another axiom. This option will also reduce the query time.

sm453:
Actually, declaring
`LowerCalorificValue rdfs:subClassOf CalorificValue`
`HigherCalorificValue rdfs:subClassOf CalorificValue`
`FuelType hasLowerCalorificValue LowerCalorificValue`
`FuelType hasHigherCalorificValue HigherCalorificValue`
and not declaring
`FuelType hasCalorificValue CalorificValue`
has the advantage that one cannot associate an instance of CalorificValue (i.e. without specifying whether it is lower or higher) with a FuelType instance, because that would be nonsense.

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
```diff
@@ -0,0 +1,132 @@
...
+hasCumulativeEnergyCap,Object Property,,,Tier,http://www.ontology-of-units-of-measure.org/resource/om-2/Energy,,A relationship between a tier and its cumulative energy cap,http://www.theworldavatar.com/kg/ontoheatnetwork/,has cumulative energy cap
+hasUnitPrice,Object Property,,,Tier,HeatUnitPrice,,A relationship between a tier and the associated heat unit price,http://www.theworldavatar.com/kg/ontoheatnetwork/,has unit price
+hasTier,Object Property,,,TieredUnitPrice,Tier,,A relationship between tiered unit price and tier,http://www.theworldavatar.com/kg/ontoheatnetwork/,has tier
+isHoliday,Data Property,,,CalendarEffect,boolean,,Indication whether a particular date is holiday,http://www.theworldavatar.com/kg/ontoheatnetwork/,is holiday
+isVacation,Data Property,,,CalendarEffect,boolean,,Indication whether a particular date is vacation,http://www.theworldavatar.com/kg/ontoheatnetwork/,is vacation
+http://purl.org/dc/elements/1.1/date,Data Property,,,CalendarEffect,http://www.w3.org/2001/XMLSchema#dateTime,,A date associated with an event in the life cycle of the resource,http://purl.org/dc/elements/1.1/,date
```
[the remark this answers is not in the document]
markushofmeister: @xuehansong The "dc:date" issue still seems unsolved in the current version.

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
markushofmeister: @xuehansong please resolve

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
```diff
@@ -0,0 +1,132 @@
...
+https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries,Object Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure UNION http://www.theworldavatar.com/ontology/ontoems/OntoEMS.owl#Forecast UNION Availability UNION http://www.theworldavatar.com/ontology/ontoems/OntoEMS.owl#AirTemperature UNION CalendarEffect UNION CoGenRevenueInTimeInterval,https://www.theworldavatar.com/kg/ontotimeseries/TimeSeries,,A relationship between a measure or forecast and the respective time series.,https://www.theworldavatar.com/kg/ontoems/,has time series
+https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries,Object Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure UNION http://www.theworldavatar.com/ontology/ontoems/OntoEMS.owl#Forecast UNION Availability UNION http://www.theworldavatar.com/ontology/ontoems/OntoEMS.owl#AirTemperature UNION CalendarEffect UNION CoGenRevenueInTimeInterval,https://www.theworldavatar.com/kg/ontotimeseries/TimeSeries,exactly 1,,https://www.theworldavatar.com/kg/ontoems/,has time series
```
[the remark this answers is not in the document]
markushofmeister: @xuehansong please resolve. They are not defined by OntoEMS but OntoHeatNetwork

JPS_Ontology/ontology/ontoheatnetwork/ontoheatnetwork.csv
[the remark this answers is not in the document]
markushofmeister: <owl:DatatypeProperty rdf:about="http://nomisma.org/ontology#hasDate">
        <rdfs:domain rdf:resource="https://www.theworldavatar.com/kg/ontoheatnetwork/CalendarEffect"/>
        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#date"/>
        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">A date associated with an event in the life cycle of the resource</rdfs:comment>
        <rdfs:isDefinedBy rdf:datatype="http://www.w3.org/2001/XMLSchema#string">http://purl.org/dc/elements/1.1/</rdfs:isDefinedBy>
        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">date</rdfs:label>
    </owl:DatatypeProperty>

@xuehansong This is confusing: You solved the `DataProperty` issue, but might want to update the defined by as well. I inspected the ontology in WebVowl and could not identify any changes

markushofmeister (changes requested):
Please 1) resolve the remaining `defined by` issues and 2) ideally separate the agent from the ontology branch so that we can merge the ontology finally to main without needing to wait until the agent is finished.

markushofmeister:
@xuehansong Is this PR still relevant?

xuehansong:
> @xuehansong Is this PR still relevant?

@markushofmeister Yes, the instantiation agent is under it

markushofmeister:
> > @xuehansong Is this PR still relevant?
> 
> @markushofmeister Yes, the instantiation agent is under it

I see. Then please be careful when merging at some point as there are conflicts with the heating network ontology files!

xuehansong:
> > > @xuehansong Is this PR still relevant?
> > 
> > 
> > @markushofmeister Yes, the instantiation agent is under it
> 
> I see. Then please be careful when merging at some point as there are conflicts with the heating network ontology files!

@markushofmeister noted with thanks

sm453:
What's the status of this? We need to urgently complete this.

xuehansong:
> What's the status of this? We need to urgently complete this.

I passed this instantiation agent to Srishti about 4 weeks ago and she was able to deploy it successfully at her local PC and then she did some test for it. I will check with her on next Monday about the test.

I think Markus and Magnus did some update to the district heating network miro board. I will check this also on next Monday for anything that is different from what I had 4 weeks ago and amend the agent accordingly.

sm453:
@xuehansong, can you please rename this pull request to accurately reflect what is being done here? There are now no changes regarding ontology/TBox on this branch any more - only the agent is left.

sm453:
The main folder is called `DistrictHeatingNetworkAgent`, which contains a subfolder `DistrictHeatNetworkAgent`, whereas the image is called `heatnetwork-agent` and the container is called `HeatNetwork-agent`. This isn't helpful. In order to introduce some consistency, can you please rename both folders to `DistrictHeatingAgent` (as a compromise between descriptiveness and brevity), and both the image and container to `district-heating-agent`?

sm453:
`DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/heatnetwork-agent.iml`
Does this file need to be versioned? I suspect not...

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+		<version>1.0.0</version>
+	</parent>
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.1</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
```
sm453, line 20: Please use something more recent, e.g. at least 11.

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>uk.ac.cam.cares.jps</groupId>
+  <artifactId>heatnetwork-agent</artifactId>
+  <version>1.0.0</version>
+  <packaging>war</packaging>
+	<!-- Parent POM -->
+	<parent>
+		<groupId>uk.ac.cam.cares.jps</groupId>
+		<artifactId>jps-parent-pom</artifactId>
+		<version>1.0.0</version>
```
sm453, line 11: Please use something more recent, e.g. 2.0.0.

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+		</dependency>
+		<dependency>
+	        <groupId>xml-apis</groupId>
+	        <artifactId>xml-apis</artifactId>
+	        <version>1.4.01</version>
+		</dependency>
+     <dependency>
+		  <groupId>uk.ac.cam.cares.jps</groupId>
+		  <artifactId>jps-base-lib</artifactId>
+		  <version>1.20.2</version>
```
sm453, line 84: Please use something more recent, e.g. 1.30.0.

sm453:
There seem to be two `csv` data files for testing purposes in the `config` folder. That is not a good place. I believe something like `src/test/resources` would be more appropriate. Please move the files.

Agents/DistrictHeatingNetworkAgent/Dockerfile
```diff
@@ -0,0 +1,47 @@
+# First stage: build war file
+#==================================================================================================
+FROM maven:3.6-openjdk-11-slim as builder
+
+# Copy all files into root's home, including the source, pom file, ./m2 directory, credentials and config files
+ADD . /root
+
+# Populate settings templates with credentials, repo name
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 10: "latter, but not the former"

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/src/uk/ac/cam/cares/jps/agent/heat/HeatTest.java
```diff
@@ -0,0 +1,149 @@
...
+		        } catch (ParseException ex){ 
+		        }
+		    }
+		    
+		    //System.out.println(inDate);
+		    String front = outDate.substring(0,10); 
+		    //System.out.println(front); System.out.println(front+"AAAA");
+		    String back = outDate.substring(11);
+		   // System.out.println(back); System.out.println(front+"BBB");
+		   // System.out.println(outDate+"CCC");
```
sm453, line 144: Please tidy up all the commented-out code in this file.

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/src/main/java/uk/ac/cam/cares/jps/agent/heat/HeatNetworkInputAgent.java
```diff
@@ -0,0 +1,686 @@
...
+
+
+    // Converts the readings in form of maps to time series' using the mappings to IRI.
+    // @return A list of time series objects (one per mapping) that can be used with the time series client.
+    private List<TimeSeries<OffsetDateTime>> convertReadingsToTimeSeries(Map<String, List<?>> HeatingNetworkReadings)
+            throws NoSuchElementException, IOException {
+        List<OffsetDateTime> HeatingNetworkTimestamps = HeatingNetworkReadings.get(HeatNetworkInputAgent.timestampKey).stream()
+                .map(timestamp -> (convertStringToOffsetDateTime(timestamp.toString()))).collect(Collectors.toList());
+        List<TimeSeries<OffsetDateTime>> timeSeries = new ArrayList<>();
+        //String mappingFolder = "/Users/HXUE01/Documents/GitHub/TheWorldAvatar/Agents/DistrictHeatingNetworkAgent/config/mappings";
```
sm453, line 544: Please remove this comment.

sm453 (changes requested):

xuehansong:
> @xuehansong, can you please rename this pull request to accurately reflect what is being done here? There are now no changes regarding ontology/TBox on this branch any more - only the agent is left.

@sm453 ok, I have renamed it

xuehansong:
> `DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/heatnetwork-agent.iml` Does this file need to be versioned? I suspect not...

Right, I also don't think so

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/src/uk/ac/cam/cares/jps/agent/heat/HeatTest.java
xuehansong: This file has been removed as it was a java sample code earlier

Agents/DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/src/main/java/uk/ac/cam/cares/jps/agent/heat/HeatNetworkInputAgent.java
xuehansong: I have done all the amendments according to the comments. Thank you very much for the feedback and please let me know any further changes that are required.

sm453:
> > `DistrictHeatingNetworkAgent/DistrictHeatNetworkAgent/heatnetwork-agent.iml` Does this file need to be versioned? I suspect not...
> 
> Right, I also don't think so

Can you please remove it, then?

Agents/DistrictHeatingAgent/DistrictHeatingAgent/src/main/java/uk/ac/cam/cares/jps/agent/heat/HeatNetworkInputAgent.java
```diff
@@ -0,0 +1,679 @@
...
+
+
+        // For the Consumer part
+        UpdateBuilder Consumer_ub =
+                new UpdateBuilder()
+                        .addInsert(NodeFactory.createURI(KB + "Consumer"), NodeFactory.createURI(OHN + "hasHeatDemand"), NodeFactory.createURI(KB + "HeatDemand"))
+                        .addInsert(NodeFactory.createURI(KB + "Consumer"), NodeFactory.createURI(RDF_TYPE), NodeFactory.createURI(OHN + "Consumer"))
+                        .addInsert(NodeFactory.createURI(KB + "HeatDemand"), NodeFactory.createURI(RDF_TYPE), NodeFactory.createURI(OHN + "HeatDemand"));
+        UpdateRequest Consumer_ur = Consumer_ub.buildRequest();
+        AccessAgentCaller.updateStore("http://host.docker.internal:48888/ontoheatnet", Consumer_ur.toString());
```
sm453, line 156: Instead of putting the same URL string literal in 14 different places, please use a variable instead. That way, if you ever need to change it, you only need to change it in one place, rather than 14.

Secondly, you've already got a client.properties file in which one can configure query and update URLs for a SPARQL endpoint. Why don't you make use of that, instead of hard-coding the URL? Otherwise, what's the point of the setting in the client.properties file?

Thirdly, if you impose a URL, there is no point in using the access agent at all. You may want to ask, e.g., Hou Yee to explain how he does it in his instantiation agents.

sm453 (approved):

markushofmeister (changes requested):
As requested, I had a look at the actual triples instantiated and have the following remarks. 
**What did I do?**
We have previously developed an instantiation script in [Python](https://github.com/cambridge-cares/pirmasens/tree/main/districtheating). I am not claiming that this script is perfect; however, I invested quite some time to check the correct instantiation by that script. So, to check the correct instantiation of this agent, I "simply" exported both sets of triples and compared them. The exports are available as `.nt` files on Dropbox.

**What did I encounter?**
1) The total number of triples instantiated differs quite a bit: 522 in Python vs. 415 using this agent
2) All time series triples (i.e. `hasRDB`, ...) are missing in the instantiation of this agent
3) There seem to be typos in the instantiated concepts, e.g. `https://www.theworldavatar.com/kg/ontoheatnetwork/emandDrivenWearCost`
4) I tried to run a graph difference between the two instantiations; however, this is basically impossible as the IRI naming is very different and I could not create a fuzzy matching quickly. Hence, I focused on scrutinising the instantiated relationships (which naming should be consistent between both instantiations (i.e. no arbitrary instance names, UUIDs, etc.). I analysed which relationships are there in each instantiation and how often do they occur. The results are listed below and seem quite different:

('change', ['http://www.ontology-of-units-of-measure.org/resource/om-2/hasUnit'], (41, 54))
('change', ['http://www.w3.org/1999/02/22-rdf-syntax-ns#type'], (161, 180))
('change', ['http://www.ontology-of-units-of-measure.org/resource/om-2/hasValue'], (41, 54))
('change', ['http://www.ontology-of-units-of-measure.org/resource/om-2/hasNumericalValue'], (26, 17))
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasCO2Factor'], (4, 1))
('change', ['http://www.w3.org/2000/01/rdf-schema#label'], (7, 10))
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasHigherCalorificValue'], (4, 1))
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasLowerCalorificValue'], (4, 1))
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasCurrentUnitPrice'], (2, 1))
('add', '', [('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries', 44), 
             ('https://www.theworldavatar.com/kg/ontoheatnetwork/hasObservableProperty', 8), 
             ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeUnit', 44), 
             ('https://www.theworldavatar.com/kg/ontotimeseries/hasRDB', 44), 
             ('https://www.theworldavatar.com/kg/ontoheatnetwork/isFulfilledBy', 1)
             ])
('remove', '', [('https://www.theworldavatar.com/kg/ontoheatnetwork/hasObservableProperties', 16), 
                ('http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_performance/economic_performance.owl#hasCost', 29), 
                ('http://www.ontology-of-units-of-measure.org/resource/om-2/symbol', 11), 
                ('http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_performance/economic_performance.owl#hasRevenue', 1), 
                ('http://theworldavatar.com/ontology/ontochemplant/OntoChemPlant.owl#hasIndividualCO2Emission', 4), 
                ('https://www.theworldavatar.com/kg/ontoheatnetwork/isFullfilledBy', 1), 
                ('https://www.theworldavatar.com/kg/ontoheatnetwork/hasCoGenElectricityAmount', 1)
                ])

**How should this be read?**
`change`: The relationship is present in both instantiations, but with different number of occurrences. First number refers to occurrences in instantiation by this agent and second number to Python one
`add`: Relationship only present in Python instantiation
`remove`: Relationship only present in instantiation by this agent

**What does that mean?**
I suggest you carefully revisit the instantiation to check you are not missing / adding anything which is not wanted. Furthermore, we should include the instantiation of time series data. Otherwise, this agent doesn't really make sense to me.

@sm453 What are your thoughts?

sm453:
@xuehansong, can you please look into the differences between the two instantiations? Where is it going wrong?

xuehansong:
@markushofmeister Hi Markus, thank you for the feedback. For the data I passed to you in comparison with your python code, it does not include the time-series part. After adding the time-series part, my agent's triple is 434, still much less than yours, it is due to the following major reasons from the time-series part:

1. In your python code, ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries', 44), there are 44 hasTimeSeries triples, but in my code only 16, because there are only 16 time-series data given to me. I am not sure where are the rest 28 time-series data from? 

2.  Your code also has 44 hasTimeUnit and has RDB ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeUnit', 44) ,('https://www.theworldavatar.com/kg/ontotimeseries/hasRDB', 44), but my code only has 1 hasTimeUnit and 1 hasRDB. I have checked this with other time-series database, e.g., nusDavisWeatherStation, it only has 1 hasTimeUnit and 1 hasRDB triple

For the others like:
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasCO2Factor'], (4, 1)),
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasHigherCalorificValue'], (4, 1)),
('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasLowerCalorificValue'], (4, 1)),
('remove', 'http://theworldavatar.com/ontology/ontochemplant/OntoChemPlant.owl#hasIndividualCO2Emission', 4),
Because we have 3 instances of boilers and 1 gas turbine, therefore they are 4 triples in my case.

markushofmeister:
> 1. In your python code, ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries', 44), there are 44 hasTimeSeries triples, but in my code only 16, because there are only 16 time-series data given to me. I am not sure where are the rest 28 time-series data from?

That is hard to judge for me as I do not have your data with the added TimeSeries relationships. I suggest you use the `dh_magnus.nt` file and upload it to Blazegraph and start exploring where the differences are, e.g. by 
```
select distinct ?s ?p
where { ?s <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> ?p }
order by ?s
```

16 seem quite little as we have 4 generators and the incineration plant, which all have cost (even though some of them are constant and should be instantiated as stable time series) + the 4 observation points within the grid to observe temperature, pressure, etc. + calendar effects etc.

markushofmeister:
> 2\. Your code also has 44 hasTimeUnit and has RDB ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeUnit', 44) ,('https://www.theworldavatar.com/kg/ontotimeseries/hasRDB', 44), but my code only has 1 hasTimeUnit and 1 hasRDB. I have checked this with other time-series database, e.g., nusDavisWeatherStation, it only has 1 hasTimeUnit and 1 hasRDB triple

According to [OntoTimeSeries](https://miro.com/app/board/uXjVPFaO5As=/) each timeseries IRI should have the `hasRDB` and `hasTimeUnit` data properties. Hence, having 44 time series instances should also mean having 44 of each of those properties. I am not sure why the nusDavisWeatherStation handles that differently.

markushofmeister:
> ('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasCO2Factor'], (4, 1)),
> ('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasHigherCalorificValue'], (4, 1)),
> ('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasLowerCalorificValue'], (4, 1)),

This is not in line with the ontology. All boilers + GT should be connected to a fuel type (which should be the same instance for all generators, as they all use the same fuel in our case), which then is connected to CO2 factor etc. 
Your instantiation differs from that:

image
image

markushofmeister:
> ('remove', 'http://theworldavatar.com/ontology/ontochemplant/OntoChemPlant.owl#hasIndividualCO2Emission', 4),

This is a rather tricky one: This relationship is missing in the Python instantiation as we only instantiated historical data where we did not know the actual emissions. Our approach was to not instantiate for now and only add that data once the optimisation agent runs and estimates that data (So, I guess this is just up to preference)

markushofmeister:
@xuehansong Once you had the chance to look through the discrepancies of the instantiation, it would be good to follow up on my comments here, so that we can hopefully close this PR rather sooner than later.

xuehansong:
> @xuehansong Once you had the chance to look through the discrepancies of the instantiation, it would be good to follow up on my comments here, so that we can hopefully close this PR rather sooner than later.
image

Hi Markus, thanks for the feedback. Refer to the attached picture above, the **yellow highlighted** ones I have already done the amendments by either removing them or amending according to the python instantiation ones. The **green highlighted** ones  I would like to clarify with you. 

('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasCurrentUnitPrice'], (2, 1)): I have 2 here because there is tier 1 and tier 2 UnitPrice, therefore it turns out to be 2 has hasCurrentUnitPrice

image

('https://www.theworldavatar.com/kg/ontoheatnetwork/hasObservableProperty', 8), 'remove', '', [('https://www.theworldavatar.com/kg/ontoheatnetwork/hasObservableProperties', 16), : 1st the difference is due to singular and plural, 2nd the python code has 8 because it does not have "density" and "specific heat capacity" inside

('http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_performance/economic_performance.owl#hasCost', 29), : this one the python code does not have, then I am not how the various cost can be linked to the HeatGenerator and MunicipalUtility in the python code

image

I am working on the remaining non-highlighted ones now.

markushofmeister:
> ('change', ['https://www.theworldavatar.com/kg/ontoheatnetwork/hasCurrentUnitPrice'], (2, 1)): I have 2 here because there is tier 1 and tier 2 UnitPrice, therefore it turns out to be 2 has hasCurrentUnitPrice

I believe this is wrong. The idea is to have several tiers with different unit prices (i.e. each unit is getting cheaper the more one buys); however, there can only be one valid unit price at each point in time (which is then either tier 1 or 2 (etc.)

markushofmeister:
> ('https://www.theworldavatar.com/kg/ontoheatnetwork/hasObservableProperty', 8), 'remove', '', [('https://www.theworldavatar.com/kg/ontoheatnetwork/hasObservableProperties', 16), : 1st the difference is due to singular and plural, 2nd the python code has 8 because it does not have "density" and "specific heat capacity" inside
1) Ontologically correct is singular `hasObservableProperty` (the MIRO was outdated and I amended it there too now). Please correct.
2) True, we only instantiated temperature and pressure time series (as we had data for those). Fine if you do all 4, just wondering which data you use for those

markushofmeister:
>('http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_performance/economic_performance.owl#hasCost', 29), : this one the python code does not have, then I am not how the various cost can be linked to the HeatGenerator and MunicipalUtility in the python code

The idea was that we do not know the historic cost, so we do not know what to instantiate here. We only instantiated the unit cost and prices etc. We thought about instantiating the cost instances after having run the optimisation agent, ie instantiating them by the optimisation agent as results

xuehansong:
@markushofmeister I am done with updating the data instantiation

image

The number in red colour is from my latest committed code. In my code, all the time-series data have only one "hasTimeUnit" and "hasRDB" triple because all the time-series data are instantiated together. Those triples from "removed" parts have been removed also. 

Thank you,

xuehansong:
@markushofmeister Hi, I have finished updated the instantiation. 

image
image

This shows the difference between the two codes, the number in front is from Python, and latter one is my code.  I have added "https://www.theworldavatar.com/kg/ontoheatnetwork/hasCoGenElectricityAmount" with ts data in my code

For the type triples, 139(my code) + 43(my TS only has one "https://www.theworldavatar.com/kg/ontotimeseries/hasTimeUnit") - 1 (from CoGenElectricityAmount) -  1 ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries') = 180(the python code).

Thanks.

markushofmeister:
> For the type triples, 139(my code) + 43(my TS only has one "https://www.theworldavatar.com/kg/ontotimeseries/hasTimeUnit") - 1 (from CoGenElectricityAmount) - 1 ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries') = 180(the python code).

Thanks for the explanation, very helpful! This explains most of the changes depicted in your screenshots (btw, maybe easier to paste code inside [triple backticks](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code) next time); however, can you please elaborate
1) What is the second added `TimeSeries` in your instantiation compared to our previous one? The added `CoGenElectricityAmount` explains only one, but two seem to be added
2) The overall number of triples still differs quite significantly: 522 (Python) vs 400 (yours). Considering the impact of the different TimeSeries handling (which is ok), your equivalent number of triples would be 400 + 43 x 2 = 486, which still leaves a discrepancy of 36 triples (~10%).  Where does this come from?
3) Basically all units are wrong! First of all, you seem to have created new units for everything instead of using the existing ones from [om](https://github.com/cambridge-cares/OM) and [ontouom](https://github.com/cambridge-cares/TheWorldAvatar/tree/main/JPS_Ontology/ontology/ontouom). Secondly, multiple relationships between Quantities and their Units are wrong, e.g. pressure reported in `MWh`. This needs to be fixed!
Please check with 
```
prefix ts:<https://www.theworldavatar.com/kg/ontotimeseries/>
prefix oh:<https://www.theworldavatar.com/kg/ontoheatnetwork/>
prefix om:<http://www.ontology-of-units-of-measure.org/resource/om-2/> 

select distinct ?quantity ?unit
where { 
  ?quantity om:hasValue/om:hasUnit ?unit
}
ORDER BY ?unit
```
@sm453 Shall we use this opportunity to (finally) include all those ontouom units into our version of OM? Currently they are still kept outside in ontouom as initially done as workaround

4) Some of the actual instantiated values seem wrong or at least abbreviated, e.g. the actual `CO2 factor` of  0.2016 is instantiated as just 0.2. This might even be acceptable for our use case, but doesn't look like good practice!

General guidance:
Next time please follow our instance naming convention of 
`<Instance_IRI> = Instance_concept_name + _ + UUID4`

markushofmeister (changes requested):
> For the type triples, 139(my code) + 43(my TS only has one "https://www.theworldavatar.com/kg/ontotimeseries/hasTimeUnit") - 1 (from CoGenElectricityAmount) - 1 ('https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries') = 180(the python code).

Thanks for the explanation, very helpful! This explains most of the changes depicted in your screenshots (btw, maybe easier to paste code inside [triple backticks](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code) next time); however, can you please elaborate
1) What is the second added `TimeSeries` in your instantiation compared to our previous one? The added `CoGenElectricityAmount` explains only one, but two seem to be added
2) The overall number of triples still differs quite significantly: 522 (Python) vs 400 (yours). Considering the impact of the different TimeSeries handling (which is ok), your equivalent number of triples would be 400 + 43 x 2 = 486, which still leaves a discrepancy of 36 triples (~10%).  Where does this come from?
3) Basically all units are wrong! First of all, you seem to have created new units for everything instead of using the existing ones from [om](https://github.com/cambridge-cares/OM) and [ontouom](https://github.com/cambridge-cares/TheWorldAvatar/tree/main/JPS_Ontology/ontology/ontouom). Secondly, multiple relationships between Quantities and their Units are wrong, e.g. pressure reported in `MWh`. This needs to be fixed!
Please check with 
```
prefix ts:<https://www.theworldavatar.com/kg/ontotimeseries/>
prefix oh:<https://www.theworldavatar.com/kg/ontoheatnetwork/>
prefix om:<http://www.ontology-of-units-of-measure.org/resource/om-2/> 

select distinct ?quantity ?unit
where { 
  ?quantity om:hasValue/om:hasUnit ?unit
}
ORDER BY ?unit
```
@sm453 Shall we use this opportunity to (finally) include all those ontouom units into our version of OM? Currently they are still kept outside in ontouom as initially done as workaround

4) Some of the actual instantiated values seem wrong or at least abbreviated, e.g. the actual `CO2 factor` of  0.2016 is instantiated as just 0.2. This might even be acceptable for our use case, but doesn't look like good practice!

General guidance:
Next time please follow our instance naming convention of 
`<Instance_IRI> = Instance_concept_name + _ + UUID4`

sm453:
> Shall we use this opportunity to (finally) include all those ontouom units into our version of OM? Currently they are still kept outside in ontouom as initially done as workaround

Yes, please (see issue #576). Please make sure you keep @Abyssus1010 in the loop.

markushofmeister:
> > Shall we use this opportunity to (finally) include all those ontouom units into our version of OM? Currently they are still kept outside in ontouom as initially done as workaround
> 
> Yes, please (see issue #576). Please make sure you keep @Abyssus1010 in the loop.

@xuehansong Can you please take care of that in a **separate** pull request?

xuehansong:
@markushofmeister Many thanks for the feedback. 

1. The second added TimeSeries is Waermeeinspeisung and I have now removed this concept as it is not represented in the DHN ontology.

2. Total triples: 522 (Python) vs 399 (my), 399+43*3= 528. There are 6 difference due to the "hasCoGenElectricityAmount" ts data in my code:
"https://www.theworldavatar.com/kg/ontoheatnetwork/hasCoGenElectricityAmount"
image
image

3. All the units error has been fixed. Below is what in my:
image
 
and below is what in python code: 
image

4. CO2 factor is changed to 0.2016 and noted on the naming convention

5. Fyi, the output data NT file is located at: https://www.dropbox.com/s/yexzadi2fqeqqde/ontoheatnet_20230328.nt?dl=0

Thank you

sm453:
It should be `cubicMetrePerHour`. Please correct.

markushofmeister:
I believe we are almost there! Only two things left:
1) The `hasMaxAnnualPurchaseVolume` (and same for the minimum) point to instances without actual values. That should be fixed
2) The `hasRatedThermalPower` seem to be equivalent for all 4 heat generators. That should not be the case.

markushofmeister (changes requested):
I believe we are almost there! Only two things left:
1) The `hasMaxAnnualPurchaseVolume` (and same for the minimum) point to instances without actual values. That should be fixed
2) The `hasRatedThermalPower` seem to be equivalent for all 4 heat generators. That should not be the case.

xuehansong:
@markushofmeister Thanks

For point 1 fixed: 
image
image

For point 2 fixed
image

markushofmeister (approved):
I tested with the latest triple export from Dropbox (https://www.dropbox.com/s/56l0rnhqh3m9eei/ontoheatnet_20230329.nt?dl=0) and think that all previously discovered and discussed issues are resolved

sm453 (approved):
Builds and runs in the stack, and I could not find any further issues with the instantiated triples.

PR #341, opened 2022-09-23
Make it possible to upload CityGML and CityJSON files into the stack

gpeb2:
Use the 3DCityDB PostgrSQL and import/export containers to upload and store CityGML and CityJSON formatted data in the stack.

Here is a list of completed and pending tasks:

- [x] Update the PostGIS image to be one that includes the 3DCityDB extension.
- [x] Add a data subtype for 3DCityDB data sources (CityGML and CityJSON).
- [x] Add a `CityDBClient` class that can upload the input files to the CityDB database.
- [x] Add a `CityTilerClient` class that can generate 3Dtiles from the database.
- [x] Update the code so that each cityobject is assigned a UUID and IRI.
- [x] Modify the tiler so that the IRIs are written into the 3Dtiles so that they can be passed to the feature info agent.
- [x] Update the stack-data-uploader README.
- [x] Add existing SQL command that fixes slightly broken CityGML files. 
- [x] Write out the imported data back to CityGML files so that the IRI, UUID, and any autogenerated GMLID, values are preserved if the data subset is reuploaded.
- [ ] *Add a way to subtract elevation values from the building geometries so that they don't float in the air if there is no terrain layer.
- [ ] *Combine the generated tilesets so that they act as one with more detail added as you zoom in.
- [ ] *Add an example including an Ontop mapping.
- [ ] *Extend the tiler to handle LODs other than LOD2.

The tasks with an * might need to be pushed to a future pull request.


ChungTingLao (approved):
Tested with Pirmasens and Kings Lynn data, works as intended.

PR #342, opened 2022-09-25
168 dev monitor multiple async derivations

jb2197:
In this PR, changes were made to optimise the agent monitoring multiple asynchronous derivations:
(1) in the same monitoring period, the agent updates its record of derivation status after processing each "Requested" derivation;
(2) when changing the status of a derivation from "Requested" to "InProgress", the agent adds a data property "uuidLock" to the derivation as part of the SPARQL update, the agent will only progress to execute the job if that SPARQL update is deemed successful, otherwise, it indicates the job has been processed by another agent thread

The OntoDerivation TBox will be updated to add "uuidLock" once this PR is approved.


sm453 (approved):
Don't forget to remove the SNAPSHOT bit from the base lib pom.

PR #343, opened 2022-09-27
1416 dev split storeclient interface

mehalaga:


JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/RemoteStoreClient.java
```diff
@@ -350,14 +355,8 @@ public int executeUpdate(UpdateRequest update) {
      */
     @Override
     public int executeUpdate(String query) {
```
csl37, line 357: It would be helpful to clean up the 3 executeUpdate methods a bit.
executeUpdate() does getConnectionUrl(), but this is then repeated in connect() when executeUpdate(String query) is called. Also some checks are performed on the url which are not performed in the other methods.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/query/RemoteStoreClient.java
```diff
@@ -468,17 +467,11 @@ public JSONArray executeQuery() {
      */
     @Override
     public JSONArray executeQuery(String query) {
```
csl37, line 469: Same comment for the query methods as for the update methods above

csl37 (changes requested):

csl37 (approved):

PR #344, opened 2022-09-27
1418 dev ontordbrouter

mehalaga:


sm453 (changes requested):
Fine, as discussed. But please also add the csv file from which the owl can be generated using our TBox generation tool.

sm453 (approved):

PR #349, opened 2022-10-03
Dev onto bms

Abyssus1010:


sm453 (changes requested):
Based on our discussions, looks ok to me. Can you please also add the corresponding csv file with which one can regenerate the owl file using our TBox generation tool?

sm453 (approved):

PR #351, opened 2022-10-04
170 dev async derivation queryagain flag

jb2197:
In this PR, the queryAgain flag is set to false if any of the upstream async derivations of the Requested derivation are still outdated, avoiding flooding the triple store with status queries.


sm453 (approved):
Don't forget to remove the snapshot version labels from this and the baselib pom.

PR #364, opened 2022-10-07
32-dev-time-resolution-and-measure-type-extension-in-ontotimeseries

YiKaiTsai:


JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl
```diff
@@ -1,22 +1,43 @@
 <?xml version="1.0"?>
-<rdf:RDF xmlns="https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl#"
-     xml:base="https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl"
-     xmlns:OntoTimeSeries="https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl#"
-     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
```
markushofmeister, line 5: I know that the updated base IRI `https://www.theworldavatar.com/kg/ontotimeseries/` follows the new convention; however, that would mean that most of the current agents using the time series with the "old" base IRI `https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl` would become erroneous. I think we might keep the old base IRI for now.
@sm453 What is your opinion on that? Maybe we could also use a different version of the ontology to avoid this issue, but I am not too sure how that would work exactly (especially with referencing the right version when instantiating data)

sm453: This is frustrating.

While IRIs are completely arbitrary in general, something like https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl shouldn't really be used for anything other than development/testing purposes. From that point of view, I'd prefer a correct production IRI https://www.theworldavatar.com/kg/ontotimeseries/, although I appreciate that this doesn't play well with some tools that some people currently like to use.

More fundamentally, people shouldn't be using OWL files at all (at least not for this purpose), but I appreciate that at present there is no alternative in place.

Consistency across the code base is a practical issue, however a temporary one that will go away once we start doing things properly (which is admittedly not going to happen any time soon).

So, unfortunately, I have to agree, very reluctantly and under protest, that this awful IRI https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl may be the path of least resistance for now.

markushofmeister: >From that point of view, I'd prefer a correct production IRI https://www.theworldavatar.com/kg/ontotimeseries/, although I appreciate that this doesn't play well with some tools that some people currently like to use.

It's not that those tools, i.e. instantiation scripts and agents, would not work anymore. It's just that they instantiate "outdated", i.e. meaningless, IRIs.

> So, unfortunately, I have to agree, very reluctantly and under protest, that this awful IRI https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl may be the path of least resistance for now.

I do remember that we had exactly this discussion about the `timeseries` IRI a few months back and decided to keep it "for now" as this IRI is hardcoded in the TimeSeriesClient and would have required quite a few changes. Another perspective to look at it is the follows:
Mehal just pushed and merged changes to the TimeSeriesClient and we could easily add a hotfix and change the base IRI. Jiaru just release a new py4jps version and we could also add a hotfixed version there.
The stack (for King's Lynn) is still not working and I have to (re-)deploy all agents from scratch anyway. Yi Kai and Magnus are going to instantiate time series data soon as well. So, the cost of switching are just getting bigger and it might actually be not a bad moment to implement those changes now as at least 3 people are working with the updated version then.
@sm453 Any thoughts?

sm453: I agree. Go for it - it's not getting any better...

markushofmeister (approved):
Looks alright to me. As discussed earlier with @YiKaiTsai this change will require changes to the TimeSeriesClient to instantiate the newly defined subclasses properly. I already talked to Mehal and she is keen to support that.

The only question therefore is: Have we looked into using `https://www.w3.org/TR/owl-time/#time:TemporalDuration` (or something similar) instead of defining our own `Interval` concept and classifying this as `om:Quantity`?

sm453:
> The only question therefore is: Have we looked into using `https://www.w3.org/TR/owl-time/#time:TemporalDuration` (or something similar) instead of defining our own `Interval` concept and classifying this as `om:Quantity`?

I would support re-using time:TemporalDuration, rather than defining yet another duplicate...

JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.owl
```diff
@@ -84,12 +97,74 @@
...
+
+
+    <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontotimeseries/TimeSeries">
+        <rdfs:isDefinedBy rdf:datatype="http://www.w3.org/2001/XMLSchema#string">https://www.theworldavatar.com/kg/ontotimeseries/</rdfs:isDefinedBy>
+    </owl:Class>
+    
+
+
+    <!-- http://www.w3.org/2006/time#TemporalDuration -->
+
```
markushofmeister, line 165: Looking at `https://www.w3.org/TR/owl-time/#time:TemporalDuration` in more detail, I think this not the right class (at least it makes it unnecessary complex). 
I would suggest using `https://www.w3.org/TR/owl-time/#time:Duration` (which is a subclass of TemporalDuration), which can be directly related to the actual duration value and unit using `https://www.w3.org/TR/owl-time/#time:numericDuration` and `https://www.w3.org/TR/owl-time/#time:unitType`, respectively.
I would suggest to include all those triples into our ontology (setting the defined by column to `https://www.w3.org/TR/owl-time/`) so that people directly see how to use it. Maybe we could even include the temporal units just for reference

markushofmeister (changes requested):

markushofmeister (changes requested):
I believe this is unnecessary complex: 
image

I thought we aligned on
`AverageTimeSeries hasAveragingPeriod time:Duration`
`time:Duration time:numericDuration xsd:decimal`
`time:Duration time:unitType time:TemporalUnit`



TBox (in case the `subClassOf` caused confusion: We do not need to include this into our ontology as we define the `time` concepts correctly as `defined by` https://www.w3.org/TR/owl-time/# )
image
image
image

ABox
image

markushofmeister:
@sm453 Please correct me if I am wrong!

YiKaiTsai:
> I believe this is unnecessary complex: image
> 
> I thought we aligned on `AverageTimeSeries hasAveragingPeriod time:Duration` `time:Duration time:numericDuration xsd:decimal` `time:Duration time:unitType time:TemporalUnit`
> 
> TBox (in case the `subClassOf` caused confusion: We do not need to include this into our ontology as we define the `time` concepts correctly as `defined by` https://www.w3.org/TR/owl-time/# ) image image image
> 
> ABox image

Apologies, this is my mistake. I committed the wrong version of the files.

markushofmeister (approved):

sm453 (approved):

PR #368, opened 2022-10-11
175-dev-py4jps-1025: released py4jps 1.0.25 to both PyPI and TestPyPI.

jb2197:


kokfoongcmcl (approved):
I still think it's better to have the same version number as the Java package, at the moment I'm not sure how someone is able to determine which base lib version this is based on

jb2197:
> I still think it's better to have the same version number as the Java package, at the moment I'm not sure how someone is able to determine which base lib version this is based on

One can check the jps-base-lib version in the JPS_BASE_LIB/pom.xml when the release commit was made. This is not ideal but it will remain as it is before the automated release is set up.

PR #375, opened 2022-10-14
Dev update ontobuiltenv address data

markushofmeister:
Updated address information to better accomodate information about flat and building names (see MIRO board for details: https://miro.com/app/board/uXjVOIznYOU=/)

image


sm453:
I guess the range of isPresumedMatchOf cannot be made more specific than owl:Thing?

markushofmeister:
> I guess the range of isPresumedMatchOf cannot be made more specific than owl:Thing?

For the Land Registry data only we could use `http://landregistry.data.gov.uk/def/common/BS7666Address`, but I am not sure as this property could also be used for linking to other data sets

sm453:
> > I guess the range of isPresumedMatchOf cannot be made more specific than owl:Thing?
> 
> For the Land Registry data only we could use `http://landregistry.data.gov.uk/def/common/BS7666Address`, but I am not sure as this property could also be used for linking to other data sets

...and I imagine this would be too specific to be useful...

sm453 (approved):

Issue #376, opened 2022-10-14
Error thrown when trying to query raster data in the stack


gpeb2:
The following error is output when trying to query raster data stored in the PostGIS DB in the dynamic stack:
```ERROR: rt_band_load_offline_data: Access to offline bands disabled```

PR #377, opened 2022-10-14
376-error-thrown-when-trying-to-query-raster-data-in-the-stack

gpeb2:
Changed the command to set the "postgis.enable_outdb_rasters" and "postgis.gdal_enabled_drivers" properties when uploading a raster to the PostGIS database. Previously it was only setting it for the session, now it sets it permanently for the database. This should fix issue #376 .

### DIFF
--- a/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
+++ b/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.8'
 services:
   stack-client:
-    image: docker.cmclinnovations.com/stack-client:1.5.2-376-error-thrown-when-trying-to-query-raster-data-in-the-stack-SNAPSHOT
+    image: docker.cmclinnovations.com/stack-client:1.5.2
     configs:
       - postgis
       - geoserver
--- a/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
+++ b/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.8'
 services:
   stack-client:
-    image: docker.cmclinnovations.com/stack-client:1.5.1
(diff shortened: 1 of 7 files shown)


markushofmeister (commented):
Changes look appropriate to me (i.e. very similar to the settings I needed to apply to get raster handling working in my local PostGis). However, I am unable to spin up the stack from `Deploy\stacks\dynamic\stack-manager` using the established `bash ./stack.sh start RASTER-TEST` on this branch `376-error-thrown-when-trying-to-query-raster-data-in-the-stack`.
Initially I thought that might be due to having another stack running (i.e. on Digital Ocean), but the same problem persists on my local machine (i.e. without other stack running).

Is that expected? If so, please elaborate how I could best test the functionality?
Btw, as I will use the data uploader to upload some test raster data for this: Is there a data uploader functionality for Ontop mappings?

gpeb2:
> Changes look appropriate to me (i.e. very similar to the settings I needed to apply to get raster handling working in my local PostGis). However, I am unable to spin up the stack from `Deploy\stacks\dynamic\stack-manager` using the established `bash ./stack.sh start RASTER-TEST` on this branch `376-error-thrown-when-trying-to-query-raster-data-in-the-stack`. Initially I thought that might be due to having another stack running (i.e. on Digital Ocean), but the same problem persists on my local machine (i.e. without other stack running).

You might need to explicitly add an external port number i.e. `bash ./stack.sh start RASTER-TEST 3840`
> 
> Is that expected? If so, please elaborate how I could best test the functionality? Btw, as I will use the data uploader to upload some test raster data for this: Is there a data uploader functionality for Ontop mappings?

Yes, it's described in step 5 of the uploader documentation, https://github.com/cambridge-cares/TheWorldAvatar/blob/cfdc2d48fbfd7530e797d62fba10905a13a9722d/Deploy/stacks/dynamic/stack-data-uploader/README.md

kokfoongcmcl (approved):
Tested with CReDo phase 1

markushofmeister (approved):
Tested with OpenPopGrid .asc raster data and also successfully tested querying geospatial sum via Ontop mapping.

PR #378, opened 2022-10-17
37 dev district heating ontology

xuehansong:


markushofmeister (approved):

sm453 (changes requested):
The ontology and the agent should be in separate pull requests.

Regarding the ontology: I am not convinced it makes sense to distinguish between holidays and vacations in this context. Or is the intention here to distinguish between bank holidays (where _everything_ is closed) and school holidays (where children don't go to school but everything else is otherwise open)? If the latter, I believe the property definitions need to be clarified. Otherwise, I believe we only need a single notion.

Regarding the agent: You define a class called HeatNetworkAgent that does not extend any (agent) class and another class called HeatNetworkQuery that extends JPSAgent. To me it seems like HeatNetworkAgent shouldn't be called 'agent' at all, and any class that extends JPSAgent should definitely be called SomethingAgent (and not SomethingQuery). Please change this.

Your HeatNetworkQuery class seems to contain sample data. Please remove this. If this is there for testing purposes, then this should be moved to a proper unit or integration test.

The class also seems to refer to a properties file with a hard-coded path in your personal userspace. That is very bad - it is not going to work for anyone except yourself. Please fix this.

Your HeatNetworkAgent class seems to contain numerous hard-coded URLs like "localhost:48888". This is not going to be useful. Please fix it.

There are also commented out sections with references to files in your personal userspace. Please tidy this up.

markushofmeister (approved):

markushofmeister:
@xuehansong Can you please clarify why there are 2 pull requests for the ontology? Can you please consolidate AND remove all Java agent from teh ontology branch. After that we can have a final review! Otherwise, I am afraid we mix things up!

sm453 (changes requested):
Please revert all changes to java code in this pull request (assuming that any such changes should be part of PR #339).

Regarding the ontology, I don't believe changing "Holiday" to "PublicHoliday" makes things any clearer. Could you please, at the very least, provide in the "comment" column sufficient information so that one has a chance to understand what the difference is between a holiday and a vacation?

markushofmeister:
Regarding the representation of `calendar effects` (required for forecasting). What do you think about the following amendments (Seb's comment about meaningful "comments" still apply in the actual implementation!)
1) Remove the `hasDate` as this would effectively store a time series of redundant information, i.e. hourly time steps in the time column and the date (i.e. extractable from that more granular time information) in the value column
2) Define `IsPublicHoliday` and `IsSchoolVacation` as subclasses of `CalendarEffect`, which should have a clear description that it's values are stored as booleans. So, the timeseries would be time:boolean. Otherwise, one could add another triple along the lines `CalendarEffect - hasValue - boolean`; however, that somehow feelds odd to me.
3) Add the representative Location for the calendar effect, as especially school vacations differ quite significantly between various regions

https://miro.com/app/board/uXjVOhnB9_4=/
image

xuehansong:
@sm453 I have already updated the csv and owl file with amended CalendarEffect part. Please advise for further amendment if needed

sm453:
I'm confused now. This PR still contains agent as well as ontology changes, but you closed it several days ago. In light of this, I cannot make sense of your comment. Please clarify.

xuehansong:
@sm453 sorry for the confusion. I am not sure why I closed #378, maybe by mistake, then I created new pull request #402 on last Friday. Now I cannot revert to reopen #378, since it says there is already an open pull request #402.

sm453:
Ok, fine. It does not make sense to open multiple pull requests (to main) from one and the same branch - that's why GitHub doesn't let you do it. Let's continue on #402, then, but please note the comments posted here still apply.

PR #381, opened 2022-10-19
Dev property sales instantiation agent

markushofmeister:


sm453 (approved):
As commented elsewhere, the use of the term "knowledgebase" is deprecated (amongst ourselves). I acknowledge that it may be impractical to change this here, because it's all over the place, but for future reference, please do not use this term.

markushofmeister:
> As commented elsewhere, the use of the term "knowledgebase" is deprecated (amongst ourselves). I acknowledge that it may be impractical to change this here, because it's all over the place, but for future reference, please do not use this term.

Well noted. I was not aware of this as it is still widely used (even among "us").
I will go ahead and merge this branch back to main. However, I'll keep this branch open in order to incorporate a few more enhancements (i.e. update logging, derivation markup method, etc.). I'll try to eliminate the usage of `knowledgebase` as good as I can before creating another PR for the coming changes.

PR #383, opened 2022-10-19
DTVF v3.1.0

mdhillmancmcl:
This new version of the framework brings in minor features, bug fixes, and documentation updates.  A full list is detailed below.

Bug Fixes:
- Scroll bar sizing in the Layers tree
- Default visibility of layers in Cesium visualisations
- Toggling visibility for KML layers in Cesium visualisations
- Allow specification of positions for 3D tile data sources
- Fixed hover effects for features containing properties with null values

Features:
- Imagery layer with labelled satellite data
- Ability to define default expansion states in the layer tree
- Ability to define layer order globally (i.e. irrespective of grouping)
- Can now use PAGE_UP and PAGE_DOWN buttons to zoom in Cesium visualisations
- Can now select the feature name to fly to it in Cesium visualisations

Misc:
- Updated GitHub wiki content
- Updated READMEs
- Refined camera options


sm453:
Please remove the file \web\digital-twin-vis-framework\library\utils\python.log. It's 81 MB...

sm453:
Please update the DTVF version number in web/digital-twin-vis-framework/README.md.

sm453:
In web/digital-twin-vis-framework/README.md, you seem to have changed "CesiumJS can handle 2D data" to "CesiumJS can handle 2D raster data" and you have added "at the time of writing we have not implemented functionality to support 2D vector data". I find this highly misleading. The Cesium version handles vector data just fine (WMS layers of points, lines, and polygons). Do you mean vector layers are not currently served as vector to the front-end, but instead as image/png? If so, I would also add that that applies to both the Cesium and Mapbox versions. Either way, please clarify.

sm453:
You have added
`MapHandler.MAP_USER = "";`
to the Cesium version of index.html, but not the Mapbox version. Is this intentional?

mdhillmancmcl:
> Please remove the file \web\digital-twin-vis-framework\library\utils\python.log. It's 81 MB...

Removed, and added an ignore file to prevent future *.log files being committed.

mdhillmancmcl:
> Please update the DTVF version number in web/digital-twin-vis-framework/README.md.

Done, nice catch.

mdhillmancmcl:
> In web/digital-twin-vis-framework/README.md, you seem to have changed "CesiumJS can handle 2D data" to "CesiumJS can handle 2D raster data" and you have added "at the time of writing we have not implemented functionality to support 2D vector data". I find this highly misleading. The Cesium version handles vector data just fine (WMS layers of points, lines, and polygons). Do you mean vector layers are not currently served as vector to the front-end, but instead as image/png? If so, I would also add that that applies to both the Cesium and Mapbox versions. Either way, please clarify.

I have updated the README, it reads clearer to me now but if you'd like further changes then please post on this PR.

mdhillmancmcl:
> I assume commenting out the (previously uncommented) search bits in web/digital-twin-vis-framework/library/src/ts/manager.ts is intentional?

Yes, this is a feature that's still in progress. It will be continued in v3.2.0 (rather than holding up this PR).

mdhillmancmcl:
> You have added `MapHandler.MAP_USER = "";` to the Cesium version of index.html, but not the Mapbox version. Is this intentional?

Funnily enough, this is less critical in the Mapbox visualisation than the Cesium one. For consistency I have added it to the template `index.html` for the Mapbox visualisation too.

sm453 (approved):
I have tested all bugfixes and added features, and can confirm that they are working. All comments have been addressed.

PR #385, opened 2022-10-20
dev-alignOntoTimeSeriesKBNamespace

markushofmeister:
Updated namespace base IRI for all time series instances to be created by the client to follow new format, i.e. "https://www.theworldavatar.com/kb/ontotimeseries/"


sm453 (changes requested):
It should be "kg" instead of "kb" - we deprecated the term "knowledgebase" (amongst ourselves) years ago.

sm453 (approved):

Issue #321, opened 2022-09-07
PostGISClient driver not found


kokfoongcmcl:

gpeb2:
Reopening as the underlying cause has potentially been identified. The version of the PostGIS library is quite old as they changed the groupId for the next version and  didn't notice. This older version depends on an old version of the Postgres driver library, which also has since changed groupId. This change in groupId for the Postgres driver meant that both it and the newer version are getting loaded with the code randomly picking one to use at runtime.

kokfoongcmcl:
Tested the proposed solution but does not work

Issue #389, opened 2022-10-21
RemoteStoreClient closes the shared connection after each query


jb2197:
The latest RemoteStoreClient now closes the connection after each query, this causes an issue if an agent is querying from the ontop endpoint repeatedly

_Originally posted by @kokfoongcmcl in https://github.com/cambridge-cares/TheWorldAvatar/issues/388#issuecomment-1285512449_
      

Besides the above issue, if the agent querying from blazegraph endpoint with multi-thread, there exists a situation where the `conn.close()` will affect other queries. An example is `java.sql.SQLException: The Statement is closed` exception below:

```
Traceback (most recent call last):
File "~/twa/TheWorldAvatar/JPS_BASE_LIB/python_derivation_agent/pyderivationagent/kg_operations/sparql_client.py", line 74, in performQuery
response = self.kg_client.execute(query)
File "~/anaconda3/envs/twa/lib/python3.8/site-packages/py4j/java_gateway.py", line 1304, in __call__
return_value = get_return_value(
File "~/anaconda3/envs/twa/lib/python3.8/site-packages/py4j/protocol.py", line 326, in get_return_value
raise Py4JJavaError(
py4j.protocol.Py4JJavaError: An error occurred while calling o9.execute.
: uk.ac.cam.cares.jps.base.exception.JPSRuntimeException: The Statement is closed
at uk.ac.cam.cares.jps.base.query.RemoteStoreClient.executeQuery(RemoteStoreClient.java:476)
at uk.ac.cam.cares.jps.base.query.RemoteStoreClient.execute(RemoteStoreClient.java:434)
at jdk.internal.reflect.GeneratedMethodAccessor2.invoke(Unknown Source)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.base/java.lang.reflect.Method.invoke(Method.java:566)
at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
at py4j.Gateway.invoke(Gateway.java:282)
at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
at py4j.commands.CallCommand.execute(CallCommand.java:79)
at py4j.GatewayConnection.run(GatewayConnection.java:238)
at java.base/java.lang.Thread.run(Thread.java:829)
[...]

PR #393, opened 2022-10-24
179 dev remote store client connection

jb2197:
In this PR, three changes were made:

- The commit that updates the version of the "postgis-jdbc" dependency in the jps-base-lib is cherry-picked so that it can be merged to the main branch quickly to fix issue #321.
- The Connection and Statement object stored as fields in RemoteStoreClient is removed, instead, they are generated following try-with-resources.
- The complete SPARQL query/update string is added to the error message when executeQuery/executeUpdate runs into an exception.


jb2197:
As now the Connection and Statement object is properly handled, this PR should also close #389.

kokfoongcmcl:
Please also update the http post methods

kokfoongcmcl:
This needs some testing, but I think you can also remove the line `Class.forName(driver);` in the RemoteRDBStoreClient, if the driver update fixes the issue

gpeb2 (approved):

PR #398, opened 2022-10-26
Dev average square metre price agent

markushofmeister:
First Derivation Agent designed to work within the Stack, incl. localised integration tests


Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/agent/avgprice_estimation.py
```diff
@@ -0,0 +1,285 @@
...
+        -> relevant for asynchronous derivation
+
+        Arguments:
+            inputs {dict} -- Dictionary of inputs with input concepts as keys and values as list
+
+        Returns:
+            postcode_iri {str}, ppi_iri {str}, tx_records {list}
+        """
+
+        #TODO: Potentially add derivationIRI to logger and exception message
```
jb2197, line 68: This is currently not easily possible, derivationIRI is hidden to `process_request_parameters`. Maybe we can change to `process_request_parameters(self, derivation_iri, derivation_inputs, derivation_outputs)` in the next version of pyderivationagent.

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/kg_operations/tsclient.py
```diff
@@ -0,0 +1,72 @@
...
+                       (see: https://www.jooq.org/javadoc/dev/org.jooq/org/jooq/impl/SQLDataType.html)
+            rdb_url (str): URL of relational database
+            rdb_user (str): Username for relational database
+            rdb_password (str): Password for relational database
+        """
+
+        # 1) Create an instance of a RemoteStoreClient (to retrieve RDB connection)
+        try:
+            connection = TSClient.jpsBaseLibView.RemoteRDBStoreClient(rdb_url, rdb_user, rdb_password)
+            self.conn = connection.getConnection()
```
jb2197, line 43: Instead of storing the connection object, it might be better to create a connection every time you start a query. This might be relevant #397. But I also wonder how does try-with-resources work with a python wrapper?

Agents/AverageSquareMetrePriceAgent/setup.py
```diff
@@ -0,0 +1,34 @@
...
+setup(
+    name='avgsqmprice',
+    version='0.0.1',
+    author='Markus Hofmeister',
+    author_email='mh807@cam.ac.uk',
+    license='MIT',
+    description="The `avgsqmprice` agent calculates the average square metre price of properties for a particular postcode and populates the result to knowledge graph as part of The World Avatar project.",
+    long_description=open('README.md').read(),
+    long_description_content_type="text/markdown",
+    packages=find_packages(exclude=("tests")),
```
jb2197, line 12: `find_namespace_packages` is now recommended for python 3.x projects

Agents/AverageSquareMetrePriceAgent/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# define agent logging dependency here as there are issues including it in install_requires in setup.py
```
jb2197, line 1: I would recommend pinning the working version of the list of dependencies here to be installed from Dockerfile for production.

jb2197:
I have not checked spinning up the agent in a stack but tested the integration test works fine on my machine.

You may want to add a reminder for installing `libpq-dev`, `python-dev`, and `gcc` to the system if one is developing/testing in WSL2 as they seem to be required to build the `psycopg2` package.

I have two questions regarding the stack and potential dockerised integration test:
- Is `docker stack deploy` not able to read `env_file` declared in `docker-compose.yml`?
- Can one mock the imports when spinning up the docker container of the agent? Like how you did with the localised integration test.

Otherwise it looks good to me.

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/agent/avgprice_estimation.py
markushofmeister: Elaborated the comment in commit `[f813e37](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/f813e37ede1bb95d38dd8c67ef1b03cf675734fc)` to keep track of this

Agents/AverageSquareMetrePriceAgent/setup.py
markushofmeister: Addressed in commit [f813e37](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/f813e37ede1bb95d38dd8c67ef1b03cf675734fc)

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/kg_operations/tsclient.py
markushofmeister: Starting a new connection for every query was the default design of the TimeSeries client. Last month it got re-designed to use a connection object in order to 1) be able to use the access agent also for TimeSeries client with centrally referenced connection objects and 2) to improve performance. Currently, it seems this design becomes partially revoked; however, the methods with connection object will stay in place.

Hence, as this agent relies on recurring calls to the same Relational Database, I believe the implementation with the connection object seems suitable. Therefore, the try-with-resource should not be applicable.

markushofmeister:
> I have not checked spinning up the agent in a stack but tested the integration test works fine on my machine.
> 
> You may want to add a reminder for installing `libpq-dev`, `python-dev`, and `gcc` to the system if one is developing/testing in WSL2 as they seem to be required to build the `psycopg2` package.
 
Addressed by [5da0a18](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/5da0a181fa1bb8a946959ede7fdc482711260d6a)

> I have two questions regarding the stack and potential dockerised integration test:
> 
> * Is `docker stack deploy` not able to read `env_file` declared in `docker-compose.yml`?

The official [Docker Compose documentation](https://docs.docker.com/compose/environment-variables/) states:
`Note when using docker stack deploy:`
`The .env file feature only works when you use the docker-compose up command and does not work with docker stack deploy.`

> * Can one mock the imports when spinning up the docker container of the agent? Like how you did with the localised integration test.

This could potentially work, but would that make much of a difference? Would this improve the test quality, i.e. testing functionality currently not tested, or do you simply have ease of execution in mind?

> 
> Otherwise it looks good to me.

Agents/AverageSquareMetrePriceAgent/requirements.txt
jb2197: Requirements files and `setup.py` are designed for different purposes. The former tends to be as specific as possible for reproducibility and production, whereas the latter normally gives a wide range to not limit the choice of packages in development. If you let the pip decide the version of packages, there's a chance it pulls a version that potentially breaks the code. You can read more about these two things here https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

jb2197:
> > * Is `docker stack deploy` not able to read `env_file` declared in `docker-compose.yml`?
> 
> The official [Docker Compose documentation](https://docs.docker.com/compose/environment-variables/) states: `Note when using docker stack deploy:` `The .env file feature only works when you use the docker-compose up command and does not work with docker stack deploy.`

Thanks for info on this, I didn't know about it :)

> > * Can one mock the imports when spinning up the docker container of the agent? Like how you did with the localised integration test.
> 
> This could potentially work, but would that make much of a difference? Would this improve the test quality, i.e. testing functionality currently not tested, or do you simply have ease of execution in mind?

The dockerised test is mainly to check if all environment is set up correctly in the docker image, which might not always be the case. Also, I guess you are developing in the windows environment, so it's worth checking in linux.

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/kg_operations/tsclient.py
jb2197: Storing a connection object can lead to some funny behaviour when running an agent in multi-thread - if one thread closes the connection then it will affect all other running queries. Given that the derivation agent allows setting `MAX_THREAD_MONITOR_ASYNC_DERIVATIONS` in config, it seems to me best to use one connection object per query, or at least per thread.

kokfoongcmcl: Yes, do not store connection objects, I believe in python it will be try-catch-finally, where you do connection.close() in the finally block

markushofmeister: Address by [829cc40](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/829cc406fac7ff3f5891771a3fd12f241b62679c)

Agents/AverageSquareMetrePriceAgent/tests/conftest.py
```diff
@@ -0,0 +1,373 @@
...
+    ):
+        agent_config = config_derivation_agent(AGENT_ENV)
+        agent = AvgSqmPriceAgent(
+            register_agent=agent_config.REGISTER_AGENT if not register_agent else register_agent,
+            agent_iri=agent_config.ONTOAGENT_SERVICE_IRI if not random_agent_iri else 'http://agent_' + str(uuid.uuid4()),
+            time_interval=agent_config.DERIVATION_PERIODIC_TIMESCALE,
+            derivation_instance_base_url=agent_config.DERIVATION_INSTANCE_BASE_URL,
+            kg_url=host_docker_internal_to_localhost(QUERY_ENDPOINT),
+            kg_update_url=host_docker_internal_to_localhost(UPDATE_ENDPOINT),
+            # NOTE For agent endpoint, we keep this as it is for now (i.e. start with http://host.docker.internal)
```
markushofmeister, line 210: @jb2197: This is based on your example agent. I do not fully understand how that properly tests the dockerised agent as the endpoints will always be `localhost` and not use the Docker container internal connections to actually use/test the internal agent connections. 
Your tests and also my tests pass (both locally run as well as dockerised agent ones); however, I have the impression that the tests ar enot fully decoupled from one another. Any clarification and more elaborate description of the design might be very helpful

markushofmeister (commented):
The tests pass when running them individually; however, running all 4 parametrised tests from command line frequently creates Docker issues due to ambiguous network detection, i.e. several Docker networks with the same name being created simultaneously. Have we een that before and know how to resolve this?

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/kg_operations/tsclient.py
```diff
@@ -0,0 +1,86 @@
...
+            timeclass: Java time class objects supported by PostgreSQL
+                       (see: https://www.jooq.org/javadoc/dev/org.jooq/org/jooq/impl/SQLDataType.html)
+            rdb_url (str): URL of relational database
+            rdb_user (str): Username for relational database
+            rdb_password (str): Password for relational database
+        """
+
+        # 1) Create an instance of a RemoteStoreClient (to retrieve RDB connection)
+        try:
+            self.conn = TSClient.jpsBaseLibView.RemoteRDBStoreClient(rdb_url, rdb_user, rdb_password)
```
gpeb2, line 44: This variable should be renamed `self.conn`.

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/kg_operations/tsclient.py
gpeb2: The `@contextmanager` decorator, along with the `try`-`catch` and `yield`, seems to do what we need, so good find.

gpeb2:
> The tests pass when running them individually; however, running all 4 parametrised tests from command line frequently creates Docker issues due to ambiguous network detection, i.e. several Docker networks with the same name being created simultaneously. Have we een that before and know how to resolve this?

I've not seen this quite so explicitly but I have seen a few issues with Docker containers randomly clashing during tests. Would it make sense/be possible to initialise the container(s) once for all of the parametrisations?

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/agent/avgprice_estimation.py
jb2197: With the latest version of [`pyderivationagent==1.3.0`](https://pypi.org/project/pyderivationagent/1.3.0/) (you also need `py4jps>=1.0.27`), you can now get the derivation IRI by `derivation_inputs.getDerivationIRI()`

jb2197:
> The tests pass when running them individually; however, running all 4 parametrised tests from command line frequently creates Docker issues due to ambiguous network detection, i.e. several Docker networks with the same name being created simultaneously. Have we een that before and know how to resolve this?

How are you running them? I am running all tests via `pytest -s --docker-compose=./docker-compose-test.yml` in WSL2 and only one set of docker containers is created for parameterised tests.

> I've not seen this quite so explicitly but I have seen a few issues with Docker containers randomly clashing during tests. Would it make sense/be possible to initialise the container(s) once for all of the parametrisations?

In theory, this should already be the case as the scope for the pytest fixtures involved are `session` and `module`.

Agents/AverageSquareMetrePriceAgent/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Asynchronous derivation operation
+Once the Agent is deployed, it periodically (every week, defined by `DERIVATION_PERIODIC_TIMESCALE`) checks the derivation that `isDerivedUsing` itself (parameter `ONTOAGENT_SERVICE_IRI`) and acts based on the status associated with that derivation. Although the [Derivation Agent] suggests the use of `.env` files to specify environment variables for agent configurations, this approach does not work properly with Docker stacks, i.e. `docker stack deploy`. Hence, the agent configuration is moved to the [docker compose file] instead.
+
+## Prior derivation markup
+
+For the Agent to detect outdated information, a proper mark up of the relevant derivation inputs (i.e. *pure* inputs) is required. (Please note, that another pre-requisite for detecting derivation inputs is the registration of the agent in the KG, i.e. `REGISTER_AGENT=true` in the [docker compose file].) The following methods from the `pyderivationagent` package shall be used to mark up derivation inputs within the KG (for illustration purposes only):
+```bash
+# Retrieve derivation client from derivation agent
+deriv_client = agent.derivationClient
```
jb2197, line 151: If you use the latest `pyderivationagent==1.3.0`, it has been updated to `agent.derivation_client` to follow the python naming convention.

Agents/AverageSquareMetrePriceAgent/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+# Add and update time stamp to all pure input instances (i.e. inputIRIs)
+deriv_client.addTimeInstance(inputIRI)
```
jb2197, line 154: With the latest `pyderivationagent==1.3.0`, the timestamp for pure inputs will be added automatically when marking up the derivations, so you don't need to manually add them here.

Agents/AverageSquareMetrePriceAgent/README.md
```diff
@@ -0,0 +1,255 @@
...
+deriv_client.updateTimestamp(inputIRI)
```
jb2197, line 155: As we discussed, updating the timestamp here will be sufficient when your input agent updates the values.

Agents/AverageSquareMetrePriceAgent/tests/test_example_agent.py
```diff
@@ -0,0 +1,172 @@
...
+    # Verify inputs (i.e. derived from)
+    for i in inputs:
+        for j in inputs[i]:
+            assert j in derivation_input_set
+            derivation_input_set.remove(j)
+    assert len(derivation_input_set) == 0
+
+    print("All check passed.")
+
+    # Shutdown the scheduler to clean up if it's local agent test (as the doe_agent scheduler must have started)
```
jb2197, line 170: You may want to change the "doe_agent" bit in the comment.

Agents/AverageSquareMetrePriceAgent/requirements.txt
markushofmeister: Pinned fox version number in [9a93285](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/9a932853c74dce979f8618b402de1bc46559c8d6)

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/agent/avgprice_estimation.py
markushofmeister: Included in [79b346f](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/79b346fa5a11c56f66186d806360318ce641955c)

Agents/AverageSquareMetrePriceAgent/avgsqmpriceagent/kg_operations/tsclient.py
markushofmeister: Addressed by [d2c4b29](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/d2c4b29a79ee004e72df5814f734354f4782d85e)

Agents/AverageSquareMetrePriceAgent/README.md
markushofmeister: Addressed in [c16916b](https://github.com/cambridge-cares/TheWorldAvatar/pull/398/commits/c16916b01d0ccd8a7a440be8ebf2e82bf3627acf). With marking up you refer to the `createAsyncDerivationForNewInfo` method for example, correct?

jb2197: Correct. This refers to all methods that create derivations, as listed in the section `## Methods for creating derivation markup ##` of `pyderivationagent.PyDerivationClient`, e.g. `createDerivation`, `createDerivationWithTimeSeries`, etc.

Agents/AverageSquareMetrePriceAgent/tests/test_example_agent.py
```diff
@@ -0,0 +1,202 @@
...
+    assert len(postcode) == 1
+    assert postcode[0] == expected_postcode
+    # Verify price
+    assert len(price) == 1
+    assert price[0] == expected_avg
+    # Verify inputs (i.e. derived from)
+    for i in inputs:
+        for j in inputs[i]:
+            assert j in derivation_input_set
+            derivation_input_set.remove(j)
```
jb2197, line 195: This line (`derivation_input_set.remove(j)`) effectively removes the element from the list `cf.DERIVATION_INPUTS_1` as they share the same memory address, therefore `derivation_input_set` is passed in as an empty list in your third test parameterised test. This results in creating derivation with no input IRIs, hence the error you see when running multiple tests in one-go.

Maybe you want to use `copy.deepcopy(derivation_input_set)` to create a new list for each test and remove element from there.

markushofmeister: Updated as suggested and now all tests pass when running them 1) individuall, 2) all together via command line using `pytest -s --docker-compose=./docker-compose-test.yml` as well as 3) all together using Testing in VSCode.

jb2197 (approved):
Could you please change the base branch from `dev-PropertySalesInstantiationAgent` to `main`? As `dev-PropertySalesInstantiationAgent` is already merged back to `main`, this PR doesn't seem to carry commits from other agents anymore.

Other than that, all looks good to me.

PR #400, opened 2022-10-27
dev-dtvf-3.2.1

mdhillmancmcl:
Added additional null checks for potential cases in which features may have no property objects. Should fix/help to further diagnose an issue that @markushofmeister has reported.


markushofmeister:
The visualisation loads both buildings and reporting stations properly. However, when trying to click on feature to trigger the FeatureInfoAgent, I still the following error
image

markushofmeister (changes requested):
The visualisation loads both buildings and reporting stations properly. However, when trying to click on feature to trigger the FeatureInfoAgent, I still the following error. The FeatureInfoAgent queries work in the Blazegraph GUI
image

mdhillmancmcl:
This PR has been closed without merging as further bugs have been discovered and it's easier to deal with this in the upcoming feature release (v3.3.0).

PR #402, opened 2022-10-28
37 dev district heating ontology

xuehansong:
The updated district heating ontology with amended "CalendarEffect" part


xuehansong:
> This is the third ( ! ) branch/pull request relating to district heating that contains both agent as well as ontology changes. This is disastrous. Can you please get in touch through separate channels in order to clean up this mess?

I have deleted the agent files in this ontology branch. I thought I have deleted the agent files long ago but just realized when I did this on the web browser, I did not commit it. Now I learn that even delete files on the web browser, still need to commit it. That's why these agent files are still there. Again very sorry about this. Thank you

markushofmeister (approved):
I had another look at the `.owl` file and believe the ontology looks alright. Only question: Have you updated the MIRO board accordingly? I just realised that `UnitRate` has forecasted values in the ontology (which they should); however, that's not represented in MIRO.

markushofmeister:
@sm453: When reviewing the ontology, can you please check the representation of calendar effects (particularly). That has been updated mostly compared to previous versions

xuehansong:
> I had another look at the `.owl` file and believe the ontology looks alright. Only question: Have you updated the MIRO board accordingly? I just realised that `UnitRate` has forecasted values in the ontology (which they should); however, that's not represented in MIRO.

I hear you. Yes, the forecast of `UnitRate`  was not there. I have updated it in the MIRO already. Thank you

sm453:
> @sm453: When reviewing the ontology, can you please check the representation of calendar effects (particularly). That has been updated mostly compared to previous versions

I think this is good enough for now.

sm453:
Regarding tiered unit price contracts, I think it is a shame that `Tier hasUnitPrice HeatUnitPrice` rather than `Tier hasUnitPrice UnitPrice`. That one statement makes it specific to heat - everything else is completely generic, and one could easily re-use this for, say, electricity, etc. Can you change this?

xuehansong:
> Regarding tiered unit price contracts, I think it is a shame that `Tier hasUnitPrice HeatUnitPrice` rather than `Tier hasUnitPrice UnitPrice`. That one statement makes it specific to heat - everything else is completely generic, and one could easily re-use this for, say, electricity, etc. Can you change this?

Yes, I have changed it. Thank you.

sm453 (approved):
Good enough.

PR #403, opened 2022-10-28
dev-Ifc2OntoBIM-agent

qhouyee:
A new agent that IFC files into OntoBIM Aboxes


sm453 (approved):
Extensive use of string literals all over the place is problematic. If anyone ever needs to change anything in this code, they will have one heck of a find-and-replace exercise in front of them... I'm not suggesting to change this as part of this pull request, but for future reference, please consider defining constants, so that (heavily re-used) strings are defined in one place, and one place only.

PR #405, opened 2022-10-31
dev-OntoBIM

qhouyee:
A new ontology for BIM entities


sm453 (changes requested):
Would you also be able to add the csv file from which the owl file can be generated using our TBox generation tool?

sm453 (approved):

PR #416, opened 2022-11-01
181 dev add derivation error status

jb2197:
In this PR, any exception thrown during handling asynchronous derivation will be handled by flipping the derivation status to "Error" and logging the error message/stack traces as `rdfs:comment` of the status instance. Both unit and integration tests were added.

Once approved, `OntoDerivation` TBox will also be updated.


sm453 (approved):
Fine from my side. @kokfoongcmcl may want to have a closer look...

kokfoongcmcl:
I don't fully understand why do you want to store this in the KG, I think it should be just logged in the log file. 

If you really want to do this, I'd prefer returning List<Derivation> with an extra field in the Derivation class rather than Map<String, String>. Even better if there's some sort of class to store stack traces

jb2197:
> I don't fully understand why do you want to store this in the KG, I think it should be just logged in the log file.

Compared to just hanging as "InProgress" (where most of the failures happen), changing the status to Error better reflects the situation of derivation. Storing the error message in KG thus potentially helps programmatical recovery of the derivation in future (although not any time soon).

> If you really want to do this, I'd prefer returning `List<Derivation>` with an extra field in the Derivation class rather than Map<String, String>. Even better if there's some sort of class to store stack traces

I have changed `DerivationClient::getDerivationsInErrorStatus` to return `List<Derivation>`, the error message can be obtained via `Derivation::getErrMsg`.

As discussed, I am proceeding with merging now.

PR #425, opened 2022-11-07
1423 dev modify instantiation in timeseriesclient

mehalaga:


JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -78,35 +80,52 @@ public void setKBClient(TripleStoreClientInterface kbClient) {
...
+		}
+		else if(type.equals(TimeSeriesSparql.Instantaneous)){
+			tsIRI = TimeSeriesSparql.Instantaneous + "Timeseries_" + UUID.randomUUID();
+		}
+		else if(type.equals(TimeSeriesSparql.Average)){
+			tsIRI = TimeSeriesSparql.Average + "Timeseries_" + UUID.randomUUID();
+		}
+		else {
+			throw new JPSRuntimeException(exceptionPrefix + "TimeSeries type: " + type + " is invalid");
+		}
```
kokfoongcmcl, line 122: Avoid these sort of if-else codes

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -78,35 +80,52 @@ public void setKBClient(TripleStoreClientInterface kbClient) {
...
-		} catch (SQLException e) {
-			// this ensures rdfClient.initTS always has a valid string
-			LOGGER.warn(e.getMessage());
-			LOGGER.warn("Failed to get RDB URL from connection object, setting RDB URL to = \"\"");
-			rdbURL = "";
-		}
-		
-   		try {
-   			rdfClient.initTS(tsIRI, dataIRIs, rdbURL, timeUnit);
+			rdfClient.initTS(tsIRI, dataIRIs, conn.getMetaData().getURL(), timeUnit, type, duration, unit);
```
kokfoongcmcl, line 128: This should use rdbURL
`try {
			rdbURL = conn.getMetaData().getURL();
		} catch (SQLException e) {
			// this ensures rdfClient.initTS always has a valid string
			LOGGER.warn(e.getMessage());
			LOGGER.warn("Failed to get RDB URL from connection object, setting RDB URL to = \"\"");
			rdbURL = "";
		}`

kokfoongcmcl:
General comment on TimeSeriesClient class:
1) Don't deprecate the existing methods, instead create new methods, e.g. initAverageTimeSeries, initInstantaneousTimeSeries
2) Avoid inputs as string, i.e. the "type" in your methods

kokfoongcmcl:
What's the difference between instantaneous time series with the "normal" time series?

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
mehalaga: should I replace it with switch case? if not, what would be the alternative?

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
kokfoongcmcl: Have a look at the "deprecated" initTimeSeries

mehalaga:
> General comment on TimeSeriesClient class:
> 
> 1. Don't deprecate the existing methods, instead create new methods, e.g. initAverageTimeSeries, initInstantaneousTimeSeries
> 2. Avoid inputs as string, i.e. the "type" in your methods

Do we want to create individual instantiation methods for each type of time series?

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
kokfoongcmcl: see comment below, have different methods, initAverageTimeSeries, this is what I prefer in this case. Alternative will be to use enums

kokfoongcmcl:
> > General comment on TimeSeriesClient class:
> > 
> > 1. Don't deprecate the existing methods, instead create new methods, e.g. initAverageTimeSeries, initInstantaneousTimeSeries
> > 2. Avoid inputs as string, i.e. the "type" in your methods
> 
> Do we want to create individual instantiation methods for each type of time series?

Yes for public facing ones, they all should route to an internal method

mehalaga:
> What's the difference between instantaneous time series with the "normal" time series?

I think its just the subclassification

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -143,30 +180,120 @@ public boolean checkTimeUnitExists(String tsIRI) {
...
+		else if(type.equals(CumulativeTotal)){
+			if (checkCumulativeTotalTimeSeriesExists(timeSeriesIRI)) {
+				throw new JPSRuntimeException(exceptionPrefix + "Cumulative Total Time series " + timeSeriesIRI + " already in the Knowledge Graph");
+			}
+			// define type
+			modify.insert(tsIRI.isA(CumulativeTotalTimeSeries));
+		}
+		else if(type.equals(Instantaneous)){
+			if (checkInstantaneousTimeSeriesExists(timeSeriesIRI)) {
+				throw new JPSRuntimeException(exceptionPrefix + "Instantaneous Time series " + timeSeriesIRI + " already in the Knowledge Graph");
```
kokfoongcmcl, line 244: Similar comment to a previous 1, have separate methods in the TimeSeriesClient, and pass the IRI here

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -143,30 +180,120 @@ public boolean checkTimeUnitExists(String tsIRI) {
...
-		// Check that the time series IRI is not yet in the Knowledge Graph
-		if (checkTimeSeriesExists(timeSeriesIRI)) {
-			throw new JPSRuntimeException(exceptionPrefix + "Time series " + timeSeriesIRI + " already in the Knowledge Graph");
+		ModifyQuery modify = Queries.MODIFY();
+		// set prefix declarations
+		modify.prefix(prefix_ontology, prefix_kb, prefix_time);
+
+		if (type.equals(null)){
+			// Check that the time series IRI is not yet in the Knowledge Graph
+			if (checkTimeSeriesExists(timeSeriesIRI)) {
```
kokfoongcmcl, line 222: checks are not required if you are using UUID

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -175,114 +302,277 @@ protected void initTS(String timeSeriesIRI, List<String> dataIRI, String dbURL,
...
+	 * @param timeSeriesIRI StepwiseCumulativeTimeSeries IRI provided as string
+	 * @return True if a StepwiseCumulative time series instance with the tsIRI exists, false otherwise
+	 */
+	private boolean checkStepwiseCumulativeTimeSeriesExists(String timeSeriesIRI) {
+		String query = String.format("ask {<%s> a <%s>}", timeSeriesIRI, (ns_ontology + "StepwiseCumulativeTimeSeries"));
+		kbClient.setQuery(query);
+		return kbClient.executeQuery().getJSONObject(0).getBoolean("ASK");
+	}
+
+	protected void bulkInitTS(List<String> tsIRIs, List<List<String>> dataIRIs, String rdbURL, List<String> timeUnit, List<String> type, List<Duration> durations, List<ChronoUnit> units) {
```
kokfoongcmcl, line 404: This method should only interact with the triple-store once (one sparql update), you are sending lots of queries here (e.g. checkTimeSeriesExist) which will slowthings down

kokfoongcmcl:
Further to the discussion, uk.ac.cam.cares.jps.base.derivationValuesPattern.addValuePairForMultipleVariables is probably what you need

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -86,43 +92,43 @@ public void setKBClient(TripleStoreClientInterface kbClient) {
...
+		else if(type.equals(Type.CUMULATIVETOTAL)){
+			timeseriesType = TimeSeriesSparql.CumulativeTotalTimeSeries;
+			tsIRI = TimeSeriesSparql.CumulativeTotal+  "Timeseries_" + UUID.randomUUID();
+		}
+		else if(type.equals(Type.INSTANTANEOUS)){
+			timeseriesType = TimeSeriesSparql.InstantaneousTimeSeries;
+			tsIRI = TimeSeriesSparql.Instantaneous + "Timeseries_" + UUID.randomUUID();
+		}
+		else {
+			throw new JPSRuntimeException(exceptionPrefix + "TimeSeries type: " + type + " is invalid");
```
kokfoongcmcl, line 129: You don't need this, it's impossible to define an invalid type with enum

mehalaga: removed it

kokfoongcmcl:
I spotted duplicate code for initTimeSeries on line 104 and 1060, I believe they can be combined, e.g. having the one on 1060 call the one on 104

kokfoongcmcl:
Same applies for bulkInitTimeSeries

mehalaga:
> 

they code in lin

> I spotted duplicate code for initTimeSeries on line 104 and 1060, I believe they can be combined, e.g. having the one on 1060 call the one on 104

what would be the enum Type in this case needed to call the method in 104

kokfoongcmcl:
You can create an extra type, or pass null, not sure if that works, you can decide

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -150,21 +187,49 @@ public void bulkInitTimeSeries(List<List<String>> dataIRIs, List<List<Class<?>>>
...
-    public void bulkInitTimeSeries(List<List<String>> dataIRIs, List<List<Class<?>>> dataClass, List<String> timeUnit, Integer srid, Connection conn) {
-        // create random time series IRI
-    	List<String> tsIRIs = new ArrayList<>(dataIRIs.size());
-    	
-    	for (int i = 0; i < dataIRIs.size(); i++) {
-    		String tsIRI = TimeSeriesSparql.ns_kb + "Timeseries_" + UUID.randomUUID();
-    		tsIRIs.add(i, tsIRI);
-    	}
-    	
+    public void bulkInitTimeSeries(List<List<String>> dataIRIs, List<List<Class<?>>> dataClass, List<String> timeUnit, Integer srid, Connection conn, List<Type> type, List<Duration> durations, List<ChronoUnit> units) {
```
kokfoongcmcl, line 194: should this be private?

mehalaga: this is the public facing method used for bulk instantiation. why would we want it to be private?

kokfoongcmcl: This contains srid and the duration parameters (which will be an unlikely combination to be called outside), I thought this is the private method we agreed on that will be called by the other public facing ones?

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -172,13 +172,28 @@ else if(type.equals(Type.GENERAL)){
...
+	 * similar to initTimeSeries, but uploads triples in one connection
+	 * @param dataIRIs
+	 * @param dataClass
+	 * @param timeUnit
+	 * @param srid
+	 * @param conn
+	 * @param type
+	 * @param units
+	 */
+	public void bulkInitTimeSeries(List<List<String>> dataIRIs, List<List<Class<?>>> dataClass, List<String> timeUnit, Integer srid, Connection conn, List<Type> type, List<ChronoUnit> units) {
```
kokfoongcmcl, line 193: In what circumstances do we need this combination? I thought Duration and ChronoUnit need to be provided as a pair

mehalaga: this is a mistake, I'll rectify it.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -41,27 +45,104 @@
...
+
+	// Fields for class specific exceptions
 	private final String exceptionPrefix = this.getClass().getSimpleName() + ": ";
 
-    /**
+	//Time Series Type
+	public static final String Average = ns_ontology + "Average";
+	public static final String StepwiseCumulative = ns_ontology + "StepwiseCumulative";
+	public static final String CumulativeTotal = ns_ontology + "CumulativeTotal";
+	public static final String Instantaneous = ns_ontology + "Instantaneous";
```
kokfoongcmcl, line 82: This is again nitpicking, a cleaner way is to have something like
public static final String AVERAGE_STRING = ns_ontology + "AverageTimeSeries";
private static final Iri AVERAGE = iri(AVERAGE_STRING);

so in your deleteTimeSeries method you should not do something like
Average + "TimeSeries" to match the type

kokfoongcmcl: instead match with AVERAGE_STRING

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -300,7 +380,38 @@ public void deleteTimeSeries(String tsIRI, Connection conn) {
...
+		}
+		else if(type.equals(TimeSeriesSparql.StepwiseCumulative+"TimeSeries")){
+			timeSeriesType = TimeSeriesSparql.StepwiseCumulativeTimeSeries;
+		}
+		else if(type.equals(TimeSeriesSparql.CumulativeTotal+"TimeSeries")){
+			timeSeriesType = TimeSeriesSparql.CumulativeTotalTimeSeries;
+		}
+		else if(type.equals(TimeSeriesSparql.ns_ontology+"TimeSeries")){
+			timeSeriesType = TimeSeriesSparql.TimeSeries;
+		}
```
kokfoongcmcl, line 401: It's highly unlikely for this to be an issue, but there is potential for timeSeriesType to be null after this

kokfoongcmcl: Maybe the time series triples were instantiated with an older version of the base lib and if the namespace is updated, this will fail

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -41,27 +45,104 @@
...
+
+	//Values Pattern for TimeSeries types
+	private static final List<RdfObject> types = Arrays.asList(TimeSeries, CumulativeTotalTimeSeries, StepwiseCumulativeTimeSeries, AverageTimeSeries, InstantaneousTimeSeries);
+	private static final Variable tsType = Queries.SELECT().var();
+	private static final ValuesPattern vp = new ValuesPattern(tsType, types);
+
+	private final ArrayList<String> temporalUnitType = new ArrayList<>(Arrays.asList("unitSecond", "unitMinute", "unitHour", "unitDay", "unitWeek", "unitMonth", "unitYear"));
+
+	//EnumMap of allowed temporalUnit types
+	public static EnumMap<ChronoUnit, String> temporalUnitMap = new EnumMap<>(ChronoUnit.class);
```
kokfoongcmcl, line 92: should be private

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -41,27 +45,104 @@
...
+		temporalUnitMap.put(ChronoUnit.YEARS, ns_time+"unitYear");
+	}
+
+	/**
+	 * Logger for error output.
+	 */
+	private static final Logger LOGGER = LogManager.getLogger(TimeSeriesSparql.class);
+
+	//Custom Class to store numeric duration and the corresponding temporal unit pair to be able to map the pair to the averaging period iri
+	public class CustomDuration {
```
kokfoongcmcl, line 109: could be protected

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -110,18 +191,154 @@ protected void loadSparqlConfigs(String filepath) throws IOException {
...
+
+		return durationIRI;
+	}
+
+	/**
+	 * Creates a hashMap containing the (numericDuration, temporalUnit) mapping to averaging period IRI
+	 * for all the averaging period Iris in the kb
+	 * @return hashMap containing (numericDuration, temporalUnit) mapping to averaging period IRI
+	 */
+	public HashMap createDurationIRIMapping(){
```
kokfoongcmcl, line 257: The recommended return type is Map<CustomDuration, String>, HashMap is more like an implementation class

kokfoongcmcl: https://rules.sonarsource.com/java/RSPEC-1319

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -143,30 +360,78 @@ public boolean checkTimeUnitExists(String tsIRI) {
...
+		if(type.equals(AverageTimeSeries)){
+
+			if(duration.getNano()!=0){
+				LOGGER.warn("Nano is ignored");
+			}
+
+			//numeric Duration
+			if(duration.getSeconds() <=0.0){
+				throw new JPSRuntimeException(exceptionPrefix + "Numeric Duration must be a positive value");
+			}
```
kokfoongcmcl, line 407: I think these checks are impossible to be triggered? Surely the Duration class has its own checks

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -480,4 +831,69 @@ private List<String> getInstances(SelectQuery instanceSelectQuery, String placeh
...
+			modify.insert(iri(durationIRI).isA(Duration));
+			modify.insert(iri(durationIRI).has(unitType, iri(unit)));
+			modify.insert(iri(durationIRI).has(numericDuration, duration));
+
+		}
+
+		// Check that the data IRIs are not attached to a different time series IRI already
+		for (String iri: dataIRI) {
+			String ts = getTimeSeries(iri);
+			if(!(ts == null)) {
```
kokfoongcmcl, line 877: (ts != null) is cleaner, also applicable in a few other places

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -480,4 +831,69 @@ private List<String> getInstances(SelectQuery instanceSelectQuery, String placeh
...
+		// define type
+		modify.insert(tsIRI.isA(type));
+
+		if(type.equals(AverageTimeSeries)){
+
+			//numeric Duration
+			if(duration <=0.0){
+				throw new JPSRuntimeException(exceptionPrefix + "Numeric Duration must be a positive value");
+			}
+
```
kokfoongcmcl, line 860: see previous comment

kokfoongcmcl:
also there are some commented codes, if they are no longer needed remove them to make the code tidier

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
mehalaga: should we use camel case instead of AVERAGE_STRING?

There is a discrepancy:
for time series IRI its ns_ontology + "AverageTimeseries"+UUID.randomUUID();
whereas for type its AverageTimeSeries

hence one with came case and the other is not.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
mehalaga: the work around could be 

else { timeSeriesType = TimeSeriesSparql.TimeSeries; }

this would ensure if none of the types match, then it has to be the general time series type

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
mehalaga: Duration allows objects like Duration.ofDays(-12) which will trigger this

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
mehalaga: this method deals with double duration and not the duration object

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
kokfoongcmcl: Not necessarily, if the namespace is different from the current version none will match, I suggest throwing an error

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
kokfoongcmcl: there is a method Duration.isNegative(), that will be preferable than this. I still feel that it's up to the users to store whatever they want and it shouldn't be TimeSeriesClient's responsibility to check this

mehalaga: Yi Kai wanted this check to be there as he is going to be using this at the moment, that's why I added it there

mehalaga: Duration.isNegative() won't cover 0 which is invalid, either duration.isNegative() and Duration.isZero() should be used together or the existing check can be left as it is.

kokfoongcmcl: Then please use the isNegative method, it will be tidier

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
kokfoongcmcl: Since you already make a check before uploading this I don't think this is necessary

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -262,9 +263,11 @@ else if(type.get(i).equals(Type.GENERAL)){
    	    			throw new JPSRuntimeException(exceptionPrefix + "Inconsistent state created when initialising time series " + tsIRIs.get(i) +
    							" , as database related instantiation failed but KG triples were created.");
    	    		}
-   	    		throw new JPSRuntimeException(exceptionPrefix + "Timeseries was not created!", eRdbCreate);
+				failedTSIris.add(tsIRIs.get(i));
    	    	}
-   		}	
+   		}
+		if (!failedTSIris.isEmpty())
+			throw new JPSRuntimeException(exceptionPrefix + "Timeseries was not created!");
```
kokfoongcmcl, line 270: The failed dataIRIs should be included in the error message, as time series IRIs are hidden from users

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -41,27 +42,113 @@
...
+	public static final Iri AverageTimeSeries = iri(AverageTypeString);
+	public static final Iri InstantaneousTimeSeries = iri(InstantaneousTypeString);
+	private static final Iri Duration = prefix_time.iri("Duration");
+
+	//Timeseries IRI prefix
+	public static final String AverageIRIString = ns_ontology + "AverageTimeseries";
+	public static final String StepwiseCumulativeIRIString = ns_ontology + "StepwiseCumulativeTimeseries";
+	public static final String CumulativeTotalIRIString = ns_ontology + "CumulativeTotalTimeseries";
+	public static final String InstantaneousIRIString = ns_ontology + "InstantaneousTimeseries";
+	public static final String TimeSeriesIRIString = ns_ontology+"Timeseries";
```
kokfoongcmcl, line 87: either use ns_kb here or combine ns_ontology and ns_kb into a single constant since they are the same

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -480,4 +841,68 @@ private List<String> getInstances(SelectQuery instanceSelectQuery, String placeh
 
 		return instanceIRIs;
 	}
+
+	protected void initTS(String timeSeriesIRI, List<String> dataIRI, String dbURL, String timeUnit, Iri type, String durationIRI, Double duration, String unit){
```
kokfoongcmcl, line 845: is this still needed? you are not matching the unit by string anymore?

mehalaga: this is called from deleteTimeSeries in TimeSeriesClient

kokfoongcmcl: Do you need to check the unit? aren't you just adding what was in the KG previously? if it's needed shouldn't it be matched with the full IRI?

mehalaga: > Do you need to check the unit? aren't you just adding what was in the KG previously? if it's needed shouldn't it be matched with the full IRI?

at the moment its being called from deleteTimeSeries but the design should not be based on that. In future if someone wants to use it directly, removing the checks eliminate that usage. 
Okay, I'll match with the full IRI.

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
mehalaga: resolved

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
mehalaga: resolved

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesClient.java
```diff
@@ -112,23 +112,23 @@ public void initTimeSeries(List<String> dataIRIs, List<Class<?>> dataClass, Stri
...
 		else if(type.equals(Type.INSTANTANEOUS)){
 			timeseriesType = TimeSeriesSparql.InstantaneousTimeSeries;
-			tsIRI = TimeSeriesSparql.Instantaneous + "Timeseries_" + UUID.randomUUID();
+			tsIRI = TimeSeriesSparql.InstantaneousIRIString + "_" + UUID.randomUUID();
 		}
 		else if(type.equals(Type.GENERAL)){
 			timeseriesType = TimeSeriesSparql.TimeSeries;
-			tsIRI = TimeSeriesSparql.ns_kb + "Timeseries_" + UUID.randomUUID();
+			tsIRI = TimeSeriesSparql.TimeSeriesIRIString + "_" + UUID.randomUUID();
 		}
```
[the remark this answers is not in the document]
mehalaga: resolved

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
mehalaga: resolved

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -41,27 +42,107 @@
...
+	public static final Iri STEPWISE_CUMULATIVE_TIMESERIES = iri(STEPWISE_CUMULATIVE_TYPE_STRING);
+	public static final Iri CUMULATIVE_TOTAL_TIMESERIES = iri(CUMULATIVE_TOTAL_TYPE_STRING);
+	public static final Iri AVERAGE_TIMESERIES = iri(AVERAGE_TYPE_STRING);
+	public static final Iri INSTANTANEOUS_TIMESERIES = iri(INSTANTANEOUS_TYPE_STRING);
+	private static final Iri Duration = PREFIX_TIME.iri("Duration");
+
+	//Values Pattern for TimeSeries types
+	private static final List<RdfObject> types = Arrays.asList(TIMESERIES, CUMULATIVE_TOTAL_TIMESERIES, STEPWISE_CUMULATIVE_TIMESERIES, AVERAGE_TIMESERIES, INSTANTANEOUS_TIMESERIES);
+	private static final Variable tsType = Queries.SELECT().var();
+	private static final ValuesPattern vp = new ValuesPattern(tsType, types);
```
kokfoongcmcl, line 84: This is slightly risky because of how the query builder works, it starts from ?x0, ?x1, and so on for each SelectQuery object. If some method has another SelectQuery object and uses this, it will have 2 ?x0. Better put this within a method

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -110,341 +191,618 @@ protected void loadSparqlConfigs(String filepath) throws IOException {
...
+		// link each data to time series
+		for (String data : dataIRI) {
+			TriplePattern ts_tp = iri(data).has(hasTimeSeries, tsIRI);
+			modify.insert(ts_tp);
+		}
+
+		// optional: define time unit
+		if (timeUnit != null) {
+			modify.insert(tsIRI.has(hasTimeUnit, literalOf(timeUnit)));
+			//modify.insert(tsIRI.has(hasTimeUnit, iri(timeUnit)));
```
kokfoongcmcl, line 455: remove this?

JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/TimeSeriesSparql.java
```diff
@@ -110,341 +191,618 @@ protected void loadSparqlConfigs(String filepath) throws IOException {
...
+			// link each data to time series
+			for (String data : dataIRIs.get(i)) {
+				TriplePattern tsTp = iri(data).has(hasTimeSeries, tsIRI);
+				modify.insert(tsTp);
+			}
+
+			// optional: define time unit
+			if (timeUnit.get(i) != null) {
+				modify.insert(tsIRI.has(hasTimeUnit, literalOf(timeUnit.get(i))));
+				//modify.insert(tsIRI.has(hasTimeUnit, iri(timeUnit)));
```
kokfoongcmcl, line 539: same here

kokfoongcmcl (approved):
Double check everything and you can merge once you are happy

PR #435, opened 2022-11-11
dev-dtvf-3.3.1

mdhillman:
This PR brings in a fix to the bug that prevented the determination of the stack URL for a selected feature.

This was due to the fact that Cesium has no central abstraction for individual features, each data format has its own internal data structure and the determination of which group a feature came from has to be done on a case-by-case basis.


sm453 (approved):
Tested, works.

PR #440, opened 2022-11-12
190 dev package agentlogging in py4jps

jb2197:
This PR packaged `agentlogging` from [TheWorldAvatar/Agents/utils/python-utils](https://github.com/cambridge-cares/TheWorldAvatar/tree/f290fb98ce746b591d8b8c93cca1e89a409c959e/Agents/utils/python-utils) in `py4jps` and released version [1.0.29](https://pypi.org/project/py4jps/1.0.29/) to both PyPI and TestPyPI. Package `pyderivationagent` is also updated and released as [1.4.0](https://pypi.org/project/pyderivationagent/1.4.0/).

### DIFF
--- a/Agents/utils/python-utils/README.md
+++ b/Agents/utils/python-utils/README.md
@@ -1,6 +1,6 @@
 # TheWorldAvatar - Python Utils
 
-This Python package contains a number of logging utilities that may be useful to any Python-based project within The World Avatar (TWA) ecosystem. At the time of writing, this project builds an isolated package named `agentlogging` that users can import in their own code. In the future, this package may be bundled with the Python wrapper for the JPS Base Library so that only one dependency is required.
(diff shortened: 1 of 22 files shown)


sm453:
Did you simply copy (and hence duplicate) the logging code into py4jps? I can't see any deletions. I'm guessing you didn't delete anything because that may break a lot of things. If that is the case, I would recommend putting massive deprecation notices ("This is deprecated. Do not use. Do not develop. Instead, use...") into the readme and the sources of the old version.

jb2197:
> Did you simply copy (and hence duplicate) the logging code into py4jps? I can't see any deletions. I'm guessing you didn't delete anything because that may break a lot of things. If that is the case, I would recommend putting massive deprecation notices ("This is deprecated. Do not use. Do not develop. Instead, use...") into the readme and the sources of the old version.

As discussed, the deprecation warning is added to the original folder in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/440/commits/76c19d9860bcf8d107e8d1cb5b20ba822bd4c04a
The wiki is also updated accordingly: https://github.com/cambridge-cares/TheWorldAvatar/wiki/Logging/dcab40e4a37fc0b3a8a9ee986f7d9ef312d5ab60

sm453 (approved):

PR #444, opened 2022-11-15
1440 dev extend ontouom and ontoems to include additional concentration concepts

Abyssus1010:


JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -640,7 +640,7 @@
 
 
     <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/AirPollutantConcentration">
-        <rdfs:subClassOf rdf:resource="http://theworldavatar.com/resource/ontouom/MassOfSubstanceConcentration"/>
+        <rdfs:subClassOf rdf:resource="http://theworldavatar.com/resource/ontouom/Concentration"/>
```
markushofmeister, line 643: In the latest OntoUOM extension, Yi Kai followed the new naming convention for new ontouom concepts, e.g. "https://www.theworldavatar.com/kg/ontouom/MassOfSubstanceConcentration" (https://github.com/cambridge-cares/TheWorldAvatar/blob/main/JPS_Ontology/ontology/ontouom/MassOfSubstanceConcentration)
However we do it, it will be inconsistent, but maybe we should continue with this new approach (at least for new concepts) even though that means that the IRIs will not resolve as URLs. @sm453 What is your perspective on that?

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -835,7 +835,28 @@
...
+
+    <!-- https://www.theworldavatar.com/kg/ontoems/CarbonMonoxideConcentration -->
+
+
+    <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/CarbonMonoxideConcentration">
+        <rdfs:subClassOf rdf:resource="https://www.theworldavatar.com/kg/ontoems/AirPollutantConcentration"/>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The concentration of carbon monoxide in the ambient air.</rdfs:comment>
+        <rdfs:isDefinedBy rdf:datatype="http://www.w3.org/2001/XMLSchema#string">https://www.theworldavatar.com/kg/ontoems/</rdfs:isDefinedBy>
+        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Carbon monoxide concentration</rdfs:label>
+    </owl:Class>
```
markushofmeister, line 860: I think you missed the definition of "MassOfSubstanceConcentration" as subClass of "Concentration".

image

markushofmeister: Furthermore, I believe we can simply use OM's "VolumeFraction" concepts for "VolumeOfSubstanceConcentration", see: https://raw.githubusercontent.com/cambridge-cares/OM/master/om-2.0.rdf

```
<owl:Class rdf:about="&om;VolumeFraction">
  <rdfs:label xml:lang="en">volume fraction</rdfs:label>
  <rdfs:label xml:lang="nl">volumefractie</rdfs:label>
  <om:alternativeLabel xml:lang="en">concentration (v/v)</om:alternativeLabel>
  <rdfs:subClassOf rdf:resource="&om;Quantity"/>
  <om:symbol>φ</om:symbol>
</owl:Class> 

<owl:Class rdf:about="&om;VolumeFraction">
  <om:commonlyHasUnit rdf:resource="&om;percent"/>
  <om:commonlyHasUnit rdf:resource="&om;partsPerMillion"/>
</owl:Class>
```

JPS_Ontology/ontology/ontouom/VolumeOfSubstanceConcentration
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0"?>
```
markushofmeister, line 1: Using OM's "VolumeFraction" concept shall make this obsolete

JPS_Ontology/ontology/ontouom/VolumeOfSubstanceConcentration-Dimension
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0"?>
```
markushofmeister, line 1: Using OM's "VolumeFraction" concept shall make this obsolete

JPS_Ontology/ontology/ontouom/VolumeOfSubstanceConcentrationUnit
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
```
markushofmeister, line 1: Using OM's "VolumeFraction" concept shall make this obsolete

markushofmeister (changes requested):

markushofmeister:
The subordination of Concentration and VolumeFraction / MassOfSubstanceConcentration is still wrong. Please pay attention to your base IRIs as some of them are using `http` while others contain `https` which makes the relationships break!
Please review all of them before re-requesting a review
image

JPS_Ontology/ontology/ontoems/OntoEMS.owl
sm453: Absolutely agree.

sm453:
Your most recent commit seems to contain only changes to the OWL file, but no corresponding changes in the underlying CSV. I hope you just forgot to also include the CSV file into the commit? The OWL should simply be generated from the CSV. Otherwise, if you edit the CSV and the OWL independently, we have a serious problem...

Abyssus1010:
> 
Yes. I have not included the csv file in the commits as I want to make sure that the changes to the ontology was correct. I am reviewing the base IRIs as Markus mentioned and I will commit both after I have checked through everything.

markushofmeister:
> Your most recent commit seems to contain only changes to the OWL file, but no corresponding changes in the underlying CSV. I hope you just forgot to also include the CSV file into the commit? The OWL should simply be generated from the CSV. Otherwise, if you edit the CSV and the OWL independently, we have a serious problem...

Fully agree; however two comments: Last time I used the TBox Generator (a few months back), it
1) Had issues with recognising our new IRI schema, i.e. ending with `/` instead of `#`. Hence, manual string replacement was required which could keep things naturally out of sync
2) Did not support xsd types likes `date`, `dateTime`, etc. which become more relevant for our use cases which required again manual string replacement

Not sure if they are solved now, but if not that should be a priority as well then. Just mentioning this here for the records

sm453:
> > Your most recent commit seems to contain only changes to the OWL file, but no corresponding changes in the underlying CSV. I hope you just forgot to also include the CSV file into the commit? The OWL should simply be generated from the CSV. Otherwise, if you edit the CSV and the OWL independently, we have a serious problem...
> 
> @sm453 Fully agree; however two comments: Last time I used the TBox Generator (a few months back), it
> 
> 1. Had issues with recognising our new IRI schema, i.e. ending with `/` instead of `#`. Hence, manual string replacement was required which could keep things naturally out of sync
> 2. Did not support xsd types likes `date`, `dateTime`, etc. which become more relevant for our use cases which required again manual string replacement
> 
> Not sure if they are solved now, but if not that should be a priority as well then. Just mentioning this here for the records

Thanks for pointing this out - it's of course not good. I've asked @msff2 to fix this as a matter of urgency.

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -626,6 +620,17 @@
...
+
+
+
+    <!-- http://www.ontology-of-units-of-measure.org/resource/om-2/VolumeFraction -->
+
+
+    <owl:Class rdf:about="http://www.ontology-of-units-of-measure.org/resource/om-2/VolumeFraction">
+        <rdfs:subClassOf rdf:resource="https://www.theworldavatar.com/kg/ontouom/Concentration"/>
+        <rdfs:isDefinedBy rdf:datatype="http://www.w3.org/2001/XMLSchema#string">http://www.ontology-of-units-of-measure.org/resource/om-2/</rdfs:isDefinedBy>
+        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">volume fraction</rdfs:label>
```
markushofmeister, line 632: All other labels are capitalised. Try to be consistent in the future

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -626,6 +620,17 @@
...
     <owl:Class rdf:about="http://www.ontology-of-units-of-measure.org/resource/om-2/VolumetricFlowRate"/>
+
+
+
+    <!-- http://www.ontology-of-units-of-measure.org/resource/om-2/VolumeFraction -->
+
+
+    <owl:Class rdf:about="http://www.ontology-of-units-of-measure.org/resource/om-2/VolumeFraction">
+        <rdfs:subClassOf rdf:resource="https://www.theworldavatar.com/kg/ontouom/Concentration"/>
+        <rdfs:isDefinedBy rdf:datatype="http://www.w3.org/2001/XMLSchema#string">http://www.ontology-of-units-of-measure.org/resource/om-2/</rdfs:isDefinedBy>
```
markushofmeister, line 631: Strictly speaking this should be `https://www.theworldavatar.com/kg/ontoems/`. The VolumeFraction itself is specified in OM, yes, but this particular relationship is defined in OntoEMS

markushofmeister (approved):

sm453 (approved):

PR #457, opened 2022-11-17
dev-8-historical-house45-utilities-agent

qhouyee:
An agent that instantiates the utility consumption stored in Excel as a time series and links it to the building.

This agent requires a time series client that is currently in development and on a feature branch. The required jps-base-lib dependency version will need to be updated once it is ready.


sm453 (approved):
Tested, works.

PR #458, opened 2022-11-17
dev-ontoubemmp-gas-consumption-extension: Extended ontology with a ne…

qhouyee:
…w GasConsumption class


YiKaiTsai (changes requested):
Update OntoBIM URI to http://www.theworldavatar.com/ontology/ontobim/ontoBIM#, thanks.

YiKaiTsai (approved):

PR #466, opened 2022-11-17
FeatureInfoAgent v2.0.0

mdhillman:
This PR brings in major changes to the FeatureInfoAgent. These changes include:

- Separation into multiple classes
- Addition of JUnit Tests
- Updated documentation
- Seperation of queries for metadata and timeseries
- Automatic discovery and federation of Blazegraph endpoints

For more details please see the associated README file.


mdhillman:
This PR is a draft and will not be published until the following conditions are met:

- The stack client library is merged to make the getServiceUrl() is available on a non-SNAPSHOT version.
- The agent has been tested within a real stack, with real data.
  - This has not been possible during development as the overhead of creating a mock stack with mock KG and RDBs is too high.
- A GitHub action have been added to run units tests upon PR submission.
- A GitHub action has been added to build and push the docker image after a PR is merged.

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/Utils.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Constructor.
+     */
+    private Utils() {
+        // No
+    }
+
+    /**
+     * Combine timeseries objects into a single timeseries objects. For some reason this is
```
kokfoongcmcl, line 36: This is specific to the river sensor stations due to the way I initialised them, one station may contain multiple tables, typically one station will contain a table with multiple columns if it measures more than 1 thing. It's a combination of laziness and how the API returns data

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/FeatureInfoAgent.java
```diff
@@ -0,0 +1,417 @@
...
+ * 
+ * Note: this version of the agent is a VERY rough implementation and should not be used outside of very specific
+ * use cases. It currently relies on hardcoded SPARQL queries, and will not work out-of-the box for any old data sets.
+ * Future improvements will make this more generic, at which point the agent could be used more widely.
+ *
+ * @author Michael Hillman {@literal <mdhillman@cmclinnovations.com>}
+ */
+@Controller
+@WebServlet(urlPatterns = {"/get", "/status"})
+public class FeatureInfoAgent extends JPSAgent {
```
kokfoongcmcl, line 52: Seb will disagree on this, but there's no point in using JPSAgent if you're not using processRequestParameters

Agents/FeatureInfoAgent/code/WEB-INF/web.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kokfoongcmcl, line 1: I believe this file is redundant when you use the @WebServlet annotation in your FeatureInfoAgent class

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/FeatureInfoAgent.java
kokfoongcmcl: Ok I see now that you are using readJsonParameter, then this is not applicable, documentation somewhere needs to say this takes in a "query" parameter with a JSON string in it

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/Utils.java
mdhillman: If this code is run on non-river sensor stations what will happen? 

This agent is only going to be able to handle timeseries if timeseries are storred generically for all station/location types. I don't really want to be adding an ever growing list of "if(thisTypeOfStation){}" clauses.

Agents/FeatureInfoAgent/code/WEB-INF/web.xml
mdhillman: You may be right, I'll make a note to try removing it in future (it's not critical to do so now).

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/Utils.java
kokfoongcmcl: Based on the way you create the time series objects it won't make any difference, but it's a bit inefficient

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/kg/TimeHandler.java
```diff
@@ -0,0 +1,386 @@
...
+        String fixedIRI = measureIRI;
+        if (fixedIRI.startsWith("<") && fixedIRI.endsWith(">")) {
+            fixedIRI = fixedIRI.substring(1, fixedIRI.length() - 1);
+        }
+
+        // Build then return the object
+        TimeSeries<Instant> result = null;
+        if(this.hours < 0) {
+            // Get all data
+            result = this.tsClient.getTimeSeries(new ArrayList<>(Arrays.asList(fixedIRI)));
```
kokfoongcmcl, line 347: In the latest base lib, if you don't pass a connection object the these methods, it will open and close a connection for each query you make, if you want to make it faster, pass a connection object and close it at the end using try-with-resource

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/Utils.java
mdhillman: I'd suggest we leave it as-is then. For this PR I'm more interested in getting the functionality correct than it being more efficient.

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/kg/TimeHandler.java
mdhillman: This is something I can add in. Is there a method in the base lib to create the initial connection, or do I just use generic JDBC code?

kokfoongcmcl: There's a RemoteRDBStoreClient.getConnection() method, which is basically calling the jdbc code

mdhillman:
> This PR is a draft and will not be published until the following conditions are met:
> 
> * The stack client library is merged to make the getServiceUrl() is available on a non-SNAPSHOT version.
> * The agent has been tested within a real stack, with real data.
>   
>   * This has not been possible during development as the overhead of creating a mock stack with mock KG and RDBs is too high.
> * A GitHub action has been added to run units tests upon PR submission.
> * A GitHub action has been added to build and push the docker image after a PR is merged.

The two mentioned GitHub actions have now been added.

Agents/FeatureInfoAgent/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+#==================================================================================================
+FROM maven:3.8.5-openjdk-17-slim as builder
+
+# Copy all files into root's home, including the source, pom file, ./m2 directory and credentials
+ADD . /root
+
+# Populate settings templates with credentials, repo name
+WORKDIR /root/.m2
+
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 11: since encrypted passwords can contain the LATTER, but not the FORMER

Agents/FeatureInfoAgent/code/pom.xml
```diff
@@ -0,0 +1,178 @@
...
+    <build>
+        <plugins>
+
+            <!-- Compile and build with Java 11 -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
+                <configuration>
+                    <release>17</release>
```
sm453, line 60: The comment says "with Java 11"...

Agents/FeatureInfoAgent/README.md
```diff
@@ -0,0 +1,94 @@
...
+    <img src="time-query-example.jpg" alt="Example result of a timeseries query" width="75%"/>
+</p>
+
+### Requests
+
+All incoming requests should use the `/get` route, containing a `query` parameter that has a JSON body (compatible with the agent framework in the JPS Base Lib), which in turn contains a single `iri` parameter. In this version of the agent, **no** other parameters (e.g. `endpoint`, `namespace`) are required.
+
+## Deployment
+
+To deploy the agent, follow the generic steps for deploying a stack-enabled agent. You _may_ need to ensure that your stack contains a valid route so that outside requests can read the agent's container.
```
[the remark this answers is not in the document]
markushofmeister: But I guess building would not _always_ be required and people should rather just pull the image, right?

markushofmeister (changes requested):
I am unable to build (and push) the Docker image using `bash ./stack.sh build`. I am getting a `denied: permission_denied: write_package` error despite being able to push other Docker images to ghcr.io (from the same machine). The latest version of the published image is 4 days old. Is this still the latest and worth testing the functionality with it?

Agents/FeatureInfoAgent/README.md
```diff
@@ -0,0 +1,94 @@
...
+   4. Dynamically discover all endpoints provided by Blazegraph.
+2. Check the incoming route.
+   1. Throw error if an unknown route.
+3. If `/status` route:
+   1. Return current status of the agent.
+4. If `/get` route:
+   1. Check validity of incoming request.
+      1. If malformed, send a BAD_REQUEST response.
+      2. Determine the class representing the IRI.
+         1. If query successful but no class returned, send a NO_CONTENT response.
```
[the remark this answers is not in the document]
gpeb2: I think Michael means the official HTTP response code 204 "NO_CONTENT" ([List_of_HTTP_status_codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)). I'm not suggesting there isn't a better code though.

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
gpeb2: Yes, building should _only_ be done if changes have been made to that code.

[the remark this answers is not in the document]
sm453: Correct.

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
mdhillman: Yes, I'd like to stick to approved HTTP codes if possible. I think NO_CONTENT conveys that the connection to the KG _was_ made okay, but it just didn't contain details on the class; but I'm open to suggestions.

Agents/FeatureInfoAgent/README.md
```diff
@@ -0,0 +1,94 @@
...
+      3. Look up corresponding meta query for that class.
+         1. If none found, send a NO_CONTENT response.
+      4. Run query to get metadata.
+      5. Parse resulting metadata into approved DTVF format.
```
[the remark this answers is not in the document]
mdhillman: The last think the FIA does is takes that `property`, `value`, `unit` structure that is expected from queries and parses it to something a little more human-readable. For instance...

```json
[{
  "Property": "Length",
  "Value": "100",
  "Unit": "m" 
}]
```

...would get parsed into the following before being passed back to the visualisation.

```json
[{
  "Length": "100 [m]"
]}
```

Agents/FeatureInfoAgent/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Restrictions
+
+At the time of writing, the FIA has a few restrictions that all users should be aware of. These are as follows:
+
+- The FIA can only be run within a [stack](https://github.com/cambridge-cares/TheWorldAvatar/tree/main/Deploy/stacks/dynamic/stack-manager).
+- The FIA can only report on metadata and timeseries that are contained within the same stack.
+- The FIA can only return timeseries data on series that use the Instant class.
+- The FIA can only return timeseries data from a single PostGreSQL database.
+- The FIA cannot handle large, intensive queries (i.e. anything that takes more than a minute to return).
```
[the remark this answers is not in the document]
mdhillman: First three points are across everything the FIA does right now, whether the queries are federated or not.

The last point is more of a user facing requirement; in theory the FIA can handle as large a query as the base lib can, but we don't want the user to be waiting within the visualisation for minutes.

Agents/FeatureInfoAgent/README.md
```diff
@@ -0,0 +1,94 @@
...
+         1. If none found, send a NO_CONTENT response.
+      4. Run query to get metadata.
+      5. Parse resulting metadata into approved DTVF format.
+      6. Look up corresponding timeseries query for that class.
+         1. If none found, log warning but continue.
+         2. If found, run query.
+            1. If no data found, quietly continue.
+      7. Build and return final JSON object.
+
+It's also worth noting that in the current version of the FIA, any queries to the knowledge graphs are sent to all discovered namespace endpoints via federation. Whilst this will marginally increase processing times, these queries should be pretty quick, and shouldn't be triggered too often, so the risk of delay is hopefully less than the benefit of not having to specify the endpoint beforehand.
```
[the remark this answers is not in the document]
mdhillman: That does sound like a reasonable idea. Would a query to just see if an IRI exists within a 8-10 Mil triplestore be lengthy?

Agents/FeatureInfoAgent/README.md
```diff
@@ -0,0 +1,94 @@
...
+- `[ONTOP]`: The URL of the Ontop service within the stack will be injected
+
+Queries for metadata should not concern themselves with data relating to timeseries (that can be handled within the timeseries query). Queries here need to return a table with two (or optionally three) columns. The first column should be named `Property` and contains the name of the parameter we're reporting, the second should be `Value` and contain the value. The optional third column is `Unit`. Any other colums will be ignored.
+
+
+<p align="center">
+    <img src="meta-query-example.jpg" alt="Example result of a metadata query" width="50%"/>
+</p>
+
+Queries for timeseries data need to return the measurement/forecast IRIs (that will be used to grab the actual values from PostGreSQL), as well as parameters associated with each measurement/forecast. Required columns are `Measurement` (or `Forecast`) containing the IRI, `Name` containing a user facing name for this entry, and `Unit` containing the unit (which can be blank). In this case, any other columns reported by the query **will** be picked up and passed back to the visualisation as regular key-value properties.
```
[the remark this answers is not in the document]
mdhillman: Yes, I believe so.

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
mdhillman: I have updated the README, but if anyone has any details on how to update the stack routing then it may be worth adding/linking to that.

mdhillman:
Connection object has been added for the timeseries RDB queries, will need deployment within a stack to test if this improves the efficiency.

mdhillman:
> I am unable to build (and push) the Docker image using `bash ./stack.sh build`. I am getting a `denied: permission_denied: write_package` error despite being able to push other Docker images to ghcr.io (from the same machine). The latest version of the published image is 4 days old. Is this still the latest and worth testing the functionality with it?
> 
> Is this due to the fact that publishing the image is now done by github actions? Just confused as the latest image is much older than your tag `ready for review`.

How the image is push shouldn't affect who else can push it (as far as I'm aware). There is a kink with uploading images to GitHub though; the first time it's done they appear to default to private. I believe I've changed the settings to make the FIA image public, but I will double check.

mdhillman:
> > I am unable to build (and push) the Docker image using `bash ./stack.sh build`. I am getting a `denied: permission_denied: write_package` error despite being able to push other Docker images to ghcr.io (from the same machine). The latest version of the published image is 4 days old. Is this still the latest and worth testing the functionality with it?
> > Is this due to the fact that publishing the image is now done by github actions? Just confused as the latest image is much older than your tag `ready for review`.
> 
> How the image is push shouldn't affect who else can push it (as far as I'm aware). There is a kink with uploading images to GitHub though; the first time it's done they appear to default to private. I believe I've changed the settings to make the FIA image public, but I will double check.

I had made it public, but only for download. It looks like there's a second stage (that I missed) that allows permissions inheritance from the source repository. I've enabled that now so everyone should have the same access that they have to the TWA code.

For now, you may need to build the image locally. The current actions will only push it _after_ the PR is closed. As this doesn't help you test it, what I should really do it have the PR push a SNAPSHOT version then a non-SNAPSHOT once it's approved. I'll need to discuss with @George on how to achieve this.

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
markushofmeister: Sounds sensible. It was not clear to me that your refer to official HTTP response codes here

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
markushofmeister: Makes perfect sense; "approved" could be elaborated here for clarify, e.g. Parse resulting metadata from SPARQL query into visualisation ready format

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
markushofmeister: It should not. Queries for specific IRIs should be very quick generally; however, depending on the metadata query and in case an IRI exists in multiple namespaces, this could lead to lengthy queries in case the query is federated to the unnecessary endpoint first.

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/kg/TimeHandler.java
```diff
@@ -0,0 +1,396 @@
...
+     * @param endpoints Blazegraph endpoints for the KG.
+     */
+    public TimeHandler(String iri, String classMatch, Connection rdbConnection, List<ConfigEndpoint> endpoints) {
+        String fixedIRI = iri;
+        if(!fixedIRI.startsWith("<")) fixedIRI = "<" + fixedIRI;
+        if(!fixedIRI.endsWith(">")) fixedIRI = fixedIRI + ">";
+
+        this.iri = fixedIRI;
+        this.classMatch = classMatch;
+        this.rdbConnection = rdbConnection;
```
kokfoongcmcl, line 112: It looks like you are storing the connection object, it's not recommended, you should do something like
`try (Connection conn = rdbClient.getConnection()) {
    timeSerieClient.getTimeSeries(....., conn)
}`

Agents/FeatureInfoAgent/README.md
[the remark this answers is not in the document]
mdhillman: Unless we find the current implementation too slow, then I suggest we leave this to be looked into as a new feature within a subsequent version of the agent.

sm453:
For time series queries, would it be possible to make each of the `queries` have its own `hours` parameter (in the config json)? A use case may contain many time series of substantially different scale (e.g. hours vs. years).

mdhillman:
> For time series queries, would it be possible to make each of the `queries` have its own `hours` parameter (in the config json)? A use case may contain many time series of substantially different scale (e.g. hours vs. years).

Yes this should be possible (although not trivial), I will take a look.

mdhillman:
> > For time series queries, would it be possible to make each of the `queries` have its own `hours` parameter (in the config json)? A use case may contain many time series of substantially different scale (e.g. hours vs. years).
> 
> Yes this should be possible (although not trivial), I will take a look.

This change has now been implemented, a `timeLimit` parameter can be set for each class to set an limit for timeseries data (in hours, defaults to 24 if not present).

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/kg/TimeHandler.java
```diff
@@ -358,18 +362,25 @@ protected TimeSeries<Instant> buildTimeseriesObject(String measureIRI, Connectio
...
             // Get all data
+            LOGGER.debug("Getting timeseries without limit...");
             result = this.tsClient.getTimeSeries(new ArrayList<>(Arrays.asList(fixedIRI)), rdbConnection);
+
         } else {
+            LOGGER.debug("Getting timeseries with limit of last {} hours...", this.hours);
+
             // Determine bounds
             Instant lowerBound = LocalDateTime.now().minusHours(this.hours).toInstant(ZoneOffset.UTC);
             Instant upperBound = LocalDateTime.now().toInstant(ZoneOffset.UTC);
```
kokfoongcmcl, line 373: Are we sure that this is what is required? Should it be something like taking the latest entry in the time series table and subtracting that instead?

sm453: I'm inclined to agree with @kokfoongcmcl here. It is likely more useful to see the so many _most recent_ hours, rather than so many hours back from _now_ (which is likely to be nothing at all very quickly!).

mdhillman: The current design we've used in all previous visualisations is to show data from the last 24 hours.

We can change this to range from N hours since the latest data point, but this is something I would require to be done within the TimeseriesClient (I suggest as a new method). Unless some derivation calculation is being made, I don't think agents should handle filtering the timeseries at all, this should be done centrally.

kokfoongcmcl: The easiest thing to do now is to use getLatestData, get the value of the time from the TimeSeries object, then calculate the range and use getTimeSeriesWithinBounds.

It might be tricky to implement this in the TimeSeriesClient itself, because I allowed the Time class to be anything (e.g. Integer, Double, etc.), if it's Double, and you provide a Duration object, I don't know how to deal with it

mdhillman: Would this trigger 2 separate connections to the RDB?

I strongly think the TimeseriesClient should be responsible for any filtering, as I don't want to tie any agents directly to the format in which timeseries are returned. Additionally, this sounds like something we'll want to use in different agents, so should be done centrally.

kokfoongcmcl: You can use the same connection object for both calls.

It's not straightforward to support that, it will need to have one method for each class, I can't just do "-" unless they're subclass of Number

mdhillman: I suggest this request goes on the back burner then until such time that it can be added centrally.

kokfoongcmcl: As Seb said, we don't think this is a good idea anyway because you'll get empty results quite quickly, depending on the data of course

sm453:
> For time series queries, would it be possible to make each of the `queries` have its own `database_name` parameter (in the config json)? In a complex use case it may not be practical to have all time series in the same database, hence a single, global setting is a bit restrictive.

Or even better, read the hasRDB property from the KG, but whatever is quicker to implement wins for now...

kokfoongcmcl:
> > For time series queries, would it be possible to make each of the `queries` have its own `database_name` parameter (in the config json)? In a complex use case it may not be practical to have all time series in the same database, hence a single, global setting is a bit restrictive.
> 
> Or even better, read the hasRDB property from the KG, but whatever is quicker to implement wins for now...

just have to be careful, the URL is probably the one accessed by the agent instantiating the time series, not the external one

markushofmeister:
> I'm not sure if _all_ data has the `hasRDB` relationship yet, pretty sure the data on our development servers (which I use for testing) doesn't. Until that happens, I'd be wary of relying on it.

I might be mistaken, but I do not see why earlier timeseries should not have the `hasRDB` relationship. I just checked again, and `hasRDB` has been part of the initial commit of the TimeSeries Ontology

sm453:
I just checked the triples generated by a few current instantiation agents, they all put the stack-internal JDBC URLs, which is exactly what we need. However, if it is easier/quicker/less problematic for now to allow the config to specify a DB name per class, then that is the way to go. Better solutions can be left for later.

markushofmeister (changes requested):
After adding `<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.29.4/moment.min.js" ></script>` to the `index.html` I am able to visualise retrieved meta and timeseries data. However, retrieving time series data still causes issues in cases where units have encoding problems.
As "wrong" instantiation of non-ascii characters within Blazegraph seems to be a known issue (see issue #513), we should try to have a robust solution for that.

To reproduce the issue, please use the MetOffice stations `Marham` (airfield south of King's Lynn, `<https://www.theworldavatar.com/kg/ontoems/ReportingStation_897428dc-6340-4c56-b632-50bd7f128f61>`) at the endpoint `http://165.232.172.16:3838/blazegraph/ui/#query` (namespace `stations`). I am using the timeseries.sparql file as below.
Without commenting out both `  ?Measurement ^om:hasValue/rdf:type <https://www.theworldavatar.com/kg/ontoems/RelativeHumidity> . `, I can retrieve the one time series successfully. However, when trying to retrieve all time series by commenting those two lines out, the time series panel remains completely empty, although Blazegraph returns "correct" results:
image

For other stations, time series with wrongly encoded units are shown as below:
image


```
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
prefix om:  <http://www.ontology-of-units-of-measure.org/resource/om-2/>
prefix ems: <https://www.theworldavatar.com/kg/ontoems/>

SELECT ?Measurement ?Name ?Unit
WHERE {
  {
  # Observations/Measurements
  [IRI] ems:reports/om:hasValue ?Measurement . 
  ?Measurement ^om:hasValue/rdf:type <https://www.theworldavatar.com/kg/ontoems/RelativeHumidity> . 
  ?Measurement ^om:hasValue/rdf:type ?type ; 
               om:hasUnit/om:symbol ?Unit .                
  BIND (concat(replace(str(?type), str(ems:), ''), ' observation') as ?Name)               
  # Forecasts
  } UNION {
  [IRI] ems:reports/ems:hasForecastedValue ?Measurement . 
  ?Measurement ^om:hasValue/rdf:type <https://www.theworldavatar.com/kg/ontoems/RelativeHumidity> . 
  ?Measurement ^ems:hasForecastedValue/rdf:type ?type ; 
               om:hasUnit/om:symbol ?Unit .                
  BIND (concat(replace(str(?type), str(ems:), ''), ' forecast') as ?Name)    
  }
}
```

mdhillman:
> https://www.theworldavatar.com/kg/ontoems/ReportingStation_897428dc-6340-4c56-b632-50bd7f128f61

I've tried reproducing this with the given stack URL and IRI, unfortunatly I cannot get the FIA to return any timeseries data (the error "Could not produce JSON objects from TimeSeries instances!" is given).

Mocking the repsonse from the agent with a JSON string previously sent to me by @markushofmeister does not reproduce the issue either. All timeseries (even those with degrees in the name) appear to work fine.

markushofmeister:
> I've tried reproducing this with the given stack URL and IRI, unfortunatly I cannot get the FIA to return any timeseries data (the error "Could not produce JSON objects from TimeSeries instances!" is given).

I believe this (unfortunately) needs further investigation before closing this PR. I just had another try and can "confirm" that

1) There are issues with retrieving particular time series (currently, this seems to affect primarily time series with non-ascii units which might be "wrongly" encoded), e.g. for `Tottenhill` some time series get displayed correctly, while others are not at all ...

Properly shown timeseries            | "Undefined" timeseries
:-------------------------:|:-------------------------:
image |  image

... despite the time series data seems to be available for all time series in PostGIS



2) Some stations are not "detectable" by FIA at all, e.g. for `Marham` the FIA returns the fallback values from PostGIS despite both `sparql` queries seem to work when executed in Blazegraph directly
image

Metadata query            |  Timeseries query
:-------------------------:|:-------------------------:
image  |  image

sm453:
Another request: It would be great if the metadata query in the config json was optional, too (with the front-end displaying just the PostGIS metadata if no such query is provided). This would allow attaching time series to datasets that have metadata from PostGIS, without having to write an OBDA mapping and a metadata query that pulls out the metadata that would be displayed otherwise anyway.

mdhillman:
I have updated the FeatureInfoAgent, and the DTVF (deployed as v3.3.4) on this branch to address the below issues. The earlier these can be tested and confirmed as fixed (or re-addressed), the sooner this branch can be merged (I would like this to be done ASAP so that upcoming CReDo work doesn't have to be based on a feature branch).

- Metadata queries should now be optional, simply omit the "metaFile" entry for a class in the configuration file.
- Timeseries should no longer combine time values, this was due to older code that was created when some assumptions still held true.
- Issue #513 should now be resolved, fixing issues with the units in timeseries causing issues.

Agents/FeatureInfoAgent/code/src/main/java/com/cmclinnovations/featureinfo/kg/TimeHandler.java
```diff
@@ -0,0 +1,428 @@
...
+            // Filter the results
+            jsonResult = this.filterJSON(jsonResult);
+
+            // Transform into expected JSON objects
+            return this.populateTimeseries(jsonResult);
+        } catch(Exception exception) {
+            LOGGER.warn("Query to get measurement details has caused an exception!");
+
+            response.setStatus(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode());
+            response.getWriter().write("{\"description\":\"intCould not determine a valid Blazegraph endpo.\"}");
```
sm453, line 196: Please fix the message string. A funny drag-and-drop accident, I guess...

web/digital-twin-vis-framework/example-cesium-vis/webspace/index.html
```diff
@@ -136,8 +137,8 @@
 
             // ===== CUSTOMISABLE =====
             // Enter Mapbox account name and API key here, even in Cesium visualisations!
-            MapHandler.MAP_USER = "";
-            MapHandler.MAP_API = "";
+            MapHandler.MAP_USER = "cmclinnovations";
+            MapHandler.MAP_API = "pk.eyJ1IjoiY21jbGlubm92YXRpb25zIiwiYSI6ImNrbGdqa3RoNDFnanIyem1nZXR3YzVhcmwifQ.hVk983r6YYlmFE8kSMbzhA";
```
sm453, line 141: Please remove.

web/digital-twin-vis-framework/example-cesium-vis/webspace/data.json
```diff
@@ -7,6 +7,12 @@
             "type": "tiles",
             "uri": "https://kg.cmclinnovations.com/cdn/buildings/nyc/tileset.json"
         },
+        {
+            "id": "house45",
+            "type": "tiles",
+            "uri": "https://kg.cmclinnovations.com/cdn/buildings/house45/tileset.json",
+            "position": [-74.01499545037598, 40.70216915119465, 0]
+        },
```
sm453, line 15: You may want to remove this.

web/digital-twin-vis-framework/example-cesium-vis/webspace/data.json
```diff
@@ -22,6 +28,11 @@
 			"name": "Buildings of New York", 
             "source": "3d-tiles-source"    
         },
+        {
+            "id": "house45-layer",
+			"name": "House 45", 
+            "source": "house45"    
+        },
```
sm453, line 35: You may want to remove this.

web/digital-twin-vis-framework/example-cesium-vis/webspace/settings.json
```diff
@@ -1,8 +1,12 @@
...
-        "center": [-74.03847455978394, 40.69933253480097, 2500],
+        "center": [-74.02208684116992, 40.69885291708034, 400],
         "heading": 50,
-        "pitch": -40,
+        "pitch": -30,
         "roll": 0.0
+    },
+    "fields": {
+        "name": "Name",
+        "description": "Description"
```
sm453, line 10: You may want to remove this.

markushofmeister (approved):
I tested the FeatureInfoAgent 2.0 with DTVF 3.3.4 (including UTF-8 charset) for the latest MetOffice agent data in the Mapbox visualisation and can confirm that
* Meta-data gets correctly retrieved and displayed
* Time series data gets correctly retrieved and displayed, even for time series with non-ascii characters in units

sm453 (approved):
Tested the following (with the Cesium version of the DTVF only):
- different databases for different queries
- metadata query only, time series query only, both queries, neither
- sizing of the pop-up window for named features
All working.

PR #471, opened 2022-11-23
Add sparql service filter

gpeb2:
Add a method that can replace service names with their IRI in the ```SERVICE``` patterns within a SPARQL query. Currently limited to ```<ontop>```.


kokfoongcmcl:
Might be worth mentioning somewhere that this will only work with "ontop" and nothing else

kokfoongcmcl (approved):
function wise it works

PR #473, opened 2022-11-24
Dev mods agent evaluating surrogates

cmclscourtney:


gpeb2 (changes requested):
Other than correcting the two method names this looks fine.

gpeb2 (approved):

PR #478, opened 2022-11-27
193-hotfix-derivation-agent-python-example

jb2197:
It just occurred to me that `"${AGENTDIR}:create_app()"` is not resolved correctly in the previous way of expressing `ENTRYPOINT`, breaking the example production image published on GitHub. The hotfix should address this issue.


markushofmeister (approved):
Looks good to me. 

The observed issue occurs as variable substitution is only enabled in the `shell form` of ENTRYPOINT and not in the `exec form` (see https://stackoverflow.com/a/37904830)

PR #479, opened 2022-11-28
46 dev extend ontouom to include electrical power classes

Abyssus1010:


Abyssus1010:
image

Abyssus1010:
Hi @sm453 , can you help me to review this extension for ontouom when you have the time? Thank you.

Abyssus1010:
@sm453 In addition, can I check with you whether it is a good idea to include prefixed units such as attowatt, centiwatt, gigawatt etc in ontouom:ElectricalPowerUnit as well?

sm453:
I believe this is not the right approach at all. Power, from a physical point of view, is just that, with SI unit Watt. Whether it is electrical or otherwise, or AC or DC, is not a question of unit/dimension (otherwise one would have to define large numbers of units!). You are of course free (and encouraged) to mark up your power quantities as electrical/AC/DC as you like, but their unit should always be the same Watt (perhaps with prefixes k, M, etc.). Such mark-up belongs into the relevant domain ontologies, though, _not_ into any unit ontology.

Abyssus1010:
@sm453 , oh okay. I think I understand what you mean. Will it be better if I do it this way? Where I include the addition AC and DC power classes in the OntoBMS ontology instead.
image

PR #481, opened 2022-11-28
24 develop sewerage ontology based on available datasets

NazaninMTafreshi:
Add OntoSewage


JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
```diff
@@ -0,0 +1,169 @@
+Source,Type,Target,Relation,Domain,Range,Quantifier,Comment,Defined By,Label
```
markushofmeister, line 1: I always thought we place both ths `.csv` and `.owl` file into the same folder under `JPS_Ontology/ontology/...`
@sm453 Am I mistaken?

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
```diff
@@ -0,0 +1,169 @@
...
+OntoSewage,TBox,1,http://www.w3.org/2002/07/owl#versionInfo,,,,,,
+OntoSewage,TBox,OntoSewage is an ontology developed for representing sewerage system.,http://www.w3.org/2000/01/rdf-schema#comment,,,,,,
+OntoSewage,TBox,,http://www.w3.org/2002/07/owl#imports,,,,,,
+SewerageComponent,Class,,,,,,"is the infrastructure that conveys sewage or surface runoff (stormwater, meltwater, rainwater) using sewers.",https://www.theworldavatar.com/kg/ontosewage/,Sewerage component
+SewerageNetwork,Class,,,,,,is a conglomeration of pipes and sewers that work together to collect and drain wastewater from the population or industrial centers to off-site treatment facilities.,https://www.theworldavatar.com/kg/ontosewage/,Sewerage network
+SubNetwork,Class,SewerageNetwork,IS-A,,,,contains branch pipes,https://www.theworldavatar.com/kg/ontosewage/,Sub network
+MainNetwork,Class,SewerageNetwork,IS-A,,,,contains main pipes,https://www.theworldavatar.com/kg/ontosewage/,Main network
+SewerageUsage,Class,,,,,,representing for what the sewerage is used,https://www.theworldavatar.com/kg/ontosewage/,Sewerage usage
```
markushofmeister, line 10: Comments are used very inconsistently, i.e.
- some (sub-)classes do not have any description
- some descriptions (especially the ones starting with verbs) sound odd to me. I suggest changing e.g. "is the type of a manhole." to "Type of manhole"
- as we had many discussions about the precise meaning of particular properties, please try to incorporate all those insights into the comments, so that we do not face those ambiguities again (wherever possible)

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
```diff
@@ -0,0 +1,169 @@
...
+hasElevation,Object Property,,,https://schema.org/location,https://schema.org/Elevation,,A relationship between a location and elevation,https://www.theworldavatar.com/kg/ontosewage/,has elevation
+hasElevationAboveSeaLevel,Object Property,,,https://schema.org/Elevation,http://www.ontology-of-units-of-measure.org/resource/om-2/Height,,A relationship which shows the elevation above see level of a location,https://www.theworldavatar.com/kg/ontosewage/,has elevation above sea level
+hasElevationReference,Data Property,,,https://schema.org/Elevation,String,,A relationship which shows the elevation reference,https://www.theworldavatar.com/kg/ontosewage/,has elevation reference
+hasElevationAccuracy,Data Property,,,https://schema.org/Elevation,String,,A relationship which shows the elevation accuracy,https://www.theworldavatar.com/kg/ontosewage/,has elevation accuracy
+componentID,Data Property,,,SewerageComponent,String,,sewerage component ID,https://www.theworldavatar.com/kg/ontosewage/,component ID
+hasSewagePlantID,Data Property,,,SewerageRecords,Integer,,sewerage plant ID,https://www.theworldavatar.com/kg/ontosewage/,has sewage plant ID
```
markushofmeister, line 148: I am (still) not convinced that this is the right way to do it. Potentially that's good enough for now, but ideally one would like to have a relationship along the lines of `sewerage component` - `feeds into` - `sewage treatment plant` and then several instances of `sewage treatment plant` instead of having a "random" Integer associated with the `sewerage component` which only makes sense after being looked up its meaning in some table

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
```diff
@@ -0,0 +1,169 @@
...
+ConnectionType,Class,,,,,,is the type of the node connected to a manhole.,https://www.theworldavatar.com/kg/ontosewage/,Connection type
+StreetInletConnectionType,Class,ConnectionType,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Street inlet connection type
+FictitiousManholeConnectionType,Class,ConnectionType,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Fictitious manhole connection type
+StandardShaftCircularProfile,Class,ConnectionType,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,standard shaft circular profile
+RectangularShaft,Class,ConnectionType,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Rectangular shaft
+Cover,Class,,,,,,"is a removable plate forming the lid over the opening of a manhole, an opening large enough for a person to pass through that is used as an access point for an underground vault or pipe.",https://www.theworldavatar.com/kg/ontosewage/,Cover
+CoverClass,Class,,,,,,is the type of manhole's cover.,https://www.theworldavatar.com/kg/ontosewage/,Cover class
+FictionalShaft,Class,CoverClass,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Fictional shaft
```
markushofmeister, line 89: I do not think that makes sense. I believe `Fictional` was included here in the initial data model to re-emphasise that the shaft is fictional and, hence, has no cover. Ideally, we would just not add such a triple then instead of having `Fictional` as subclass

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
```diff
@@ -0,0 +1,169 @@
...
+PolyvinylChlorideHard,Class,https://w3id.org/digitalconstruction/0.3/BuildingMaterials#Material,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Polyvinyl chloride hard
+ReinforcedConcrete,Class,https://w3id.org/digitalconstruction/0.3/BuildingMaterials#Material,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Reinforced concrete
+Stoneware,Class,https://w3id.org/digitalconstruction/0.3/BuildingMaterials#Material,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Stoneware
+GrayCastIron,Class,https://w3id.org/digitalconstruction/0.3/BuildingMaterials#Material,IS-A,,,,,https://www.theworldavatar.com/kg/ontosewage/,Gray cast iron
+ChannelType,Class,,,,,,,https://www.theworldavatar.com/kg/ontosewage/,Channel type
+https://saref.etsi.org/saref4watr/Pipe,Class,SewerageComponent,IS-A,,,,,https://saref.etsi.org/saref4watr/,
+CrossSection,Class,,,,,,"a surface or shape exposed by making a straight cut through pipe, especially at right angles to an axis",https://www.theworldavatar.com/kg/ontosewage/,Cross section
+Shape,Class,,,,,,"is the external form of a Manhole's cover, flume, shaft, or cross section.",https://www.theworldavatar.com/kg/ontosewage/,Shape
```
markushofmeister, line 39: Some of them seem redundant, e.g. `circular` vs. `circle` vs `circle profile`. What's the difference?

image

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
```diff
@@ -0,0 +1,169 @@
...
+hasMainPipe,Object Property,,,BranchConnection,https://saref.etsi.org/saref4watr/Pipe,,A relationship between a branch connection and a pipe,https://www.theworldavatar.com/kg/ontosewage/,has main pipe
+hasOutletLocation,Object Property,,,https://saref.etsi.org/saref4watr/Pipe,https://schema.org/location,,A relationship between a pipe and location (outlet),https://www.theworldavatar.com/kg/ontosewage/,has outlet location
+hasInletLocation,Object Property,,,https://saref.etsi.org/saref4watr/Pipe,https://schema.org/location,,A relationship between a pipe and location (inlet),https://www.theworldavatar.com/kg/ontosewage/,has inlet location
+http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#hasLocation,Object Property,,,BranchConnection UNION https://saref.etsi.org/saref4watr/Manhole,https://schema.org/location,,,http://www.ontologydesignpatterns.org/ont/dul/DUL.owl,
+hasDistanceFromMainPipe,Object Property,,,BranchConnection,http://www.ontology-of-units-of-measure.org/resource/om-2/Distance,,A relationship of distance between a branch connection and main pipe,https://www.theworldavatar.com/kg/ontosewage/,has distance from main pipe
```
markushofmeister, line 132: What is this? If this is intended to specify the distance of a branch connection from the end of the main pipe, this seems wrong! The branch connection in ON the main pipe, not somewhere away FROM the main pipe, only away FROM one end of the pipe

markushofmeister (commented):
I had a look over both the csv and owl file, but have to admit that I am not fully on top of your latest design choices. So, if all of you are convinved that the current design suits the purpose, please go ahead.

A few comments:
- Are you planning to include an ABox `.csv` and `.owl` file as part of this PR or will that be included separately later? Some `subClasses` are already included here and some not, e.g. `Channel Type` is not, while `Sewerage Usage` is.
- From the latest MIRO board it does not get clear which `subClasses` are finally represented in the ontology
- Some of the `subClasses` do not match the translations explicitly requested from me. Is this on purpose? e.g. 401
- Some of the concepts are still unclear to me and also the provided explanations/comments are not really helpful (e.g. connection ID) - more specific explanantions would be really helpful

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
sm453: You are correct - both the `csv` and the `owl` file should be in the same subfolder of `JPS_Ontology/ontology/`. Unfortunately, many people have done this wrong, so it is easy to follow the numerous bad examples...

sm453:
I agree with basically all of @markushofmeister's comments. I believe they should be addressed. Once done, I would suggest to go ahead and merge this PR. There will always be open questions about design choices - whether this is good enough we will see when we start using it...

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: - I moved csv file to the same directory as owl file.
- `ChannelType` sub-classes are added. We only added sub-classes which have more than 1% instances.
- The Miro board is now updated.
- Some sub-classes have same English translations. So, for these cases, we changed the translation. For example, `Manhole` sub-calss changed to `ManholeEndpoint`and `ManholeNodeType` for `Endpoint` and `NodeType` classes respectively.
- We are also not sure about the meaning of connection ID.

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: We tried to write description for as much sub-classes as we could. Some sub-classes such as `40T` or `12.5T` are not clear in definition.
Some of the comments are updated as you suggested.

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: If we change it to this, do we have any instance for that in the Data? Could you please explain more about this?

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: Redundant sub-classes are removed.

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: `hasDistanceFromMainPipe` is changed to `relativeDistanceOnMainPipe`. Also, its comment is changed.

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: `FictionalShaft` sub-class is changed to `Fictional`.

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
markushofmeister: > * Some sub-classes have same English translations. So, for these cases, we changed the translation. For example, `Manhole` sub-calss changed to `ManholeEndpoint`and `ManholeNodeType` for `Endpoint` and `NodeType` classes respectively.

I am not sure if that is a very useful approach (but maybe it is). I think the terms `Type` and `Node` add unnecessary confusion here:
image

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
markushofmeister: The guiding principle should be that somebody not too familiar with the topic/domain should get a rough idea about a concept by reading those comments. I am (still) not convinved that this is the case for all concepts, e.g. the comment `Being extended` is quite meaningless, especially for the concept `Constructed` as a Functional State. Please revisit the comments and try to be more specific

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
markushofmeister: What's the difference between `round` and `circle`?

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
markushofmeister: Sorry for not making may point clear enough: I believe they used `Fictional` as cocer class to set the cover class for fictional shafts (as relational databases might require some sort of entry, even if they are not very meaningful). In a graph, on the contrary, we might just decide not to instantiate this triple at all. This would mean we do not need the `Fictional` class and for all fictional shafts we simply do not instantiate any cover class, which seems more sensible then instantiating a fictional cover class (at least to me). Is that clearer?

JPS_Ontology/KBTemplates/TBox/OntoSewage.csv
NazaninMTafreshi: Based on Cambridge dictionary:
round is shaped like a ball or circle, or curved.
circle is a continuous curved line, the points of which are always the same distance away from a fixed central point, or the area inside such a line.
Circle is definitely a 2-D. While ‘round' we can refer to 3-D.
But, I am not sure if here round is different with circle or not. Because, it can also refer to a circle.

markushofmeister: I guess we are not concerned about spheres here, but if the data contains both `round` and `circle`, then please go ahead keeping both.

markushofmeister (approved):
I guess this is as good as it gets (for now). Just two clarification questions:

1) `Rainwater` is a subclass of both `Sewerage Usage` as well as `Sewerage Fluid`. Is this intentional? Even though both concepts might have subclasses calles `Rainwater`, conceptually they do not feel congruent to me 

2) I am (still) not a big fan of specifying `subClasses` without specifying in the comments why they exist. If one cannot specify what this subsclass specifically refers to (in comparison to another subclass, one could argue why this subclass exists in the first place... e.g. `ShaftNode` vs. `Structure`


image

PR #482, opened 2022-11-29
hotfix-avgsqmprice-ensure-new-derivation-iri

markushofmeister:
Ensure that a new avgsm_price_iri is used for every update (to avoid potential issues with IRIs already associated with a derivation). All tests pass


jb2197 (approved):
Looks good to me.

PR #485, opened 2022-11-29
Add ability to add additional containers to a stack when starting it up

gpeb2:
These changes should allow user defined containers to be spun up within the stack using the stack-manager tool.
This will ensure that the new containers can access the endpoint information (URL, username, password, etc.) for other containers in the stack and connect to them using the stack-client APIs.

### DIFF
--- a/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
+++ b/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.8'
 services:
   stack-client:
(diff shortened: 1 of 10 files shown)


sm453 (changes requested):
Tested, works.

Can you please add a sentence to the readme file that clarifies that this allows inclusion of additional containers both at initial start-up of a new stack and addition of new containers to an already running stack?

Can you also add a sentence that clarifies that, at time of addition, the docker image needs to have been built but the container must not already have been started?

sm453 (approved):

PR #486, opened 2022-11-29
196 dev average square metre price agent update

jb2197:
In this PR, `AverageSquareMetrePriceAgent` is updated to use the latest `py4jps` and `pyderivationagent`, specifically, the `agentlogging` installing from GitHub is removed. The design of using both `setup.py` and `requirements.txt` is also adapted.

All tests passed.

### DIFF
--- a/Agents/AverageSquareMetrePriceAgent/Dockerfile
+++ b/Agents/AverageSquareMetrePriceAgent/Dockerfile
@@ -24,12 +24,12 @@ COPY ./tmp_stack ./tmp_stack
 WORKDIR /app
 COPY ./README.md .
 COPY ./setup.py .
+COPY ./requirements.txt .
 
 # Install Python package and the required libraries
 RUN python -m pip install --upgrade pip
+RUN python -m pip install -r requirements.txt
 RUN python -m pip install -e .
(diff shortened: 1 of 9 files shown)


markushofmeister (changes requested):
Honestly speaking, I am not a big fan of _splitting_ `setup.py` and `requirements.txt`. I do understand that they are meant for different purposes, but my impression is that they rather add confusion than being actually helpful to most people.
If you have strong feelings about it, please
1) Add a short section to the `README` explaining why there are both files
2) Ideally amend the `PropertyValueEstimation` agent similarly (in a subsequent PR). Ideally, we should keep things consistent

jb2197:
> 1. Add a short section to the `README` explaining why there are both files

Addressed in commit https://github.com/cambridge-cares/TheWorldAvatar/pull/486/commits/19d2f741b526d48d433efa9fdd1421dd08e30407

> 2. Ideally amend the `PropertyValueEstimation` agent similarly (in a subsequent PR). Ideally, we should keep things consistent

Opened PR https://github.com/cambridge-cares/TheWorldAvatar/pull/489

markushofmeister (approved):

PR #488, opened 2022-11-30
Dev virtual sensor update

kokfoongcmcl:
Merging this because it won't be worked on for a while, 99% of changes are isolated within the JPS_VIRTUALSENSOR folder

Changes outside the JPS_VIRTUALSENSOR folder:
- WeatherAgent: added ability to request data for a specific timestamp instead of just getting the latest data
- FileServer: integrated into the stack

Major notable change:
- No longer using Episode for dispersion calculations, now using AERMOD which is much easier to compile and the code looks tidier if we want to dive into it

Dispersion calculations are completed in about 3 commands documented in the README
1) Initialise a dispersion area by giving a WKT literal
2) Trigger ship input agent to get latest ship location and speeds
3) Trigger update derivation for the dispersion calculation, which triggers update the derivation of ship emission calculations internally before proceeding to start the dispersion calculations


Agents/WeatherAgent/docker-compose.yml.template
```diff
@@ -10,7 +10,7 @@ services:
         condition: none
     environment:
       - STACK_NAME=${STACK_NAME}
-      - API_KEY=
+      - API_KEY=21db8f6413957d7fdb6cfd98e05e03eb
```
sm453, line 13: Are you sure you want to commit this? I'm guessing not... Also, what is desperately needed here is some documentation, probably in the main Readme file of this agent (which doesn't even exist!) that the user needs to add an API key here before attempting to build the agent, together with an explanation of what API this is meant to be, ideally with a web-link where to obtain such a key.

JPS_VIRTUALSENSOR/DispersionInteractor/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+#==================================================================================================
+FROM maven:3.8.3-adoptopenjdk-11 as retriever
+
+# Copy in Maven settings templates and credentials 
+COPY docker/credentials /root/credentials
+COPY docker/.m2 /root/.m2
+
+# Populate settings templates with credentials
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 11: "latter, but not the former"

JPS_VIRTUALSENSOR/EmissionsAgent/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+#==================================================================================================
+FROM maven:3.8.3-adoptopenjdk-11 as retriever
+
+# Copy in Maven settings templates and credentials 
+COPY docker/credentials /root/credentials
+COPY docker/.m2 /root/.m2
+
+# Populate settings templates with credentials
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 11: "latter, but not the former"

JPS_VIRTUALSENSOR/ShipInputAgent/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+#==================================================================================================
+FROM maven:3.8.3-adoptopenjdk-11 as retriever
+
+# Copy in Maven settings templates and credentials 
+COPY docker/credentials /root/credentials
+COPY docker/.m2 /root/.m2
+
+# Populate settings templates with credentials
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 11: "latter, but not the former"

sm453 (changes requested):
I understand that this project isn't 'finished'. However, what worries me is that you already know that you won't be looking at this for quite some time. In reality, this may even end up on someone else's plate. For that reason, please expand the Readme files for the ShipInputAgent and the EmissionsAgent, and please add one for the WeatherAgent. An example of a good Readme is: https://github.com/cambridge-cares/TheWorldAvatar/tree/dev-historical-pump-data-instantiation/Agents/HistoricalPumpDataInstantiation

Agents/WeatherAgent/docker-compose.yml.template
kokfoongcmcl: It's in the Setup section of the readme, the key will be removed

sm453 (approved):

PR #490, opened 2022-11-30
46 dev extend ontobms to include electrical power classes

Abyssus1010:
Hi @sm453 , I have created a new branch and added in the electrical power classes into the ontoBMS ontology. 
image


Abyssus1010:
Can you help me to review the changes when you have the time? Thank you.

sm453 (changes requested):
Surely, the definition of om:Power as a subclass of om:Quantity is already made in OM (where it should be made) and does not need to be repeated here in OntoBMS?

Abyssus1010:
> 

Oh okay. So as long as the definitions exist in another ontology, there is no need to repeat it for new ontologies and extensions. I will take note of that for future additions/changes. I have made the changes accordingly.

sm453 (approved):

PR #500, opened 2022-12-02
195 dev optimise derivation sparql

jb2197:
In this PR, the derivation variable identified from the first Basic Graph Pattern using the property paths is bound to another variable in `DerivationSparql::getDerivations` to limit the possible combination of the following GraphPattern where information about timestamp/derivation inputs/derivation type is queried. This was the root cause of the issue observed in the King's Lynn use case for derivation MVP where the agent hangs at `DerivationSparql::getDerivationWithImmediateDownstream` when calling `DerivationClient::cleanUpFinishedDerivationUpdate` to update the async derivation at `Finished` status. This issue is due to a large number of triples (~700k) existing in the knowledge graph and such a situation is mimicked in the stress test and working fine.

### DIFF
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -8,7 +8,7 @@
 
     <!-- Please refer to the Versioning page on TheWorldAvatar wiki for
     details on how version numbers should be selected -->
-    <version>1.30.0</version>
+    <version>1.31.0</version>
 
     <!-- Project Properties -->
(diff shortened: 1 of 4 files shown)


sm453 (approved):
There are a couple of spelling mistakes in the comments, but apart from that, fine by me.

kokfoongcmcl (approved):
Looks fine in general, I think that the stress tests do not need to be committed, but up to you

PR #504, opened 2022-12-04
Derivation MVP

jb2197:
In this PR, scripts are provided to mark up a derivation chain consisting of mixed sync and async in the context of flood impact assessment. Both automated information generation and existing information updates are demonstrated. The final results are visualised in DTVF.


markushofmeister (approved):
Looks good to me (although, I did not test/run the agents on my machine).

Please make sure that the implemented markup logic is properly reflected in the UML Lucidchart diagrams, i.e. by amending the diagrams and/or adding notes with comments there

jb2197:
> Looks good to me (although, I did not test/run the agents on my machine).

I will fix the version of the pushed docker images on GitHub once all reviewers are happy with the PR - just in case anyone would like to reproduce this work at a later date.

> Please make sure that the implemented markup logic is properly reflected in the UML Lucidchart diagrams, i.e. by amending the diagrams and/or adding notes with comments there

I have updated the diagrams/left comments accordingly.

sm453 (approved):
As a mock-up for the paper, it's fine.

In future, we should probably reconsider the use of geojson files. Rather than read/write/amend/copy around such static files, we may want to e.g. programmatically manipulate GeoServer layers...

markushofmeister:
> As a mock-up for the paper, it's fine.
> 
> In future, we should probably reconsider the use of geojson files. Rather than read/write/amend/copy around such static files, we may want to e.g. programmatically manipulate GeoServer layers...

The actual King's Lynn implementation will not use geojson files, but the Stack incl. visualisation served by Geoserver. This was simply a way to circumvent the use of the Stack at all for the paper.

PR #506, opened 2022-12-05
41 dev solarkatasteragent

YiKaiTsai:


Agents/SolarkatasterAgent/README.md
```diff
@@ -0,0 +1,60 @@
...
+### Building the agent
+To build and start the agent, you simply need to spin up a container from the image.
+In Visual Studio Code, ensure the Docker extension is installed, then right-click docker-compose.yml and select 'Compose Up'.
+Alternatively, from the command line, and in the same directory as this README, run
+
+```
+docker-compose up -d
+```
+
+The agent is reachable at "solarkataster_agent/run" on localhost port 10101.
```
sm453, line 60: Please move this statement to the "post request" section above. It would also be very helpful to include an example curl command that one can just copy-paste.

sm453:
The readme file refers to "the Solarkataster data", even though there are really two separate datasets. I guess one has to run the agent twice, once for the Solarthermie table, and once for the Photovoltaik table? And I assume that it doesn't cause clashes or other problems if one uses the same time series database for both? It would be helpful if you could clarify both these points in the readme.

Agents/SolarkatasterAgent/README.md
sm453: e.g. `curl -X POST --header "Content-Type: application/json" -d "{'table':'solarthermie','chunk':50}" http://localhost:10101/solarkataster_agent/run`

sm453 (changes requested):
Tested on solar thermal dataset - runs.

During the image build process, it reports `solarkataster_agent Error` (without any further details), although this does not seem to be of any consequence. The image builds fine, the agent runs, and there is nothing in the logs that indicates any problems. I guess we can ignore this.

In future, we probably need to re-think our approach to instantiating huge numbers of time series that duplicate data, but not as part of this pull request.

In order to merge this PR, please address the above comments on the readme file.

YiKaiTsai:
> image



> The readme file refers to "the Solarkataster data", even though there are really two separate datasets. I guess one has to run the agent twice, once for the Solarthermie table, and once for the Photovoltaik table? And I assume that it doesn't cause clashes or other problems if one uses the same time series database for both? It would be helpful if you could clarify both these points in the readme.

Thanks for pointing these out. I should have been more clear in the readme. 

The agent should just be ran once, either for the Photovoltaik dataset or the Solarthermie dataset. I have checked, for Piramsens, all the relevant parameters in the agent has the same value for both Photovoltaik and Solarthermie, so running for just one dataset is sufficient. If trying to run twice, once for each, problem will occur and the time series client will raise error message along the line of "Time series already exists for dataIRI".

Also, the agent should not be run for the full Photovoltaik dataset nor the full Solarthermie dataset, since the volume will be too large for the stack to handle and the agent will fail in the end.

Since we only need the time series for Piramsens, in my local testing, I just uploaded  `a00000027.*` files from Photovoltaik as a table to the stack and passed that table in the post request.

sm453 (approved):

PR #509, opened 2022-12-05
1451 dev update timeseries agents to use latest timeseries client version

Abyssus1010:


sm453 (approved):
Changes look ok. I have tested the Thingspeak agent (works), but none of the others.

PR #514, opened 2022-12-07
Dev mods agent sensitivity analysis

cmclscourtney:
adding sensitivity analysis to MoDS Simple Agent


gpeb2 (approved):

gpeb2 (changes requested):
It would probably be better to have the save and load surrogate flags in the algorithm node, in case there are multiple surrogate algorithms. For the load flag this could be placed in an otherwise empty algorithm node.

cmclscourtney:
> It would probably be better to have the save and load surrogate flags in the algorithm node, in case there are multiple surrogate algorithms. For the load flag this could be placed in an otherwise empty algorithm node.

addressed in [f11af2a](https://github.com/cambridge-cares/TheWorldAvatar/pull/514/commits/f11af2adc6ec521e6f820ddfc3519458dd4ac7d7)

gpeb2 (approved):

PR #516, opened 2022-12-08
Dev elaborate fia documentation

markushofmeister:
I tried to clarify a few things in the current README and included example `.sparql` files to retrieve both meta and time series data based on a simple example triple set (also provided in the readme). I hope that helps people to get up to speed quicker (and more independently).

The only remaining question is the return values of the time series sparql query, when both `Measurement` and `Forecast` are used. Is the below example as expected or
1) Should `Forecast` and `Measurement` come in the same column?
2) Is the `Name` amendment actually required if they come as `Forecast`  vs. `Measurement` or will they be displayed differently already (i.e. will the user be able to distinguish which is forecasted and which is measured)?

image


mdhillman:
1) I believe so, but I should test it to be sure.
2) These will be displayed the same to the user, so updating the name is the only real way to distinguish between measured & forecasted quantities.

markushofmeister:
> 1. I believe so, but I should test it to be sure.
> 2. These will be displayed the same to the user, so updating the name is the only real way to distinguish between measured & forecasted quantities.

re 1) Once that's tested, we should make sure to align the README accordingly. The current description rather suggests having `Forecast` and `Measurement` in separate columns as the variable names are different. If that's not expected than the variables names should be aligned.

Feel free to merge this PR if useful once tested.

markushofmeister:
I have tested 1) with the instantiated MetOffice data and can confirm that having forecasts coming as `Forecast` and actual readings as `Measurement` does NOT work. Both IRIs need to come as the same variable, i.e. `Measurement`. I have updated the README accordingly.

Agents/FeatureInfoAgent/README.md
```diff
@@ -61,29 +61,61 @@ Follow the below configuration steps within the local `queries` directory.
...
 </p>
 
-Queries for timeseries data need to return the measurement/forecast IRIs (that will be used to grab the actual values from PostGreSQL), as well as parameters associated with each measurement/forecast. Required columns are `Measurement` (or `Forecast`) containing the IRI, `Name` containing a user facing name for this entry, and `Unit` containing the unit (which can be blank). In this case, any other columns reported by the query **will** be picked up and passed back to the visualisation as regular key-value properties.
+<ins>Queries for timeseries</ins> data need to return the measurement/forecast IRIs, i.e. the IRIs which are connected via `ts:hasTimeSeries` to the actual timeseries instances. Those IRIs will be used to grab the actual values from PostGreSQL as well as parameters associated with each measurement/forecast. Required columns are `Measurement` containing the IRI, `Name` containing a user facing name for this entry, and `Unit` containing the unit (which can be blank). In this case, any other columns reported by the query **will** be picked up and passed back to the visualisation as regular key-value properties.
```
markushofmeister, line 80: "In this case, any other columns reported by the query **will** be picked up and passed back to the visualisation as regular key-value properties."

Where would those key-value pairs be displayed? I also tested with the MetOffice data and the "LastUpdated" value for forecasts does not appear in the DTVF despite being returned by the SPARQL query. So, we might need to investigate or update the README accordingly.

image
image

markushofmeister:
@mdhillman As the FIA PR got merged now, please make a decision about this PR asap before risking it getting stale/outdated. I believe the updated documentation is useful, but feel free to disagree ;-)

mdhillman (approved):

PR #517, opened 2022-12-08
dev-dtvf-3.4.0: Clipping Planes

### LINKED ISSUE #491: DTVF: Cesium 3D tiles clipping plane feature
We now have highly detailed building representations that include numerous internal features such as furniture and sensors that we want to be able to click on and view metadata etc. In order to be able to see such features inside buildings, the clipping plane feature provided by Cesium is exceedingly useful. This would need to be added to the DTVF, with suitable UI controls.
Clipping-JS-implementation-2

mdhillman:
This is a draft PR to allow @sm453 to test the _basic_ functionality of clipping planes (closes #491). Details on how to enable the clipping plane feature, and its current limitations can be found in the README for the example Cesium visualisation.

Note that this feature branch has come from main, so does not include any of the recent fixes associated with the FeatureInfoAgent. It is possible to merge the dev-feature-info-agent branch into this, but I want to avoid tying these features together if at all possible; I would rather wait until the FIA branch if merged into main.


sm453:
Tested it. Works nicely for the (BIM) tilesets that we have created ourselves, even if the bounding box is not properly set:
image

It does not seem to work for the LoD2 tileset. Most of the buildings just disappear, and there is no plane visible. However, since there is no point in using a clipping plane for low levels of detail anyway, I would suggest we ignore this problem.

In order to make this feature production-ready, having a control element in the UI to switch clipping on/off (for particular tilesets, or generally?) is _critical_.

In addition, we need either:
1) A (separate!) control to make the clipping plane(s) in-/visible.
or:
2) Keep the clipping plane(s) invisible at all times and add some other way of adjusting the height of the plane.

The second option has the advantage that one does not need to worry about awkward-looking/poorly positioned bounding boxes, and that selecting features through the plane is (presumably) not an issue. This is also, I believe, how it's done in other pieces of software. Yet another point is that, with a visible plane, when closely zoomed in, one can no longer navigate the map, as any click will be interpreted as an attempt to adjust the height of the plane.

mdhillman:
The clipping plane feature has now been updated with the below major changes. I believe it is ready for testing and (depending on the results of that testing) subsequent release.

Changes:
* Height of clipping plane is now determined via a slider
* Can select entities through the clipping plane
* Clipping plane geometry can be shown/hidden
* Supports planes for multiple tilesets (although only one can be viewed at a time)
* A set of, labelled, non-uniform increments heights can be added to the slider

Please see the "Clipping Planes" section of the Cesium example's README file for details on how to enable & configure this feature. Alternatively, the example Cesium visualisation contains some tilesets with the clipping feature enabled.

sm453 (changes requested):
Works beautifully.

Just a few points:
1) When the slider is visible and one collapses the side-panel, the position of the slider is currently not updated, leading to the slider 'floating' in the middle of the screen.
2) Sometimes (don't know how to reproduce) the current elevation of the clipping plane is shown to four decimals (rather than more sensibly two or even one). Related to that (I guess), the up and down spinner buttons don't seem to be working whenever that happens (but they do work when it doesn't).
3) At present, the string labels `Clipping Planes`, `Enable clipping planes?`, `Show plane geometry?`, and `Target layer:` that appear on the generated page seem to be hard-coded in `clip_handler.ts`. Would it be possible to move that e.g. into `index.html` (so that one can easily customise those strings)?

mdhillman:
I have implemented the aforementioned changes and can no longer produce any of the mentioned issues (either on 4K or HD monitors).

jlai2021 (approved):
I have tested this with two large BIM models, using self-defined terrain. The DTVF performs as designed and I found no major issues.

sm453 (approved):
Tested. All good.

PR #520, opened 2022-12-08
44 dev sewerage network instantiation

xuehansong:


Agents/SewageAgent/README.md
```diff
@@ -0,0 +1,54 @@
...
+curl -X POST --header "Content-Type: application/json" -d "{
+\"SewerageNetworkAgent\\":\"DataInstantiation\"}"  http://localhost:1080/seweragenetwork-agent/performsewageupdate
+```
+
+If the agent runs successfully, you should see a returned JSON Object that is similar to the one shown below.
+```
+{"Result":"Data has been instantiated."}
+```
+
+#### Import the CSV data file into the agent
```
markushofmeister, line 53: What does that mean? In case I "simply" want to instantiate the "original" sewage network as is
1) What would I need to do exactly? 
2) What .csv file(s) would I need to use? And why .csv(s) in the first place? Wasn't the initial data in .xml, so, where do those .csv files come from and how are they generated?

markushofmeister, line 53: Are those committed .csv files the actual data??

markushofmeister:
Is the instantiated data somehwere publicly available? To be honest, I find it tricky to judge solely from the code whether the instantiation works correctly. A (temporary) publicly available SPARQL endpoint would be much quicker to screen

Agents/SewageAgent/README.md
xuehansong: > What does that mean? In case I "simply" want to instantiate the "original" sewage network as is
> 
> 1. What would I need to do exactly?
> 2. What .csv file(s) would I need to use? And why .csv(s) in the first place? Wasn't the initial data in .xml, so, where do those .csv files come from and how are they generated?

We converted the initial data in .xml to CSV for data treatment with a python script. CSV is used because the instantiation code was initially written to take in CSV files.  All the CSV files can be found at "Dropbox (Cambridge CARES)\IRP3 CAPRICORN shared folder\HSXUE\Sewage Instantiation Data"

The current dockerised code only takes in "HG_instances_0.csv" and "KG_instances_0.csv" files. There are 7 HG_instances and 11 KG_instances CSV files. Only need to change the stringpath names in the Dockerfile for instantiation the other CSV files. Why like this: because last minute we notice the agent cannot take in one CSV file with all the data, it will truncate the data at some where due to the column number exceeds its limit. Previously when we tested one CSV file with only German abbreviation letters for the instances, it worked fine. Now this problem arised when we replaced all the abbreviation letters with full English meaning.

xuehansong: Yes, they are the actual data with full English meaning for each HG and KG tags. The original data comes only with abbreviation letters in most of the tags

xuehansong:
Not publicly available yet. I instantiated it at my local PC and I think there are more than 5 millions triples.  For "A temporary publicly available SPARQL endpoint", sounds good and where can I do it so that it is public available and you can check?

markushofmeister:
> Not publicly available yet. I instantiated it at my local PC and I think there are more than 5 millions triples. For "A temporary publicly available SPARQL endpoint", sounds good and where can I do it so that it is public available and you can check?

In that case: If it does not contain any time series data, a simple triple download should suffice. Might be easier to verify the "correct" instantiation in that way as compared to running everything locally again

Agents/SewageAgent/README.md
markushofmeister: I suggest you include those python scripts into this agent, potentially into some `preprocessing` subfolder. If the original data comes in `.xml` that should be the input to the agent. Otherwise, there is no control over where things break: This could be the agent, this could be the preprocessing, this could be extraction from xml into csv, ...

Haven some "random" `.csv` file as input without documentation how it has been created feels wrong to me.

markushofmeister (changes requested):
I have not scrutinised the instantiated data (yet), but I believe we should use the `.xml` file as genuine input to the agent and include the pre-processing as part of the agent.

Agents/SewageAgent/README.md
xuehansong: sure, sounds good

sm453: > Not sure if we want to version control the actual data here in this public repository (if at all, which I am rather against). But I am also not clear on what this version controlled data here represents compared to the data you mention to be located on Dropbox. @sm453 What's your take on that?

Absolutely not. Under no circumstances should actual data be version-controlled in this (or any other) git repository. Especially not confidential data - version-controlling this should never even be considered.

sm453:
There seem to be several `csv` data files in the config folder. Based on their size, I'm guessing this is the full dataset?!? In any case, such things should _never_ be version-controlled. Please remove these files asap.

Agents/SewageAgent/Dockerfile
```diff
@@ -0,0 +1,61 @@
+# First stage: build war file
+#==================================================================================================
+FROM maven:3.6-openjdk-11-slim as builder
+
+# Copy all files into root's home, including the source, pom file, ./m2 directory, credentials and config files
+ADD . /root
+
+# Populate settings templates with credentials, repo name
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 10: "latter, but not the former"

Agents/SewageAgent/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+RUN sed -i "s|MASTER_PASSWORD|$(mvn --encrypt-master-password master_password)|" settings-security.xml
+RUN sed -i "s|REPO_USERNAME|$(cat ../credentials/repo_username.txt)|;s|REPO_PASSWORD|$(cat ../credentials/repo_password.txt|xargs mvn --encrypt-password)|" settings.xml
+
+# Build
+WORKDIR /root/SewerageNetworkAgent/
+RUN --mount=type=cache,target=/root/.m2/repository mvn package
+# RUN mvn package
```
sm453, line 17: Please remove this comment. Given the previous line, it's obsolete.

Agents/SewageAgent/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+ENV KGDATA7="/root/KG_instances_7.csv"
+ENV KGDATA8="/root/KG_instances_8.csv"
+ENV KGDATA9="/root/KG_instances_9.csv"
+ENV KGDATA10="/root/KG_instances_10.csv"
+ENV KGMainNet="/root/KGMainNetwork.csv"
+ENV KGSubNet="/root/KGSubNetwork.csv"
+ENV Branch0="/root/HG_main_sub_instances_final_0.csv"
+ENV Branch1="/root/HG_main_sub_instances_final_1.csv"
+ENV Branch2="/root/HG_main_sub_instances_final_2.csv"
+ENV Branch3="/root/HG_main_sub_instances_final_3.csv"
```
sm453, line 55: Clearly, this will need to be re-worked. Data files (even once you've removed them from the repository) should not be stored in a folder named `config`...

Agents/SewageAgent/SewerageNetworkAgent/WEB-INF/web.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<web-app xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://xmlns.jcp.org/xml/ns/javaee" xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd" metadata-complete="false" version="3.1">
+    <display-name>seweragenetwork-agent</display-name>
+    <description>This agent instantaites the data for sewage network.</description>
```
sm453, line 4: Please fix the typo.

Agents/SewageAgent/SewerageNetworkAgent/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+		<version>1.0.0</version>
+	</parent>
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.1</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
```
sm453, line 20: Please use something more recent, e.g. at least 11.

Agents/SewageAgent/SewerageNetworkAgent/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+		</dependency>
+		<dependency>
+	        <groupId>xml-apis</groupId>
+	        <artifactId>xml-apis</artifactId>
+	        <version>1.4.01</version>
+		</dependency>
+     <dependency>
+		  <groupId>uk.ac.cam.cares.jps</groupId>
+		  <artifactId>jps-base-lib</artifactId>
+		  <version>1.20.2</version>
```
sm453, line 84: Please use something more recent, e.g. 1.30.0.

Agents/SewageAgent/SewerageNetworkAgent/pom.xml
```diff
@@ -0,0 +1,109 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>uk.ac.cam.cares.jps</groupId>
+  <artifactId>seweragenetwork-agent</artifactId>
+  <version>1.0.0</version>
+  <packaging>war</packaging>
+	<!-- Parent POM -->
+	<parent>
+		<groupId>uk.ac.cam.cares.jps</groupId>
+		<artifactId>jps-parent-pom</artifactId>
+		<version>1.0.0</version>
```
sm453, line 11: Please use something more recent, e.g. 2.0.0.

sm453 (changes requested):

Agents/SewageAgent/Dockerfile
xuehansong: These files will be generated by data treatment and put under resources file so that the agent can import these files there for instantiation.

Agents/SewageAgent/SewerageNetworkAgent/pom.xml
xuehansong: I have done all the amendments according to the comments. Thank you very much for the feedback and please let me know any further changes that are required.

Agents/SewageNetworkAgent/SewageNetworkAgent/src/main/java/uk/ac/cam/cares/jps/agent/sewage/SewerageNetworkAgent.java
```diff
@@ -0,0 +1,697 @@
...
+					.addInsert(NodeFactory.createURI(KB + "SewerageRecords" + HG_Instance_Name), NodeFactory.createURI(RDF_TYPE), NodeFactory.createURI(OS + "SewerageRecords"))
+
+
+					.addInsert(NodeFactory.createURI(KB + "SewerageRecords" + HG_Instance_Name), NodeFactory.createURI(OS + "hasOwnershipType"), NodeFactory.createURI(KB + "OwnershipType" + HG402))
+					.addInsert(NodeFactory.createURI(KB + "SewerageRecords" + HG_Instance_Name), NodeFactory.createURI(OS + "hasFunctionalState"), NodeFactory.createURI(KB + "FunctionalState" + HG401))
+					.addInsert(NodeFactory.createURI(KB + "OwnershipType" + HG402), NodeFactory.createURI(RDF_TYPE), NodeFactory.createURI(OS + "OwnershipType"))
+					.addInsert(NodeFactory.createURI(KB + "FunctionalState" + HG401), NodeFactory.createURI(RDF_TYPE), NodeFactory.createURI(OS + "FunctionalState"));
+			UpdateRequest SewerageComponentHG_ur = SewerageComponentHG_ub.buildRequest();
+			AccessAgentCaller.updateStore("http://host.docker.internal:48888/ontosewage", SewerageComponentHG_ur.toString());
```
sm453, line 215: Same comment as in PR #339 (https://github.com/cambridge-cares/TheWorldAvatar/pull/339#discussion_r1085418769): Please don't hard-code the endpoint URL here (and elsewhere below) - use a variable whose contents can be user-specified.

xuehansong: Yes, I have done the same amendment as it the case of PR#339. Please check. Thank you.

markushofmeister (changes requested):
I mainly looked into the nature of the instantiated data and would like to share the following observations. Please clarify or amend the instantiation accordingly:

1) Multiple `BranchConnection` s basically only has a type declaration and a relative distance on the main pipe. According to the ontology (v1.8) I would expect many more relationships (location, ...) for relevant ones
image

2) There is not a single `dul:hasLocation` relationship in the data. Is that expected?

3) There are multiple misalignments between MIRO - Ontology (on github) - Instantiated data, e.g.
- Location: should be `schema:location` according to ontology, but is `schema:Location` in Miro and data
- hasInletLocation: should be `ontosewage:hasInletLocation` according to ontology and Miro, but is `ontosewage:InletLocation` in data (same issue for outlets)
Please resolve that. It is important to have **ONE** consistent representation

4) Multiple relationships seem to have get messed up between `Location` and `Elevation` concepts. Furthermore (and that might be personal), I find the naming `InletLocation` and `OutletLocation` misleading: The relationships `hasInlet...` and `hasOutlet...` make perfect sense, but the linked Locations itself are just Points then and should not have inlets and outlets anymore. I know what you mean by it, but it is slightly confusing
image

image

5) Some of the pipes are of `PipeTypeOther`, even though that doesn't exist in ontology or Miro

I am more than happy to have another more detailed review after those initial observations are addressed.

Agents/SewageNetworkAgent/Preprocessing/README.md
```diff
@@ -0,0 +1,64 @@
+# **Sewage Data Preprocessing**
+
+This project provides functionality to preprocess data required for instantiating a sewage network. It provides tools to analyze the XML data structure and clean the provided data for further processing.
+
+## **Description of modules**
+
+The `./Data/raw/` directory must contain all the required input files including `dictionary.csv`, `main_datamodel.ttl`, `sub_datamodel.ttl`, `priority_tags_in_ontology.csv`, `sub_network_connections_consolidated_wo_geodata.csv`, and `HALTUNGEN-ALLE-5-2022.xml`. Note that `main_datamodel.ttl`, `sub_datamodel.ttl`, and `sub_network_connections_consolidated_wo_geodata.csv` are created as output files by the scripts `instantiate_main_network.py`, `instantiate_sub_network.py`, and `assess_geospatial_sub_network.py` in a private git repository.
```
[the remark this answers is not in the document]
xuehansong: I have added the explanation for 6 files inside

xuehansong:
@markushofmeister regarding to your comments above:

1.  There is no location information for BranchConnection as it is not given in the raw data and we could not figure it out that time
2.   dul:hasLocation is missed and now already added inside
3.  I have checked entire code and multiple misalignments have been fixed
4. Naming with InletLocation and OutletLocation have been changed to  LocationI and LocationO in the code
5. "PipeTypeOther" has been removed

Thank you for the feedback

markushofmeister:
> @markushofmeister regarding to your comments above:
> 
> 1. There is no location information for BranchConnection as it is not given in the raw data and we could not figure it out that time
> 2. dul:hasLocation is missed and now already added inside
> 3. I have checked entire code and multiple misalignments have been fixed
> 4. Naming with InletLocation and OutletLocation have been changed to  LocationI and LocationO in the code
> 5. "PipeTypeOther" has been removed
> 
> Thank you for the feedback

Have you made the newly instantiated triples available somewhere? Otherwise, it's hard for me to check the instantiation

xuehansong:
> > @markushofmeister regarding to your comments above:
> > 
> > 1. There is no location information for BranchConnection as it is not given in the raw data and we could not figure it out that time
> > 2. dul:hasLocation is missed and now already added inside
> > 3. I have checked entire code and multiple misalignments have been fixed
> > 4. Naming with InletLocation and OutletLocation have been changed to  LocationI and LocationO in the code
> > 5. "PipeTypeOther" has been removed
> > 
> > Thank you for the feedback
> 
> Have you made the newly instantiated triples available somewhere? Otherwise, it's hard for me to check the instantiation

Noted, I am running it now. It will take about 24 hours to complete the instantiation and export the triples out

xuehansong:
@markushofmeister Hi newly instantiated triples available at 

(https://www.dropbox.com/s/vupvt16b5cedx85/OntoSewage_20230414.nt?dl=0)

or Dropbox (Cambridge CARES)\IRP3 CAPRICORN shared folder\HSXUE\Sewage Instantiation Data\OntoSewage_20230414

markushofmeister (changes requested):
I checked the instantiated triples and have the following comments:

1) `isAssociatedWith` connects `SewerageComponent`s with `SewagePlant`s although this relationship is not included in the [ontology](https://github.com/cambridge-cares/TheWorldAvatar/blob/98f150e6a0d8ce5d11fc6d3bb3d3577b08b6d15a/JPS_Ontology/ontology/ontosewage/ontosewage.owl#L551). Please add (ideally in separate PR to fix the ontology)
2) `is isAssociatedWith`, when used for `Location` instances, seems to connect the instances directly with the ontology concept IRI. This does not seem correct. You either want to create individual instances or (potentially preferred) create instances as part of the ABox of the ontology
3) Some locations have multiple contradicting geo coordinates, e.g. `<https://www.theworldavatar.com/kb/ontosewage/GeoCoordinateKG0>`. This should be resolved.
4) Some locations have multiple contradicting elevations, e.g. `<https://www.theworldavatar.com/kb/ontosewage/GeoCoordinateKG0>`. This should be resolved.
5) Please verify that ABox instantiations are correct. For example, all `ChannelType` and `CoverClass` instances have the same type, despite the definition of multiple subclasses
image

Once those are fixed, please export another set of triples to check and re-request review.

xuehansong:
@markushofmeister Thanks for the feedback.

For point 3, do you mean there are two geo coordinates? like the one below when I query: select * where {<https://www.theworldavatar.com/kb/ontosewage/GeoCoordinateKG0> ?x ?y}

image

For the 2 geo coodinates, they are the data extracted from the xml raw data files 

Similarly for point 4 about the elevation: there are two values extracted for each KG from the raw data
image

markushofmeister:
> Similarly for point 4 about the elevation: there are two values extracted for each KG from the raw data

I do know that there can be different locations for the same location ID (e.g., `KG01`) in the raw data. However, I believe they refer to different locations of this element, e.g. when considering a shaft: one location (incl. elevation) is likely to refer to its surface location (e.g., location of cover), while the other locations refers to subsurface location (e.g., location of intersection with pipe).
If we cannot figure this out, fine, but I believe the current way of representing is not really meaningful because 1) one does not know which location is the relevant one and 2) when querying for it it could happen now that one gets location 1 with elevation 2 (which does not make any sense)

xuehansong:
@markushofmeister Hi, I have finished another run of the agent after amendment according to your feedback point 2 to 5. The new instantiated data is available at

https://www.dropbox.com/s/iylc0q8c7ednwtq/OntoSewage_20230420.nt?dl=0

For the issue of multiple geo coordinates and elevations, they refer to two points "Schachtmittelpunkt und Haltungsmittelpunkt". I have removed the Haltungsmittelpunkt and keep the Schachtmittelpunkt which has a lower altitude for the manhole location. 

Meanwhile, I will shorten the running time of the agent with the suggestions from you. Thank you

markushofmeister (changes requested):
See my comment to your last reply

xuehansong:
@markushofmeister Hi, please find latest updated instantiation triples data at 

https://www.dropbox.com/s/9lc7m8zkj3dmtzf/OntoSewage_20230426.nt?dl=0

I have re-written the entire code so that now the instantiation time has been improved from 24 hours to within 1 hour now. Earlier I have 

1) done SPARQL update for each xml tag, this is the reason why it takes so long, now I do once for all and the time has been reduced tremendously; 

2) used if-condition to not instantiate the empty data because earlier I denoted "None" string for all the non-existing data, did the SPARQL update, then removed all the triples containing "None".

For point 5, I have already used subclass definition in the ontology. 

Please advise for any further amendment if you spot any other issues. Thank you.

markushofmeister (changes requested):
Well done wrt shortening the instantiation time.

However, the instantiation seems still incorrect. The below screenshot, for example, shows that the instantiation of `location` instances is mixed up with `elevation` instances. Please fix! And also check for potentially similar issues at other "nested" object properties

image

xuehansong:
@markushofmeister Hi, thanks for the feedback. I have checked and updated the code once more and the updated instantiation data is available at

https://www.dropbox.com/s/vq7ct9ovprbd24y/OntoSewage_20230427.nt?dl=0

The mixed up of the location and elevation instance has also been fixed as shown below

image

Thank you

markushofmeister (approved):
I had a look at the latest instantiated triples and the previously identified issues seem solved. @sm453 Can you please have a final look?

sm453:
Are these blank nodes expected?
image

xuehansong:
@sm453 "hasUpstreamConnector" and "hasDownstreamConnector" are data type property, so connect to these Strings which are given by the raw data xml tags

sm453:
> @sm453 "hasUpstreamConnector" and "hasDownstreamConnector" are data type property, so connect to these Strings which are given by the raw data xml tags

These two properties are indeed datatype properties, with range `xsd:string`. That means the object of the statements in the above screenshot should be string literals, not blank nodes. Can you please fix this?

xuehansong:
> > @sm453 "hasUpstreamConnector" and "hasDownstreamConnector" are data type property, so connect to these Strings which are given by the raw data xml tags
> 
> These two properties are indeed datatype properties, with range `xsd:string`. That means the object of the statements in the above screenshot should be string literals, not blank nodes. Can you please fix this?

Hi, I am confused here, as I thought blank notes are "URI or literal is not given". 

For example, "319M33" is already the literal. Or are you saying I should put it like  319M33^^xsd:string?

sm453 (approved):

PR #524, opened 2022-12-12
Add a username and password to blazegraph container

gpeb2:
I've updated the Blazegraph image so that it requires authentication if a username and password are provided by setting the environment variables BLAZEGRAPH_PASSWORD_FILE and BLAZEGRAPH_USER and a secret with the name blazegraph_password at runtime. This will be the default in the stack with the default user name being set to "bg_user" and only the secrete file needing to be added to the stack-manager's ```inputs/config``` directory.


Deploy/stacks/db/docker-compose.deploy.yml
```diff
@@ -16,16 +16,18 @@ services:
...
   # Blazegraph
   blazegraph:
-    image: docker.cmclinnovations.com/blazegraph:1.0.0-SNAPSHOT
+    image: docker.cmclinnovations.com/blazegraph:1.1.0-510-SNAPSHOT
     container_name: "blazegraph${CONTAINER_NAME_SUFFIX}"
     environment:
-      BLAZEGRAPH_PASSWORD_FILE: /run/secrets/blazegraph_password
+    # Set these environment variables for BASIC authentication
+#      BLAZEGRAPH_PASSWORD_FILE: /run/secrets/blazegraph_password
+#      BLAZEGRAPH_USER: bg_user
```
jb2197, line 24: Maybe add a comment to indicate uncomment these two lines if one wants to set authentication.

Deploy/stacks/db/docker-compose.deploy.yml
```diff
@@ -16,16 +16,18 @@ services:
...
       LOG4J_FORMAT_MSG_NO_LOOKUPS: "true"
     pull_policy: if_not_present
     restart: unless-stopped
     # Add a secret to set the password for BASIC authentication
-    secrets:
-      - blazegraph_password
+#    secrets:
+#      - blazegraph_password
```
jb2197, line 30: Maybe add a comment to indicate uncomment these two lines if one wants to set authentication.

Deploy/stacks/db/blazegraph/docker-entrypoint.sh
```diff
@@ -4,13 +4,19 @@
...
 # Update the contents of WEB-INF in the blazegraph war file
-pushd /tmp/blazegraph
+pushd /tmp/blazegraph || exit
+
+if [[ -n $BLAZEGRAPH_USER && -n $BLAZEGRAPH_PASSWORD_FILE && -f $BLAZEGRAPH_PASSWORD_FILE ]]; then
+    # Set username and password in the tomcat-users.xml using docker secret values
+    sed -i -e "s/BLAZEGRAPH_USER/$BLAZEGRAPH_USER/" -e "s/BLAZEGRAPH_PASSWORD/$(cat $BLAZEGRAPH_PASSWORD_FILE)/" /usr/local/tomcat/conf/tomcat-users.xml
+else
+    # Remove "security-constraint" related blocks
+    sed -i '6,24d' WEB-INF/web.xml
```
jb2197, line 14: This is nitpicking, but maybe removing content matching tags of the "security-constraint" related blocks is more robust against line number changes?

jb2197 (approved):
I have tried to spin up image `docker.cmclinnovations.com/blazegraph:1.1.0-510-SNAPSHOT` with and without authentication set, and both cases are working fine. I think it would be beneficial to add some comments or README so that others are also aware of these changes and how to set up the authentication if desired.

I have not tested this as part of the stack due to the memory constraints of my machine. (But I suppose the codes work fine.)

Also, before merging this branch to the main, could you please release a stable version of blazegraph image and update the version number accordingly? That would be really helpful to many! Thanks in advance!

Deploy/stacks/db/docker-compose.deploy.yml
gpeb2: I've added a basic README file.

Deploy/stacks/db/docker-compose.deploy.yml
gpeb2: I've added a basic README file.

Deploy/stacks/db/blazegraph/docker-entrypoint.sh
gpeb2: Good point, improved in commit e1c577a.

PR #525, opened 2022-12-13
dev-historical-pump-data-instantiation

qhouyee:


sm453:
For the pumps, when generating IRIs, it appears to interpret building numbers as floating point numbers rather than strings, which causes it to trim trailing zeros and thus generate wrong IRIs. In this example, the IRI should end in "2.30" rather than "2.3":
image

sm453:
It also seems to generate two erroneous triples, like this one:
image

sm453:
After having run the agent, it does not appear to close the connection to the PostgreSQL database. I have to log into the adminer front-end and kill the running process manually.

sm453:
I think the `pump` bit of the IRI has been inserted a few extra times ;-):
image

sm453:
I suspect the agent adds one more `pump` every time it's being called...

Agents/HistoricalPumpDataInstantiation/HistoricalPumpDataInstantiationAgent/src/main/java/uk/ac/cam/cares/jps/agent/TimeSeriesClientDecorator.java
```diff
@@ -0,0 +1,173 @@
...
+            // excelReadings.get(header) returns array list, so add get method to get the value
+            List<Class<?>> classes = excelHeaders.stream()
+                    .map(header -> excelReadings.get(header).get(0).getClass())
+                    .collect(Collectors.toList());
+            try (Connection conn = rdbClient.getConnection()) {
+                tsClient.initTimeSeries(iris, classes, this.timeUnit, conn, TimeSeriesClient.Type.STEPWISECUMULATIVE, null, null);
+                LOGGER.debug(String.format("Initialized time series with the following IRIs: %s", String.join(", ", iris)));
+            } finally {
+                // Always close connection
+                rdbClient.getConnection().close();
```
sm453, line 89: Since getConnection will open a new connection, this line is totally pointless. The try-with-resources block above should already close the connection automatically.

Agents/HistoricalPumpDataInstantiation/HistoricalPumpDataInstantiationAgent/src/main/java/uk/ac/cam/cares/jps/agent/TimeSeriesClientDecorator.java
```diff
@@ -0,0 +1,173 @@
...
+     * Updates the database with new readings.
+     *
+     * @param excelReadings Excel readings parsed from the Excel Workbook
+     * @param iriMappings   Mappings between measures' names and their corresponding data IRI.
+     */
+    protected void updateData(Map<String, List<?>> excelReadings, Map<String, String> iriMappings) {
+        TimeSeries<Instant> timeSeries = convertReadingsToTimeSeries(excelReadings, iriMappings);
+        // Update each time series
+        try {
+            tsClient.addTimeSeriesData(timeSeries, rdbClient.getConnection());
```
sm453, line 133: This line is most likely to leave an open connection dangling. There also seem to be similar ones elsewhere.

Agents/HistoricalPumpDataInstantiation/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+FROM maven:3.8.6-eclipse-temurin-11-focal as builder
+
+# Copy all files into root's home, including the source, pom file, ./m2 directory, and credentials files
+ADD /HistoricalPumpDataInstantiationAgent /root/HistoricalPumpDataInstantiationAgent
+ADD /credentials /root/credentials
+ADD /.m2 /root/.m2
+
+# Populate settings templates with credentials, repo name
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 12: "the latter, but not the former"

sm453 (approved):

PR #526, opened 2022-12-13
Stack container grouping

gpeb2:
Minor change that causes the containers to be grouped by stack in GUIs (Docker Desktop, VSCode, etc.).


kokfoongcmcl:
Will test this once this is pushed..

kokfoongcmcl (approved):
Looks to be working, but I cannot spin up the full stack due to the blazegraph bug

gpeb2:
> Looks to be working, but I cannot spin up the full stack due to the blazegraph bug

This was due to an unrelated issue.

PR #527, opened 2022-12-13
Update the stack-manager to automatically mount the docker socket within custom containers

gpeb2:
This will make writing the configuration files for custom containers easier. See the attached issue for details of the problem.

### DIFF
--- a/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
+++ b/Deploy/stacks/dynamic/stack-clients/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.8'
 services:
   stack-client:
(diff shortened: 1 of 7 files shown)


markushofmeister (approved):
Looks good to me.
Does this cause issues with an "additional" mount declaration of this Docker daemon in the .json configuration to spin up an agent, or would that simply be unnecessary/superseded?

gpeb2:
> Looks good to me. Does this cause issues with an "additional" mount declaration of this Docker daemon in the .json configuration to spin up an agent, or would that simply be unnecessary/superseded?

I've configured it so that if the exact same mapping is manually added it doesn't add it again.

PR #533, opened 2022-12-20
Stack client image fix for use with py4jps

gpeb2:
Fixed some issues with the stack-client Docker image and container mainly around the location of the stack-client jar file. This allows the image to be used when trying to add the jar files to py4jps python package.


sm453 (approved):
Looks good to me, and seems to be working in at least one application.

I assume you'll be removing the SNAPSHOT version qualifiers before merging?

gpeb2:
> I assume you'll be removing the SNAPSHOT version qualifiers before merging?

Yes, that is always the intention.

PR #539, opened 2023-01-04
Corrected the version of the Blazegraph image pulled before running the stack-manager

gpeb2:
It is non-trivial to pull Docker images from private repositories from within a container so this is currently done as a pre-process in the stack start.sh script. When updating the Blazegraph image, so that it could easily be run with authentication enabled, I forgot to update the image's version number in the start.sh script. This has lead to the stack-manager code trying to pull the image and failing due to the lack of access to the credentials for the private repository it is hosted on. To fix this I have now updated the version number in the script.
I have also added a brief mention of the ability to add authentication to the Blazegraph container when running it in the stack and the potential compatibility issues with existing agents.

### DIFF
--- a/Deploy/stacks/dynamic/common-scripts/start.sh
+++ b/Deploy/stacks/dynamic/common-scripts/start.sh
@@ -8,8 +8,12 @@ init_swarm
 
 # Pull these images here because the "--with-registry-auth" argument
 # used below only seems to add credentials for Docker Hub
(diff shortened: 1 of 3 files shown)


gpeb2:
@cmclscourtney please can you test this fix by deleting your Blazegraph and GeoServer Docker images and then starting a stack from scratch?

cmclscourtney (approved):
checked and worked as intended

PR #541, opened 2023-01-04
540 ukpn pylon example external data set loop

cmclscourtney:
small fix to pylons example. Names of config files removed that use the external datasets. Typo caused a loop


cmclscourtney:
See #541

gpeb2 (approved):

PR #556, opened 2023-01-19
1478-dev-create-loudness-and-loudness-unit-classes: Add loudness and …

Abyssus1010:
…loudnessUnit concepts to ontouom


sm453:
As far as I understand, sound pressure level is measured in decibels, and it should not be confused with loudness. I know, colloquially, this distinction is not made, but I don't believe that's correct...

sm453:
If you remove dB, then you also have to remove dBFS.

I would also suggest not to define loudness at all. Instead, you may want to define sound pressure level, with dB and dBFS as units.

sm453 (approved):

PR #562, opened 2023-01-20
561 cors issue when trying to connect to a feature info agent from another stack

cmclscourtney:
@sm453 could you test this? 

You should only need to change the manager version to `1.7.1-561-SNAPSHOT` and add the following top level node to the feature info agent config.
`"type": "feature-info-agent",`


sm453 (approved):
Tested, works. At some point in the future, when it becomes relevant, we may want to revisit this from a security point of view, but for now this is fine.

PR #568, opened 2023-01-25
1458 dev metoffice agent improvements

sganguly-cares:


markushofmeister (approved):
Removed SNAPSHOT quantifiers from Docker image names and released them to github registry. Further changes:
* agentlogging now imported from py4jps
* removed unit symbol instantiation for newly instantiated readings
* re-introduced StackClients local building for tests

PR #578, opened 2023-02-01
Dev stackerise access agent

sm453:
Addition of a stack manager input config file for the access agent, together with minimal documentation of how to use.

### DIFF
--- a/JPS_ACCESS_AGENT/README.md
+++ b/JPS_ACCESS_AGENT/README.md
@@ -22,7 +22,8 @@ The purpose of the Blazegraph is to store routing information used by the triple
 Routing information is stored in the default "kb" namespace and the triple store access agent is configured to use this is as the STOREROUTER_ENDPOINT.
 
 The uploading of routing information for the RDB Access Agent is described in the [Uploading Routing Information](#Uploading-routing-information) section below.
-### Spining up the Access Agent dev stack
+
+### Spinning up the Access Agent dev stack
 
 From the command line, in the access-agent-dev-stack directory, run:
 ```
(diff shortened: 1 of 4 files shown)


mehalaga:
In the input config file, the endpoints are defined as:
```
"STOREROUTER_ENDPOINT=http://<STACK NAME>-blazegraph:8080/blazegraph/namespace/storerouter/sparql",
 "RDB_STOREROUTER_ENDPOINT=http://<STACK NAME>-blazegraph:8080/blazegraph/namespace/storerouterrdb/sparql"]
```
but the Access Agent uses the default kb namespace where the triples are uploaded and as for the RDBAccessAgent the ontordbrouter namespace needs to be either manually created or the one on theworldavatar can be used.
Does this ensure the same functionality?

mehalaga (approved):

sm453:
> In the input config file, the endpoints are defined as:
> 
> ```
> "STOREROUTER_ENDPOINT=http://<STACK NAME>-blazegraph:8080/blazegraph/namespace/storerouter/sparql",
> "RDB_STOREROUTER_ENDPOINT=http://<STACK NAME>-blazegraph:8080/blazegraph/namespace/storerouterrdb/sparql"
> ```
> 
> but the Access Agent uses the default kb namespace where the triples are uploaded and as for the RDBAccessAgent the ontordbrouter namespace needs to be either manually created or the one on theworldavatar can be used. Does this ensure a similar functionality?

The kb and ontordbrouter namespaces are defaults, but can be overridden. Also, the namespaces in the above snippet are entirely arbitrary, so can be changed to the defaults if desired. I have tested this (at least for one basic case) successfully.

PR #579, opened 2023-02-02
570 forecasting rdbnamespace per request

markushofmeister:
This PR adds the following functionality to the Forecasting Agent (as of version 1.1.0):
- connection properties (i.e. `namespace`, `database`, `db_url`, ... ) provided in the `docker-compose` files only represent default values
- those default values can be overwritten on a per request basis by respective keys in the HTTP request to the agent

The functionality has been tested successfully locally outside the stack.

### DIFF
--- a/Agents/ForecastingAgent/forecasting/flaskapp/forecastingtasks/routes.py
+++ b/Agents/ForecastingAgent/forecasting/flaskapp/forecastingtasks/routes.py
@@ -19,8 +19,8 @@ from forecasting.errorhandling.exceptions import InvalidInput
 from py4jps import agentlogging
 
 # Initialise logger instance (ensure consistent logger level`)
-logger = agentlogging.get_logger('dev')
-#logger = agentlogging.get_logger('prod')
+#logger = agentlogging.get_logger('dev')
+logger = agentlogging.get_logger('prod')
 
 
 forecastingtasks_bp = Blueprint(
--- a/Agents/ForecastingAgent/forecasting/flaskapp/forecastingtasks/routes.py
+++ b/Agents/ForecastingAgent/forecasting/flaskapp/forecastingtasks/routes.py
@@ -12,6 +12,10 @@ from flask import Blueprint, request, jsonify
 from py4jps import agentlogging
 
 from forecasting.forecasting_agent.agent import forecast
+from forecasting.utils.default_configs import STACK_NAME, NAMESPACE, DATABASE, \
+                                              DB_URL, DB_USER, DB_PASSWORD, \
+                                              QUERY_ENDPOINT, UPDATE_ENDPOINT
+from forecasting.errorhandling.exceptions import InvalidInput
 
 # Initialise logger instance (ensure consistent logger level`)
 logger = agentlogging.get_logger('prod')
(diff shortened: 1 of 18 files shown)


Agents/ForecastingAgent/forecasting/utils/default_configs.py
```diff
@@ -42,45 +48,46 @@ def retrieve_settings():
...
-            raise ValueError('No "DATABASE" value has been provided in environment variables.')
-        if DATABASE != 'postgres':
+            warning_msg = f'No default "DATABASE" value has been provided in environment variables. '
+            warning_msg += f'Ensure that "DATABASE" is provided in HTTP request to the agent to avoid issues.'
+            logger.warning(warning_msg)
+        elif DATABASE != 'postgres':
             logger.warning(f'Provided "DATABASE" name {DATABASE} does not match default database name "postgres".')
         
         # Retrieve settings from Stack Clients
+        # TODO: Review and check what happens/is returned if empty strings are provided
```
markushofmeister, line 90: This part requires testing for stack deployment, i.e. What happens when no default values are provided? In case exceptions are thrown this would require rework

Agents/ForecastingAgent/forecasting/flaskapp/forecastingtasks/routes.py
```diff
@@ -97,3 +122,53 @@ def api_forecast():
...
+                # ... otherwise use default value
+                default = eval(r.upper())
+                if default == '':                    
+                    logger.error(f'No "{r}" value provided in HTTP request despite missing default value.')
+                    raise InvalidInput(f'No "{r}" value provided in HTTP request, despite missing default value.')
+                else:
+                    stack_params[r] = default
+        # Retrieve connection parameters from stack clients (returned as:
+        # DB_URL, DB_USER, DB_PASSWORD, QUERY_ENDPOINT, UPDATE_ENDPOINT)
+        stack_conn_param = retrieve_stack_settings(**stack_params)
```
markushofmeister, line 168: This should also be verified in stack deployment

Agents/ForecastingAgent/stack-manager-input-config/forecasting-agent.json
```diff
@@ -38,7 +38,7 @@
     },
     "endpoints": {
         "rest": {
-            "url": "http://localhost:5000/api/forecastingAgent/",
+            "url": "http://localhost:5005/api/forecastingAgent/",
```
sm453, line 41: Please undo this change. It should stay at 5000. NB This is the internal number, not the external one that you changed in the docker compose files.

Agents/ForecastingAgent/forecasting/utils/default_configs.py
sm453: > This part requires testing for stack deployment, i.e. What happens when no default values are provided? In case exceptions are thrown this would require rework

If no namespace environment variable is defined in the agent config json, then the container fails to start when attempting to do so with the stack manager. The container logs show `ERROR - "NAMESPACE" name is missing in environment variables`. I am happy with this behaviour, and the error message is sufficiently clear.

sm453 (changes requested):
I have successfully managed to produce a time series forecast within a stack, with a database and namespace passed in an http request (that differ from the defaults set by environment variables).

The only change required in my view is to revert back to port 5000 in the agent config json (and only there - not in the docker compose files!).

markushofmeister:
Updated agent behavior:
1) All environment variables in `docker-compose.yml` except `STACK_NAME` are now optional: If provided they serve as default value (to be potentially overwritten by equivalent in HTTP request), otherwise no default is set and key is required in HTTP request (which is optional otherwise)
2) Updated README to 1) reflect that updated behavior and 2) include section about potentially troubleshoot nginx routing issue
3) Aligned internal agent routes to ease stack deployment using `.json` file, i.e. forecasting route and startup route follow same pattern now

Docker images have been pushed to ghcr.io

@sm453: Do we still need the house45 related files in here? Otherwise, we might want to remove them. Just placed them here initially as examples

kokfoongcmcl:
FYI I'm adding the option to specify the time series schema in the time series client, so you might not need to have multiple databases

sm453:
> FYI I'm adding the option to specify the time series schema in the time series client, so you might not need to have multiple databases

We have many entities (agents) that do many things with many different time series. It is not reasonable to assume that all time series are in the same schema/database - and they shouldn't be, for a number of reasons. Therefore, all entities (agents) _must_ offer the functionality to be passed which schema/database to use for a particular purpose.

sm453:
> @sm453: Do we still need the house45 related files in here? Otherwise, we might want to remove them. Just placed them here initially as examples

Just for the record, yes, please. I have found them useful as examples how it can be used.

PR #593, opened 2023-02-07
Dev add timeseries schema

kokfoongcmcl:
By using TimeSeriesClient.setRDBSchema, time series data can be stored/retrieved from the specified schema. Triples specifying the schema is not added, but can be added if it's useful


sm453 (approved):
Looks good, although I have not tested it.

PR #596, opened 2023-02-08
546 update stack tools to work with podman

gpeb2:
Made a range of changes so that the stack-clients, stack-manager and stack-data-uploader codes work when run through Podman rather than Docker. The main changes were finding alternative functionality to that provided by Docker Swarm and putting the containers in pods so that they can connect internally using the container names.


cmclscourtney:
Tested new functionality within credo and works well.
Have tested existing docker examples still work as expected.

cmclscourtney (approved):

cmclscourtney:
Tested that 
1. when using Podman the stack-manager does the right thing if you start a stack. When I manually stopped a container using `podman stop <container name>` and then ran the stack-manager again the stopped container was removed and replaced.
2. in Windows I tried to debug the stack-manager after having already started a stack and it doesn't allow you to pick the previously used port.

PR #605, opened 2023-02-14
dev-ontobuildingstructure

qhouyee:
An ontology developed to represent building structure components such as walls, doors, floors, roofs, columns, slabs, and stairs. It would support OntoBIM by containing all these related real-world information.


JPS_Ontology/ontology/ontobuildingstructure/OntoBuildingStructure.csv
```diff
@@ -0,0 +1,49 @@
...
+hasArea,"Object Property, Asymmetric Property",hasQuantity,IS-A,Roof,RoofArea,,A relationship between the roof of a building and the area it covers.,http://www.theworldavatar.com/kg/ontobuildingstructure/,has area of
+hasOrientation,"Object Property, Asymmetric Property",hasQuantity,IS-A,Roof,http://www.ontology-of-units-of-measure.org/resource/om-2/Angle,,A relationship between a roof and its angle/position relative to the Sun.,http://www.theworldavatar.com/kg/ontobuildingstructure/,has orientation
+hasSlope ,"Object Property, Asymmetric Property",hasQuantity,IS-A,Roof,http://www.ontology-of-units-of-measure.org/resource/om-2/Angle,,A relationship between a roof and its slope angle.,http://www.theworldavatar.com/kg/ontobuildingstructure/,has slope
```
sm453, line 36: `hasOrientation` and `hasSlope` should be properties of `InclinedRoof` rather than `Roof` (because they are meaningless for `FlatRoof`).

JPS_Ontology/ontology/ontobuildingstructure/OntoBuildingStructure.csv
```diff
@@ -0,0 +1,49 @@
...
+isPartOf,"Object Property, Asymmetric Property",,,Landing UNION Railing UNION StairFlight UNION StairStructuralComponent,Stair,only,"A relationship that links a component to their assembly. An example is that railing, landing, stair flights, and stair structural components are part of stair.",http://www.theworldavatar.com/kg/ontobuildingstructure/,is part of 
+consistsOf,"Object Property, Asymmetric Property",,,Wall,Door UNION Window,only,"A relationship that denotes the subcomposition of an assembly with their individual components. An example is that stairs consists of railing, landing, stair flights, and stair structural components.",http://www.theworldavatar.com/kg/ontobuildingstructure/,consists of
+isPartOf,"Object Property, Asymmetric Property",,,Door UNION Window,Wall,only,"A relationship that links a component to their assembly. An example is that railing, landing, stair flights, and stair structural components are part of stair.",http://www.theworldavatar.com/kg/ontobuildingstructure/,is part of
```
sm453, line 41: As usual, I would recommend to keep either `consistsOf` or `isPartOf`, but not both (I would keep the former). Otherwise, writing queries for this will be quite awkward or has to rely on assumptions that may or may not be true.

sm453 (changes requested):

JPS_Ontology/ontology/ontobuildingstructure/OntoBuildingStructure.csv
qhouyee: The dataset has these values attached to the Flat Roof as well. However, after investigation, these values for all flat roofs are -1 and 0 respectively. As this data is redundant, I have attach the properties to InclinedRoof instead where there are differences.

JPS_Ontology/ontology/ontobuildingstructure/OntoBuildingStructure.csv
qhouyee: Removed isPartOf as per recommendation

sm453 (approved):

PR #607, opened 2023-02-14
1490-dev-add-unit-tests-for-powersys-batteryretrofit

mingchuantian812:
Methods of `uk.ac.cam.cares.jps.powsys.retrofit.BetteryRetrofit` class that are unit-tested:
- `setLogger()`
- `processRequestParameters()`
- `validateInput()`

Methods that are not unit-tested:
- `retrofitEnergyStorage()`: This method initialise a private GeneralRetrofitAgent' and calls the `addGeneratorsToElectricalNetwork()` which calls class method `QueryBroker.UpdateFile`. This cannot run without a complete setup of the application and the file system and goes beyond the scope of the unit test.

### DIFF
--- a/JPS_POWSYS/pom.xml
+++ b/JPS_POWSYS/pom.xml
@@ -105,6 +105,12 @@
 		    <artifactId>commons-io</artifactId>
 		    <version>2.6</version>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>4.0.0</version>
+			<scope>test</scope>
+		</dependency>
 
 	</dependencies>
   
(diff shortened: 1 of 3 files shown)


sganguly-cares (changes requested):
The commit messages are confusing - the first one says "1490-dev-add-unit-tests-for-powersys-batteryretrofit" and the second one says "1492-dev-add-unit-tests-for-powersys-renewablegeneratorretrofit". Commit messages should start with the branch name, followed by a brief message on what changes have been made.
In this case, "1492-dev-add-unit-tests-for-powersys-renewablegeneratorretrofit" should not show up at all since that is not the branch we are on

A correct commit message would be "1490-dev-add-unit-tests-for-powersys-batteryretrofit: Added test for this class"
Can you please fix this?

PR #622, opened 2023-02-22
dev-building-matching-agent-revert-upgrades: Revert all dependency up…

markushofmeister:
…grades to ensure reliable agent operation (agent did not process requests reliably after upgrading dependencies)

As discussed, this reverts previous dependency upgrades as no reliable agent operation could be achieved in a reasonable amount of time. Potentially to be revisited in the future.


sm453 (approved):
Sad.

PR #625, opened 2023-02-23
Dev flood warning instantiation agent

markushofmeister:
Dockerised agent to instantiate and update flood warnings and alerts from the Environment Agency Real-time flood monitoring API. Stores flood area polygons as well as styling relevant information in PostGIS for visualisation via Geoserver.

The agent is intended to mark up instantiated information to be picked up by the derivation framework. Derivation mark-up function to be included in subsequent pull request to ease review.


sm453 (approved):

markushofmeister:
I added an additional comment about the required JDBC driver in the Dockerfile (necessary to retrieve PostGIS stack client properties) and released both agent images as version 1.0.0

PR #626, opened 2023-02-24
Dev stackerise feature info agent

sm453:
Added a config `json` file to the Feature Info Agent in order to facilitate spinning it up as part of a stack, together with minimal documentation.


mdhillman (approved):
No issues with these changes, thanks for looking into this @sm453.

PR #633, opened 2023-03-03
Hotfix for AERMOD agent (virtual sensor use case)

ChungTingLao:
Changed the code that formats weather data for AERMOD agent.

Changed parsing of WKT literals. Checked that it works for Plymouth and Singapore.


kokfoongcmcl:
I think favicon.ico should not be committed. Before merging to main, remove the SNAPSHOT qualifiers from the versions

kokfoongcmcl (approved):

PR #636, opened 2023-03-04
Dev flood warning derivation markup

markushofmeister:
Added derivation markup to Flood Warning Instantiation agent.

To Do upon approval:
- remove SNAPSHOT from Docker versions
- enable recurring background task


markushofmeister:
Tested in conjunction with Flood Assessment Agent as of PR #637

jb2197 (approved):

PR #637, opened 2023-03-05
Dev update flood assessment agent

markushofmeister:
Updated Flood Assessment Agent to actually assess the number of affected people using PostGIS' geospatial capabilities. Also cleaned up the agent and removed no longer required Derivation MVP aspects.
I have tried the agent on same sample data within the stack and it seems to perform as expected.

Currently, the dockerised integration tests fail as mocking of the `estimate_number_of_affected_people` method does not seem to be successful. Mocking is defined in the `test_example_agent.py` module; however, the method to be mocked belongs to the `FloodAssessmentAgent` class in the `impact_assessment.py` module. I suspect that this approach was already flawed in the previous version, but we did not notice as the mocked value coincided with the previous default return value of the method.
```
  mocker.patch('floodassessment.agent.impact_assessment.FloodAssessmentAgent.estimate_number_of_affected_people',
               return_value=None)
```
Any idea, how this could easiest be fixed?


Agents/FloodAssessmentAgent/setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='floodassessment',
-    version='0.0.1',
+    version='1.1.0',
```
jb2197, line 5: Could you please update to `pyderivationagent==1.4.4` and `py4jps==1.0.34`? This will incorporate the changes of the OntoDerivation namespace and some other bug fixes/improvements.

jb2197: Updating those pinned versions in `requirements.txt` should also work.

jb2197:
> Currently, the dockerised integration tests fail as mocking of the `estimate_number_of_affected_people` method does not seem to be successful. Mocking is defined in the `test_example_agent.py` module; however, the method to be mocked belongs to the `FloodAssessmentAgent` class in the `impact_assessment.py` module. I suspect that this approach was already flawed in the previous version, but we did not notice as the mocked value coincided with the previous default return value of the method.
> 
> ```
>   mocker.patch('floodassessment.agent.impact_assessment.FloodAssessmentAgent.estimate_number_of_affected_people',
>                return_value=None)
> ```
> 
> Any idea, how this could easiest be fixed?

I can confirm the mock is not working as expected. My guess is that the `mocker.path` only affects the python scripts that are accessed by pytest, the agent that actually handles the derivations is running behind gunicorn hence not affected. A workaround - this is not as elegant as one would like, but tested working in WSL - is adding the below line to the `test` image in `Dockerfile` before executing the `CMD` line:
```Dockerfile
RUN sed -i "s|self.estimate_number_of_affected_people(warning_iri)|None|g" ./$AGENTDIR/agent/impact_assessment.py
```

Agents/FloodAssessmentAgent/setup.py
markushofmeister: Addressed by [dcaa7b0](https://github.com/cambridge-cares/TheWorldAvatar/pull/637/commits/dcaa7b0862cedec51f9953b823d43bfb28cb02bf)

markushofmeister:
> > Currently, the dockerised integration tests fail as mocking of the `estimate_number_of_affected_people` method does not seem to be successful. Mocking is defined in the `test_example_agent.py` module; however, the method to be mocked belongs to the `FloodAssessmentAgent` class in the `impact_assessment.py` module. I suspect that this approach was already flawed in the previous version, but we did not notice as the mocked value coincided with the previous default return value of the method.
> > ```
> >   mocker.patch('floodassessment.agent.impact_assessment.FloodAssessmentAgent.estimate_number_of_affected_people',
> >                return_value=None)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any idea, how this could easiest be fixed?
> 
> I can confirm the mock is not working as expected. My guess is that the `mocker.path` only affects the python scripts that are accessed by pytest, the agent that actually handles the derivations is running behind gunicorn hence not affected. A workaround - this is not as elegant as one would like, but tested working in WSL - is adding the below line to the `test` image in `Dockerfile` before executing the `CMD` line:
> 
> ```dockerfile
> RUN sed -i "s|self.estimate_number_of_affected_people(warning_iri)|None|g" ./$AGENTDIR/agent/impact_assessment.py
> ```

Updated and amended documentation in [8098c38](https://github.com/cambridge-cares/TheWorldAvatar/pull/637/commits/8098c388eec6b4b2a545936c7236a4922e184393)

markushofmeister:
Tested successfully for mock API responses with Flood Warning Instantiation Agent as of PR #636

Agents/FloodAssessmentAgent/README.md
```diff
@@ -181,21 +181,25 @@ To run the integration tests, access to the `docker.cmclinnovations.com` registr
...
 - Adjust the import of `env_configs_mock` to `env_configs`
+- Uncomment the `affected_population = self.estimate_number_of_affected_people(warning_iri)` line (around line 172) within `impact_assessment.py` and replace with `affected_population = None`
 - Please also note that some issues with attaching the Debugger have been observed (i.e. `connect ECONNREFUSED 127.0.0.1:5678 `) using VS Code. In case the Debugger does not attach right away, try attaching using the `Python: Reattach and Debug` configuration after the test container has been started. 
 
+The tests have been successfully executed on both Windows and WSL; however, please note the following:
+- There seem to be internal communication issues when running the tests on Digital Ocean, causing an exception along the lines of ```Caused by: HttpException: -1 Unexpected error making the query: GET http://host.docker.internal:27149/blazegraph/namespace/kb/sparql?query=``` (this might be related to permission/priviliges issues)
```
jb2197, line 199: Could you please try adding the below line as suggested in https://github.com/moby/libnetwork/pull/2348#issuecomment-1177610499 and see if it works?
```yml
extra_hosts:
    - "host.docker.internal:host-gateway"
```

markushofmeister: > Could you please try adding the below line as suggested in [moby/libnetwork#2348 (comment)](https://github.com/moby/libnetwork/pull/2348#issuecomment-1177610499) and see if it works?
> 
> ```yaml
> extra_hosts:
>     - "host.docker.internal:host-gateway"
> ```

Nice one! That solves it. I included the changes in commit [61db754](https://github.com/cambridge-cares/TheWorldAvatar/pull/637/commits/61db75474794f31dac0a1200ddc63fa4163b4db5)

jb2197 (approved):

PR #640, opened 2023-03-06
615 stack data uploader error when uploading vector data without a virtual table

gpeb2:
Added null pointer check to fix this issue.


sm453 (approved):
Tested, works.

PR #642, opened 2023-03-06
641 secrets are incorrectly removed when starting multiple stacks

gpeb2:
Added label filter so that only unused secrets from the current stack are removed.


cmclscourtney:
Spun up multiple containers simultaneously and issue with secrets did not arise.

cmclscourtney (approved):

PR #644, opened 2023-03-07
Dev update avgsqmprice agent

markushofmeister:
This updates both `py4jps` and pyderivationagent` to latest versions (in line with other derivation agents required for King's Lynn use case). Tests have been extended and successfully tested on Windows (testing on WSL would be appreciated).

I have not fully dockerised the tests in this agent as I believe it makes sense to keep several different examples for people to draw ideas from and keep things simple:
- Average Square Metre Price Agent: local testing
- Property Value Estimation Agent: local + dockerised testing
- Flood Assessment Agent: only dockerised testing


markushofmeister:
Successfully tested on Windows and Linux (Digital Ocean)

jb2197 (approved):
Tested on WSL and works fine.

markushofmeister:
Released updated Docker images `v1.1.0` to ghcr.io

PR #655, opened 2023-03-10
1475 dev update ontobms

Abyssus1010:


JPS_Ontology/ontology/ontobms/OntoBMS.csv
```diff
@@ -1,88 +1,51 @@
...
+ConvectiveTransportRate,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/MassFlow,IS-A,,,,A Convective Transport Rate in the most general terms refers to the movement of molecules within fluid phases which is the sum of advective and diffusive transport.,https://www.theworldavatar.com/kg/ontobms/,Convective transport rate
+DamperFeedback,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Pressure,IS-A,,,,A signal (pressure based) that that can be used to determine the exact position of a damper.,https://www.theworldavatar.com/kg/ontobms/,Damper feedback
+DamperControl,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Pressure,IS-A,,,,A pressure based measurement used to control the volume of air that flows through your HVAC system by providing measurements to control the damper position.,https://www.theworldavatar.com/kg/ontobms/,Damper control
+SashOpening,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Percentage,IS-A,,,,The amount of opening for a sash. ,https://www.theworldavatar.com/kg/ontobms/,Sash opening
```
sm453, line 35: A sash opening itself _is_ not a percentage - only the extent to which it is opened. So, defining SashOpening as subclass of Percentage is not appropriate, I believe. How about something like SashOpening hasOpenPercentage OpenPercentage, and then OpenPercentage rdfs:subClassOf Percentage? Or perhaps you're looking for FumeHood hasSashOpenPercentage OpenPercentage, with OpenPercentage rdfs:subClassOf Percentage?

sm453 (changes requested):

JPS_Ontology/ontology/ontobms/OntoBMS.csv
```diff
@@ -32,20 +32,22 @@ HeatDuty,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/HeatFlo
...
+https://www.theworldavatar.com/kg/ontodevice/hasQuantity,Object Property,,,http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_function/process.owl#UnitOperation,http://www.ontology-of-units-of-measure.org/resource/om-2/Quantity,,"A relationship between something (it could be a device or space, a setpoint, an operating property etc) and the quantity it represents or has.",https://www.theworldavatar.com/kg/ontobms/,has quantity
+https://www.theworldavatar.com/kg/ontodevice/hasQuantity,Object Property,,,http://www.theworldavatar.com/ontology/ontocape/material/phase_system/phase_system.owl#PhaseSystem,http://www.ontology-of-units-of-measure.org/resource/om-2/Quantity,,"A relationship between something (it could be a device or space, a setpoint, an operating property etc) and the quantity it represents or has.",https://www.theworldavatar.com/kg/ontobms/,has quantity
```
sm453, line 53: You're making a number of statements here about things in OntoDevice. I believe it would be better to move these to OntoDevice, rather to leave them here in OntoBMS...

sm453 (changes requested):

sm453 (approved):

JPS_Ontology/ontology/ontobms/OntoBMS.csv
```diff
@@ -46,3 +46,6 @@ conveys,Object Property,,,https://saref.etsi.org/saref4bldg/FlowMovingDevice,htt
...
 hasHighLevelState,Object Property,,,https://saref.etsi.org/core/MultiLevelState,HighLevelState,,A relationship that describes a high level state of a multi-level state.,https://www.theworldavatar.com/kg/ontobms/,has high level state
+suppliesAirTo,Object Property,,,https://saref.etsi.org/core/HVAC,https://saref.etsi.org/saref4bldg/FlowMovingDevice,,A relationship that describes a HVAC device supplying air to a flow moving device.,https://www.theworldavatar.com/kg/ontobms/,supplies air to
+extractsAirFrom,Object Property,,,https://saref.etsi.org/core/HVAC,https://saref.etsi.org/saref4bldg/FlowMovingDevice,,A relationship that describes a HVAC device removing air from a flow moving device or a flow treatment device.,https://www.theworldavatar.com/kg/ontobms/,extracts air from
+extractsAirFrom,Object Property,,,https://saref.etsi.org/core/HVAC,https://saref.etsi.org/saref4bldg/FlowTreatmentDevice,,A relationship that describes a HVAC device removing air from a flow moving device or a flow treatment device.,https://www.theworldavatar.com/kg/ontobms/,extracts air from
```
sm453, line 51: You will need to define this object property on a single line in your csv file using the UNION keyword for the range. Otherwise you end up with two separate range definitions which is interpreted as _intersection_ (which would be wrong in this case).

Abyssus1010: Oh okay, I will modify that. Thank you

sm453 (approved):

PR #657, opened 2023-03-10
494 provide way to control which custom containers are spun up in a particular stack

gpeb2:
I've added the option to provide a stack config file in the `stack-manager/inputs/config` directory. This file can be used to explicitly include and exclude services from a stack. The custom/user-supplied service configuration files have been moved to the `stack-manager/inputs/config/services` subdirectory. Also several bugfixes and refactorings (mainly the moving of most of the code from the stack-manager project into the stack-clients project) have been merged in from the `582-add-dcat-concepts-in-data-uploader` branch to prevent merge issues in the future.


cmclscourtney (approved):
Tested new functionality and works as intended.

PR #672, opened 2023-03-22
Dev-virtual-sensor-generic-vis-container

ChungTingLao:
Use the general visualisation docker image instead of a local version of the visualisation container specific for the virtual sensor use case.


JPS_VIRTUALSENSOR/README.md
```diff
@@ -4,7 +4,8 @@ Prerequisites
 3) Set openweather API key in ../Agents/WeatherAgent, more details in that folder
 
 Stack needs to be up and running:
-1) Navigate to Deploy/stacks/dynamic/stack-manager, do all required setup as described by the README and execute (in a WSL terminal)
+1) Copy DispersionVis/dispersion-vis.json to Deploy/stacks/dynamic/stack-manager/inputs/config/services.
```
kokfoongcmcl, line 7: this step is a bit "manual", alternatively you can make a "stack manager" folder here with only the docker compose files and spin it up from here

kokfoongcmcl: revert this readme as well?

ChungTingLao: README is now updated.

kokfoongcmcl (approved):

PR #685, opened 2023-03-31
1528-dev-ontoems-domain-extension-for-hasobservationlocation: OntoEMS…

YiKaiTsai:
… domain extension for hasObservationLocation


JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
```diff
@@ -0,0 +1,107 @@
...
+hasCurrentRange,Object Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,Range,exactly 1,,https://www.theworldavatar.com/kg/ontoems/,has current range
+hasCurrentTrend,Object Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,Trend,,Relationship between a water level measure and its current trend,https://www.theworldavatar.com/kg/ontoems/,has current trend
+hasCurrentTrend,Object Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,Trend,exactly 1,,https://www.theworldavatar.com/kg/ontoems/,has current trend
+CloudCover,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Number,IS-A,,,,Fraction of sky obscured by clouds.,https://www.theworldavatar.com/kg/ontoems/,Cloud cover
+CloudCover,Class,http://purl.org/iot/vocab/m3-lite#CloudCover,EQUIVALENT-TO,,,,Fraction of sky obscured by clouds.,https://www.theworldavatar.com/kg/ontoems/,Cloud cover
+hasObservationLocation,Data Property,,,ReportingStation,http://www.bigdata.com/rdf/geospatial/literals/v1#lat-lon,,Description of observation location (i.e. point location in WGS84),https://www.theworldavatar.com/kg/ontoems/,has observation location
```
markushofmeister, line 107: Please correct me if I am wrong, but the idea is to have this geo-representation only until CEA properly works with the stack and also uses PostGIS for geospatial queries. Hence, I would suggest not to change the ontology explicitly and just instantiate this triple on an instance level. 
Looping in Seb: @sm453 What's your perspective on that?

markushofmeister (changes requested):
This commit is hard to review! If you really want to just add a domain extension, you might simply want to import the top-level OntoEMS ontology instead of copying in all the triples. This way would be much cleaner and easier to review relevant parts.

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
```diff
@@ -0,0 +1,107 @@
...
+UVIndex,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Number,IS-A,,,,The strength of the sun's ultraviolet (UV) radiation expressed as index as developed by the World Health Organization.,https://www.theworldavatar.com/kg/ontoems/,UV index
+PrecipitationProbability,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Number,IS-A,,,,The likelihood of occurrence of precipitation.,https://www.theworldavatar.com/kg/ontoems/,Precipitation probability
+PrecipitationProbability,Class,http://semanticscience.org/resource/SIO_000638,IS-A,,,,The likelihood of occurrence of precipitation.,https://www.theworldavatar.com/kg/ontoems/,Precipitation probability
+AirPollutantConcentration,Class,https://www.theworldavatar.com/kg/ontouom/Concentration,IS-A,,,,The concentration of a specific air pollutant as amount of material per unit volume of air (e.g. mass concentration).,https://www.theworldavatar.com/kg/ontoems/,Air pollutant concentration
+http://www.ontology-of-units-of-measure.org/resource/om-2/VolumeFraction,Class,https://www.theworldavatar.com/kg/ontouom/Concentration,IS-A,,,,,https://www.theworldavatar.com/kg/ontoems/,Volume fraction
```
sm453, line 41: Please note that volume fraction is _not_ a concentration (see e.g. https://en.wikipedia.org/wiki/Concentration), so please delete this. Any other statements you'd like to make about volume fraction should be in OM, not here.

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
```diff
@@ -0,0 +1,107 @@
...
+https://www.theworldavatar.com/kg/ontouom/MassOfSubstanceConcentration,Class,https://www.theworldavatar.com/kg/ontouom/Concentration,IS-A,,,,,https://www.theworldavatar.com/kg/ontouom/,Mass concentration
```
sm453, line 42: There are quite a few definitions in this CSV file of things that are defined elsewhere, e.g. OM, OntoTimeSeries, and the environment agency flood monitoring ontology. Please don't repeat other people's definitions.

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
sm453: You are right in that the long-term plan is to move all geospatial functionality to PostGIS. Until then, though, practical minimum-effort solutions are acceptable. Feel free to decide if this makes your life easier in the short term, but please do consider how much work it will be to migrate to PostGIS when the time comes...

sm453 (changes requested):

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
markushofmeister: > There are quite a few definitions in this CSV file of things that are defined elsewhere, e.g. OM, OntoTimeSeries, and the environment agency flood monitoring ontology. Please don't repeat other people's definitions.

Isn't this what the "Defined By" column is for in the .csv template? So, one can keep all relevant concepts and relationships together (mainly for overview and clarify) without re-inventing them. That was my understanding when creating the initial OntoEMS version with Feroz

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
YiKaiTsai: Decided that the ontology won't be extended explicitly, and will remove the added OntoEMSExtension.csv and OntoEMSExtension.owl.

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
sm453: While it may well not be relevant here, I'd rather not do things that are obviously wrong, if it can be avoided... Shame that OM doesn't get it right, either. I'm not suggesting that we fix problems in OM that we haven't created ourselves, but at least we shouldn't add more things that are wrong or make it worse...

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
sm453: I believe repeating other people's definitions is a bad idea as a matter of principle. If the original definitions change, then your 'repetition' becomes out-of-sync, i.e. a contradiction to the (revised) original. For that reason, I'm convinced it is best to avoid this.

JPS_Ontology/ontology/ontoems/OntoEMSExtension.csv
sm453: Looks like this pull request is empty, then. If nothing further is expected, here, then you can simply close it.

PR #686, opened 2023-03-31
Dev stack manager for virtual sensor

ChungTingLao:
Use stack manager to spin up all containers required for the virtual sensor use case.


kok-foong:
Missing instruction to create the geoserver_password and ontop_password files, also not sure about using docker.cmclinnovations.com moving forwards
It is working, my main concern is that it is not convenient to change the image name/version, I doubt the config files can take in environment variables

gpeb2 (approved):
That seems to look ok.

PR #692, opened 2023-04-06
dev-ifcowl-converter-agent

qhouyee:
This agent is intended to support the Ifc2OntoBIM agent to generate the IfcOwl Aboxes from the IFC model.

### DIFF
--- a/Agents/IfcOwlConverterAgent/Dockerfile
+++ b/Agents/IfcOwlConverterAgent/Dockerfile
@@ -8,7 +8,7 @@ COPY /.m2 /root/.m2
 
 # Populate settings templates with credentials, repo name
 WORKDIR /root/.m2
-# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain /, but not |
 RUN sed -i "s|MASTER_PASSWORD|$(mvn --encrypt-master-password master_password)|" settings-security.xml &&  \
     sed -i "s|REPO_USERNAME|$(cat ../credentials/repo_username.txt)|;s|REPO_PASSWORD|$(cat ../credentials/repo_password.txt|xargs mvn --encrypt-password)|" settings.xml
 
--- a/Agents/IfcOwlConverterAgent/Dockerfile
+++ b/Agents/IfcOwlConverterAgent/Dockerfile
@@ -1,0 +1,54 @@
+# First stage: build war file for main agent
+#==================================================================================================
+FROM maven:3.8.6-eclipse-temurin-11-focal as builder
+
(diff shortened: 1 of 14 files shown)


Agents/IfcOwlConverterAgent/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+#==================================================================================================
+FROM maven:3.8.6-eclipse-temurin-11-focal as builder
+
+# Copy Maven settings templates and credentials
+COPY /credentials /root/credentials
+COPY /.m2 /root/.m2
+
+# Populate settings templates with credentials, repo name
+WORKDIR /root/.m2
+# (Note that | rather than / is used as the sed delimiter, since encrypted passwords can contain the former, but not the latter
```
sm453, line 11: "latter, but not the former"

Agents/IfcOwlConverterAgent/IfcOwlConverter/pom.xml
```diff
@@ -0,0 +1,177 @@
...
+    <groupId>uk.ac.cam.cares.jps</groupId>
+    <artifactId>ifcowlconverter</artifactId>
+    <version>1.0.0</version>
+    <packaging>war</packaging>
+
+    <!--  Project properties  -->
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.release>11</maven.compiler.release>
+        <jps.base.version>1.21.1</jps.base.version>
```
sm453, line 16: I guess something more recent is not possible due to dependency incompatibility?

sm453 (changes requested):

Agents/IfcOwlConverterAgent/Dockerfile
qhouyee: Made the changes, and did so for the other PRs.

Agents/IfcOwlConverterAgent/IfcOwlConverter/pom.xml
qhouyee: Could I clarify if you mean for the JPS-base lib dependency? If so, updating the jps base lib to the latest version is possible. The dependency incompatibility is only for some Apache Jena packages. This have been resolved by explicitly stating the required 4.1.0 version within this project pom file.

sm453: If it doesn't break your code, please use the most recent base lib version possible.

qhouyee: Alright, I have updated the base lib version, and tested that it works.

sm453 (approved):

jlai2021 (approved):

PR #696, opened 2023-04-11
fix-dtvf-sdf-icons

mdhillman:
This PR merges in a fix that ensures icons with "-sdf" in their file name will be registered with Mapbox as SDF icons. This enables the ability to change icon color dynamically via Mapbox style expressions.

This change has been added to the DTVF currently hosted on CMCL's servers (v3.3.4) to enable testing.


markushofmeister (approved):
Tested for King's Lynn and works as expected. Please also see PR #699

PR #700, opened 2023-04-13
1533-dev-add-pvt-tube-and-plate-collectors-to-ontoubemmp: add pvt tub…

YiKaiTsai:
…e and plate collectors to ontoubemmp


JPS_Ontology/ontology/ontoubemmp/ontoubemmp.csv
```diff
@@ -103,7 +103,23 @@ NorthWallThermalTubeCollectors,Class,WallThermalTubeCollectors,IS-A,,,,Wall ther
...
 hasCalculationBasis,Data Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,string,,A relationship between a measure and a string containing information on how the measure was calculated,https://www.theworldavatar.com/kg/ontoubemmp/,
+PVTPlateCollectors,Class,PVTCollectors,IS-A,,,,Photovoltaic thermal hybrid solar collector with thermal plate collector,https://www.theworldavatar.com/kg/ontoubemmp/,
+PVTPlateCollectorsArea,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Area,IS-A,,,,The total area of photovoltaic thermal collectors with thermal plate collectors,https://www.theworldavatar.com/kg/ontoubemmp/,
+RoofPVTPlateCollectors,Class,PVTPlateCollectors,IS-A,,,,PVT Plate Collectors with roof type,https://www.theworldavatar.com/kg/ontoubemmp/,
+WallPVTPlateCollectors,Class,PVTPlateCollectors,IS-A,,,,PVT Plate Collectors with wall type,https://www.theworldavatar.com/kg/ontoubemmp/,
+EastWallPVTPlateCollectors,Class,WallPVTPlateCollectors,IS-A,,,,Wall PVT Plate Collectors with east orientation,https://www.theworldavatar.com/kg/ontoubemmp/,
```
sm453, line 114: I'm not sure it is a good idea to subclass collectors for each (north, east, south, west) orientation. For example, what are you going to do with a surface that faces exactly south-east? It would be much more elegant to associate with the collector class an orientation literal (angle), like they do in the Solarkataster data. And then you can categorise, if you must, as a post-processing step (e.g. if 45 deg < angle <= 135 deg then it's east).

sm453 (changes requested):
There are quite a few definitions in the CSV file of things that are defined elsewhere, e.g. `Measure`, `Dimension`, `Unit`, `Quantity`, `Area`, `Energy`, `Volume`, as well as a number of properties belong entirely to OM. Please do not repeat other people's definitions.

Similarly, there are quite a few classes defined in the `oema` and `saref` namespaces. If these definitions are redundant, please remove them. Otherwise the classes should be moved to a different namespace.

JPS_Ontology/ontology/ontoubemmp/ontoubemmp.csv
YiKaiTsai: I agree. As far as I know, this ontology was first created a while ago specifically for the CEA agent. The agent calls CEA to calculate the collector outputs, and CEA does not give orientation angle of the collectors, only their general orientation (North, South, East, West). Might still need a definition for collectors with only general orientation.

Not entirely sure how to go about this. I assume we will have a predicate such as "hasOrientationAngle" with "SolarEnergyGenerator" as the domain and "om:Angle" as the range. And classes or instances for the general orientation (North, South, East, West) that are defined based on the orientation angle.

sm453: If angles are not available, one could have a predicate like `hasGeneralOrientation`, with values of north, south, east, west. Unfortunately, the TBoxGenerator does not support enumerations (see issue #710), so there is no way to do this properly at the moment. In light of that, I guess using subclasses is not so bad - in my view less bad than e.g. using a string datatype property.

sm453 (approved):
Approved with the understanding that the issues regarding `saref` and `oema` definitions will be resolved elsewhere.

PR #709, opened 2023-04-19
1539 hotfix jpshttpservlettest class

csl37:


nagarajankarthik (approved):
Updated base library so that disableScenario test in JpsHttpServletTest class of JPS_BASE_LIB can run successfully.

PR #712, opened 2023-04-19
Dev modify domain range restriction generation

msff2:


sm453 (approved):
Tested for OntoBIM - works.

PR #720, opened 2023-04-26
212-dev-tbox-generator-command-line

jb2197:
This PR enables developers to call TBoxGenerator from the command line without needing to use the prompt window.


sm453 (changes requested):
Could you please add a few lines to the readme file in the same folder, explaining that this functionality exists and how to use it?

sm453 (approved):

msff2 (approved):
I tested the TBox Generator with and without an argument. I found it working as expected in both cases.

PR #721, opened 2023-04-27
1544-dev-add-dc-and-ac-electrical-power-classes-into-ontopowsys: Add …

Abyssus1010:
…electricalPower-AC and electricalpower-DC classes


sm453 (approved):
Please don't forget to remove the affected concepts from OntoBMS.

PR #733, opened 2023-05-05
Allow Superset to use Keycloak for user authentication

gpeb2:
In the CReDo project they are using Keycloak for authentication and authorisation. They have tried to do this from outside the stack and this works for most containers but the Superset webpage fails to load correctly. One way to solve this might be to move the authentication into the Superset contianer itself.


gpeb2:
Closing as we will now use Grafana rather than Superset as it is much more compatible with reverse-proxies, etc.

PR #734, opened 2023-05-05
Various fixes to allow the large amounts of data found in the CReDo project to be queriable

gpeb2:
The latest version of Ontop (5+) includes much more efficient handling of "VALUES". This means that both explicit uses of "VALUES" and there implicit use, when they are generated by another subquery, are considerably faster and support a much larger number of entries.


gpeb2:
The issue that this PR was addressing have been fixed on other branches.

PR #735, opened 2023-05-05
Fix pulling images when starting up the stack

gpeb2:
Private repos require credentials to be specified and there is a bug in the Docker API with checking if an image exists.


gpeb2:
The bug mentioned in the description should be fixed by PR #1091.
Work still needs to be done to be able to handle credentials though.

cmclscourtney (approved):
Tested and works well

sm453 (approved):

Issue #523, opened 2022-12-12
Relative paths when specifying bind mounts


gpeb2:
Allow the use of relative paths when specifying the source of bind mounts in the container config files. The Docker API only supports absolute paths but the stack-manager should be able to generate these from relative paths.

PR #742, opened 2023-05-11
Mounting data into agents in stack

gpeb2:
This pull request adds the following:

- The ability to specify in a stack config file that the content of a specified directory should be automatically copied into a specific volume when the stack is started.
- The ability to specify relative paths for bind mounts in the service config files, they are relative to the `stack-manager/inputs/data` directory for consistency with the auto-populated volumes (properly addressing issue #523).
- Built-in, but optional, service configurations for the digital twin visualisation framework (dtvf) base container and feature information agent container. The extra configuration files these container require are mounted using the auto-populated volume mechanism by default but can use relative bind mounts by providing an overriding service config file.
- Explanations of all of the above in the stack-manager README file.

Other minor changes include:

- Removing an unused argument from the `AbstractService` class and its subclasses.
- Fixed an issue with starting containers that run to completion.
- Made it clearer which services are started by default, including moving the list of default services into a separate file.


sm453:
Sorry, statically copying a folder into a volume at spin-up time isn't at all the same as using a bind mount, which allows convenient editing of a folder on the host system without having to restart the container, in other words live interaction with the container. A noteworthy example of this is editing the queries for the Feature Info Agent (which would be an infinite pain if one had to restart every time).

I'm not saying the feature of copying a folder at spin-up time isn't useful. It very much is, but perhaps reword the narrative? And I don't think this should close #523.

gpeb2:
It's not instantaneous but you just need to re-run the stack-manager without stopping any containers, which is very fast. Other than that, the only difference is that this method doesn't have a trivial way to remove files.

gpeb2:
> Sorry, statically copying a folder into a volume at spin-up time isn't at all the same as using a bind mount, which allows convenient editing of a folder on the host system without having to restart the container, in other words live interaction with the container. A noteworthy example of this is editing the queries for the Feature Info Agent (which would be an infinite pain if one had to restart every time).
> 
> I'm not saying the feature of copying a folder at spin-up time isn't useful. It very much is, but perhaps reword the narrative? And I don't think this should close #523.

I have now explicitly added support for relative bind mounts.

cmclscourtney (approved):
Tested existing and new functionality and worked as intended

markushofmeister (commented):
Unfortunately, I don't have the time to properly test it; however, I had a look through the updated README. Most of it is quite clear, only two suggestions:
1) It might be worth going through it once more to iron out a few remaining discrepancies, such as missing words/links (e.g. [here](https://github.com/cambridge-cares/TheWorldAvatar/blob/94f871004f9bc5e37fde8a7f222f165968d0dcbd/Deploy/stacks/dynamic/stack-manager/README.md?plain=1#LL238C79-L238C79)), to improve clarity
2) As the README is quite lengthy, I believe it would be very helpful to have a commented stack configuration file, which consolidates all available settings, such as exclude default services, include custom services, adding volumes, bind mounts, ...

gpeb2:
> Unfortunately, I don't have the time to properly test it; however, I had a look through the updated README. Most of it is quite clear, only two suggestions:
> 
> 1. It might be worth going through it once more to iron out a few remaining discrepancies, such as missing words/links (e.g. [here](https://github.com/cambridge-cares/TheWorldAvatar/blob/94f871004f9bc5e37fde8a7f222f165968d0dcbd/Deploy/stacks/dynamic/stack-manager/README.md?plain=1#LL238C79-L238C79)), to improve clarity
> 2. As the README is quite lengthy, I believe it would be very helpful to have a commented stack configuration file, which consolidates all available settings, such as exclude default services, include custom services, adding volumes, bind mounts, ...

I've addressed point 1 in commit 7f4d03e.
I think point 2. would be best addressed as part of issue #745.

PR #743, opened 2023-05-11
Example visualization Mapbox API credentials from Docker secret files

gpeb2:
In the DTVF examples, the Mapbox API credentials are now read from Docker secret files that are specified in a built-in but optional service config file.
The user's Mapbox username should be put in a file with the path `stack-manager/inputs/secrets/mapbox_username`
and the user's Mapbox API key should in a file with the path `stack-manager/inputs/secrets/mapbox_api_key`. The stack-manager will automatically load the content of those files into Docker secrets when it is run.


gpeb2:
There will need to be some cross-referencing between the DTVF and stack-manager documentation so I'd like to make sure that is correct before this gets merged.

mdhillman:
This PR is getting old and running the risk of becoming stale. I will update the adjusted DTVF to bring in recent changes from main, then talk to @gpeb2 about getting this merged within a week (hopefully).

cmclscourtney (commented):
Added necessary changes. Works as intended. 

Version in `web/digital-twin-vis-framework/library/version` needs to be consistent.

cmclscourtney (approved):

PR #749, opened 2023-05-16
fix-dtvf-searching

mdhillman:
Features:
* _No new features_

Bug fixes:
* Updated the search handling to better handle using numerical search tokens within string feature properties.
* Updated the search handling to better revert to original data filters (if any have been set).

Version for this fix has been set to 3.3.5


web/digital-twin-vis-framework/example-mapbox-vis/webspace/index.html
```diff
@@ -146,8 +146,8 @@
 
             // ===== CUSTOMISABLE =====
             // Enter Mapbox account name and API key here!
-			MapHandler.MAP_USER = "";
-            MapHandler.MAP_API = "";
+			MapHandler.MAP_USER = "cmclinnovations";
```
cmclscourtney, line 149: committed credentials here, was this intentional?

cmclscourtney (approved):
Tested and searching with numeric values now working as intended

PR #744, opened 2023-05-11
dev-dtvf-cesium-styling

mdhillman:
This PR brings in a new version of the DTVF with the below changes:

- Cesium styling expressions can now be used in `data.json` file
  - Ths is detailed in the "Layers" section of the README in the example Cesium visualisation
  - Example Cesium visualisation has been updated with an example style expression
- Fixed issue with removing highlights on selected Cesium features

This PR has been marked as a draft whilst other DTVF PRs take precedence (#741 and #517)

PR #763, opened 2023-05-25
dev-dtvf-config-url

mdhillman:
Brings in the ability to pass in a custom URL for the visualisation's data configuration file, along with the ability to call this functionality multiple times.

This will allow more experienced developers to add custom controls that can be used to build/determine the location of the config file to use. Once controls are added, simply use the code snippet below (also seen in the Mapbox example visualisation) to reload the visualisation with a new config file.

```javascript
// Load the data definitions
let configURL = "./data.json"
manager.loadDefinitionsFromURL(configURL).then(function() {

    // Plot the data
    manager.plotData();
});
```

Note that this PR is a draft as the version of the DTVF has been set to 3.6.0. It will remain a draft until PR #744 has been merged.


kok-foong (approved):
I won't be testing this immediately, conceptually it looks ok. As long as it does not break existing functionality you can merge this in

mdhillman:
I've tested this locally with both the Mapbox and Cesium examples. It works as designed and does not effect existing functionality. As the code changes are relatively small, I'm merging this in now rather than letting it go stale.

PR #772, opened 2023-06-05
Add more error handling to stack.sh start script

gpeb2:
dev-stack-start-error-handling: Added in code to check that docker containers started using ```./stack.sh start``` actual start and displays the error message if there is one.


gpeb2 (commented):
Added in code to check that docker containers started using ```./stack.sh start``` actual start and displays the error message if there is one.

cmclscourtney (approved):

PR #775, opened 2023-06-08
1461 create an aermod agent that retrieves building related information from the knowledge graph

kok-foong:
The existing Plymouth ship case should be functional as before, main thing for you to test.

Main things added, mainly FYI:
1) Optional "citiesnamespace" parameter in InitialiseSimulation, this is a new derivation input. If this input is present, AermodAgent will query buildings data from the given namespace
2) There are now two kinds of point sources, namely static point source, and ships that are treated specially (same as before), refer to OntoDispersion for more information on the class hierarchy. Static point sources need to have an OntoCityGML object associated with it
3) AermodAgent queries buildings that are near the given point sources to simulate downwash
4) No longer pushing images to docker.cmclinnovations.com for 2 reasons, i) I do not have the permission, ii) pulling them is extremely slow outside CMCL.

If you wish to know more contact me.


ChungTingLao (approved):
Checked that it works with default shipping case and chemical plant case.

PR #781, opened 2023-06-09
1577-dev-add-heating-and-cooling-consumption-to-ontoubemmp

YiKaiTsai:


sm453 (approved):
In principle, it would of course be preferable to use mark-up to capture over what time interval a consumption is measured, rather than use subclasses, but I guess that's probably more trouble than it's worth at this point. So, I think it's ok to leave it for now.

(However, this would also solve the problem that currently we've got `AnnualHeatingConsumption subClassOf HeatingConsumption subClassOf ThermalConsumption` and `AnnualThermalConsumption subClassOf ThermalConsumption` but not `AnnualHeatingConsumption subClassOf AnnualThermalConsumption`!)

Issue #595, opened 2023-02-08
Update GeoServer Docker image


gpeb2:
There is a newer, more official Docker image for GeoServer, which also has some useful configuration features built in.
The images are not on DockerHub but are in the official OSGeo Nexus repository.
The GitHub repository is here: https://github.com/geoserver/docker
The Docker image tags are listed here: https://docker.osgeo.org/#browse/browse:geoserver-docker:v2%2Fgeoserver%2Ftags

As well as it having the latest version of GeoServer it also has environment variables for disabling the loading of the "demo" data and for configuring CORS.

PR #789, opened 2023-06-13
Fix -t_srs option for rasters in stack-data-uploader

gpeb2:
The file format we are using to store rasters has its own option for re-projection. Using this is much simpler than adding the ability to use the ```gdalwarp``` tool although the latter is generally faster.


gpeb2:
Commit f047a13 fixes issue #595.

cmclscourtney (approved):
Tested, geoserver data directory needed to be changed. Committed these changes and now works as intended

Issue #652, opened 2023-03-09
Derivation Framework: Allow missing derivation outputs


markushofmeister:
Currently the Derivation Framework expects to always create the defined output concepts. In case no outputs are created (and hence nothing gets instantiated) by a derivation agent, the framework fails to update the `Finished` status and keeps picking up the derivation to re-try. Those re-processing attempts seem to run until either the derivation is processed (which might be impossible in specific cases) or the next periodical trigger is reached.

I believe there are valid cases where no outputs are generated by a derivation agent. Hence, we need to have a robust solution to handle such cases (both for automatic re-calculation attempts as well as when requested by a downstream derivation).

PR #790, opened 2023-06-13
216 dev derivation no outputs

jb2197:
This PR supports situations where no output is produced from updating an a-/sync derivation. This should close https://github.com/cambridge-cares/TheWorldAvatar/issues/652.


sm453 (approved):
Looks ok, although I have not tested it.

PR #792, opened 2023-06-16
dev-aermod-agent-pollutant: Marked up pollutant ID corresponding to a particular dispersion output.

nagarajankarthik:


nagarajankarthik:
The GeoServer part is still not working correctly. I am investigating the issue.

kok-foong:
I haven't looked into detail, but I don't think pollutant ID needs to be specified as a variable, the AermodAgent should just automatically run the dispersion calculations for all pollutants automatically. For now you can just hardcode 1 pollutant, i.e. SO2

nagarajankarthik:
The polygons should be visible in the GeoServer layer in the commit 6e5999a.

JPS_VIRTUALSENSOR/AermodAgent/src/main/java/com/cmclinnovations/aermod/AermodAgent.java
```diff
@@ -222,10 +218,30 @@ public void processRequestParameters(DerivationInputs derivationInputs, Derivati
...
+            JSONArray features = geoJSON.getJSONArray("features");
+            for (int j = 0; j < features.length(); j++) {
+                JSONObject feature = features.getJSONObject(i);
+                JSONObject featureProperties = feature.getJSONObject("properties");
+                featureProperties.put("derivationIRI", derivationInputs.getDerivationIRI());
+                featureProperties.put("pollutantID", EnvConfig.POLLUTANT_ID);
+                featureProperties.put("simulation_time", simulationTime);
+                featureProperties.put("height", height);
+                feature.put("properties", featureProperties);
+                features.put(j, feature);
```
kok-foong, line 238: these two lines might be redundant

JPS_VIRTUALSENSOR/AermodAgent/src/main/java/com/cmclinnovations/aermod/AermodAgent.java
```diff
@@ -222,10 +218,30 @@ public void processRequestParameters(DerivationInputs derivationInputs, Derivati
...
+            }
+
+            geoJSON.put("features", features);
```
kok-foong, line 241: this too

JPS_VIRTUALSENSOR/AermodAgent/src/main/java/com/cmclinnovations/aermod/AermodAgent.java
```diff
@@ -239,10 +255,43 @@ public void processRequestParameters(DerivationInputs derivationInputs, Derivati
...
         geoServerVectorSettings.setDefaultStyle("dispersion_style");
-        for (int i = 0; i < dispLayerNames.size(); i++) {
-            geoServerClient.createPostGISLayer(EnvConfig.GEOSERVER_WORKSPACE, EnvConfig.DATABASE, dispLayerNames.get(i),
-                    geoServerVectorSettings);
+        // Set GeoServer layer names. Include simulationDirectory in the name to prevent
+        // name collision.
+        List<String> dispLayerNames = new ArrayList<>();
+        for (int i = 0; i < receptorHeights.size(); i++) {
+            double height = receptorHeights.get(i);
+            int receptorHeightInt = (int) Math.round(height);
```
kok-foong, line 263: what if we have heights like 1.2 & 1.4 and both round to 1?

nagarajankarthik: Not sure if it is possible to have decimal points in the GeoServer layer name. Maybe I could do something like 

String receptorHeight = String.valueOf(height).replace(".","_dp_");

JPS_VIRTUALSENSOR/AermodAgent/src/main/java/com/cmclinnovations/aermod/AermodAgent.java
nagarajankarthik: I believe this line creates a new object:

JSONObject featureProperties = feature.getJSONObject("properties");

So if I modify featureProperties, don't I need to update the original object in feature accordingly?

kok-foong: I believe the methods modify the objects directly, rather than creating a new object in memory

nagarajankarthik:
> I haven't looked into detail, but I don't think pollutant ID needs to be specified as a variable, the AermodAgent should just automatically run the dispersion calculations for all pollutants automatically. For now you can just hardcode 1 pollutant, i.e. SO2

The user might not have instantiated emissions data for some pollutants especially if the data was not available.

nagarajankarthik:
I added a pollutants loop that checks the pollutants for which there is emissions data and only runs AERMOD for those pollutants.

JPS_VIRTUALSENSOR/AermodAgent/src/main/java/com/cmclinnovations/aermod/Aermod.java
```diff
@@ -526,8 +528,13 @@ int runAermet(Polygon scope) {
         }
     }
 
-    int createPointsFile(List<PointSource> pointSources, int simulationSrid) {
+    int createPointsFile(List<PointSource> pointSources, int simulationSrid, String pollutantId) {
         StringBuilder sb = new StringBuilder();
+
+        double maxFlowRate = 0.0;
+        double eps = 1.0e-20;
```
kok-foong, line 535: Is there a reason for the choice of this number? And have you checked that it is actually using 1e-20? In my debugger it just shows up as zero

JPS_VIRTUALSENSOR/AermodAgent/src/main/java/com/cmclinnovations/aermod/AermodAgent.java
```diff
@@ -177,82 +181,136 @@ public void processRequestParameters(DerivationInputs derivationInputs, Derivati
...
-            JSONObject geoJSON = aermod.getGeoJSON(EnvConfig.PYTHON_SERVICE_URL, outputFileURL, srid, height);
-            String dispLayerName = dispLayerNames.get(i);
-            gdalClient.uploadVectorStringToPostGIS(EnvConfig.DATABASE, dispLayerName, geoJSON.toString(),
-                    new Ogr2OgrOptions(), true);
-        }
+        for (String pollutId : pollutantsList) {
+            // create emissions input
+            if (aermod.createPointsFile(allSources, srid, pollutId) != 0) {
+                LOGGER.error("Failed to create points emissions file for pollutant {} at timestep {}", pollutId,
+                        simulationTime);
```
kok-foong, line 233: I'd not log an "error" for a pollutant skipped due to the maxFlowRate check, it is misleading. Ideally different log messages for different outcomes

PR #795, opened 2023-06-19
1583 dev extend ontodevice and ontobms

Abyssus1010:


JPS_Ontology/ontology/ontodevice/OntoDevice.csv
```diff
@@ -82,4 +90,5 @@ hasZComponent,Object Property,,,GravityVector,http://www.ontology-of-units-of-me
...
 hasYComponent,Object Property,,,MagneticFluxDensityVector,http://www.ontology-of-units-of-measure.org/resource/om-2/MagneticFluxDensity,only,A relationship that defines the component of a vector in y-direction.,https://www.theworldavatar.com/kg/ontodevice/,has Y component
 hasZComponent,Object Property,,,MagneticFluxDensityVector,http://www.ontology-of-units-of-measure.org/resource/om-2/MagneticFluxDensity,only,A relationship that defines the component of a vector in z-direction.,https://www.theworldavatar.com/kg/ontodevice/,has Z component
+hasCumulativeEnergyConsumption,Object Property,hasQuantity,IS-A,http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_function/process.owl#TemperatureChange UNION Device,http://www.ontology-of-units-of-measure.org/resource/om-2/Energy,,A relationship that describes the amount of cumulative energy used or consumed during a process or by a device.,https://www.theworldavatar.com/kg/ontodevice/,has cumulative energy consumption
```
sm453, line 93: `TemperatureChange hasCumulativeEnergyConsumption`? I don't know the context, but this doesn't sound right to me. Should it not always be a device that has an energy consumption? And such a device could produce a temperature change...

Abyssus1010: Hmmm so the context here is that we have a MAU and we have this variable called BTU_kWh which is supposedly the cumulative energy consumption used during the cooling process. The energy consumption here isn't "electrical" and it's actually more of the heat energy used in cooling down the outside air. 

So the idea is to link this energy consumption to the temperature change process that occurs within the MAU but yes, it does sound abit weird now that I look at it.

sm453: This explanation confirms, in my view, that it is the MAU that consumes the energy - it's not the temperature change. Whether this energy is electrical or otherwise is irrelevant here.

sm453: Ok, `ontocape/chemical_process_system/CPS_function/process.owl#TemperatureChange` is meant to capture a process (in the Process Engineering sense) that either heats or cools (i.e. "changes temperature"). It does make sense for that to have an energy consumption (e.g. it could consist of multiple devices that consume energy).

sm453 (approved):

PR #798, opened 2023-06-21
Hotfix forecasting agent

markushofmeister:
Resolved issues with loading GPU-trained transformer model on CPU-only machines:
- updated and pinned version numbers of relevant dependencies
- updated broken links to stored model on Dropbox

### DIFF
--- a/Agents/ForecastingAgent/docker-compose.yml
+++ b/Agents/ForecastingAgent/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   forecasting_agent:
-    image: ghcr.io/cambridge-cares/forecasting_agent:1.1.0
+    image: ghcr.io/cambridge-cares/forecasting_agent:1.1.1
     environment:
       # Required environment variables for both Stack and "standalone" (i.e. outside stack) deployment
       - STACK_NAME=TEST         # to be left blank for "standalone" deployment
--- a/Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
+++ b/Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
@@ -117,6 +117,6 @@ MODEL_MAPPING['TFT_HEAT_SUPPLY'] = {
         'scale_data': True,
     },
     'frequency': MODEL_MAPPING['DEFAULT']['frequency'],
-    'data_length': MODEL_MAPPING['DEFAULT']['data_length'],
+    'data_length': 24 * 365 ,
     'ts_data_type': MODEL_MAPPING['DEFAULT']['ts_data_type'],
 }
--- a/Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
+++ b/Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
@@ -110,8 +110,8 @@ MODEL_MAPPING['DEFAULT'] = {
 MODEL_MAPPING['TFT_HEAT_SUPPLY'] = {
     'load_covariates_func': get_covs_heat_supply,
     'fc_model': {
-        "model_path_ckpt_link":  "https://www.dropbox.com/s/fxt3iztbimvm47s/best.ckpt?dl=1",
-        "model_path_pth_link":  "https://www.dropbox.com/s/ntg8lgvh01x09wr/_model.pth.tar?dl=1",
+        "model_path_ckpt_link":  "https://www.dropbox.com/s/ia1nwqbepu8lbeh/best.ckpt?dl=1",
+        "model_path_pth_link":  "https://www.dropbox.com/s/8pnne7gv34btr9x/_model.pth.tar?dl=1",
         'train_again': False,
         'name': 'tft',
         'scale_data': True,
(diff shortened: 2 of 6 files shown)


Agents/ForecastingAgent/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 
 services:
   forecasting_agent:
-    image: ghcr.io/cambridge-cares/forecasting_agent:1.1.0
+    image: ghcr.io/cambridge-cares/forecasting_agent:1.1.1
```
sm453, line 5: If you update the version number here, then please don't forget to update the version number in `/ForecastingAgent/stack-manager-input-config/forecasting-agent.json`.

Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
```diff
@@ -110,13 +110,13 @@
...
-        "model_path_pth_link":  "https://www.dropbox.com/s/ntg8lgvh01x09wr/_model.pth.tar?dl=1",
+        "model_path_ckpt_link":  "https://www.dropbox.com/s/ia1nwqbepu8lbeh/best.ckpt?dl=1",
+        "model_path_pth_link":  "https://www.dropbox.com/s/8pnne7gv34btr9x/_model.pth.tar?dl=1",
         'train_again': False,
         'name': 'tft',
         'scale_data': True,
     },
     'frequency': MODEL_MAPPING['DEFAULT']['frequency'],
-    'data_length': MODEL_MAPPING['DEFAULT']['data_length'],
+    'data_length': 24 * 365 ,
```
sm453, line 120: Really? Hard-coding the data length instead of reading it in seems a step backwards to me...

Agents/ForecastingAgent/setup.py
```diff
@@ -16,11 +16,17 @@
...
         'pandas~=1.5.1',
-        'py4jps~=1.0.30', 
         'requests~=2.28.1',
-        'darts~=0.21.0',
         'configobj~=5.0.6',
-        'fire~=0.4.0'
+        'fire~=0.4.0',
+        # Pin py4jps version to avoid issues with accessing StackClients resource
+        # in latest 1.0.35 release
+        'py4jps==1.0.34',
```
sm453, line 24: Just for the record here, note issue #800.

sm453 (changes requested):

Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
markushofmeister: The default `data_length` has been hard-coded before, just that it was referenced to the value specified in the `DEFAULT` configuration [here](https://github.com/cambridge-cares/TheWorldAvatar/blob/1069a7ce7e73712988aaa7d5677f65e9a4f33e35/Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py#L93) and not with another, more appropriate value. I believe this change is useful, as the transformer model has been trained on normalised values. So, giving a `data_length` of one year means that the values are also normalised over an entire year and not just 1000 timesteps as previously.

**Please note** that this is only the default `data_length`. If the HTTP forecast request contains a `data_length` entry, of course, that one is used. So, I disagree with the step backward...

Agents/ForecastingAgent/docker-compose.yml
markushofmeister: Addressed by [c486fb9](https://github.com/cambridge-cares/TheWorldAvatar/pull/798/commits/c486fb9b1517ec6e9770a0f2a5d59e45a773bb0f) and will push docker image after successful merge to `main`

Agents/ForecastingAgent/forecasting/datamodel/data_mapping.py
sm453: How come, then, that e.g. the frequency is being read from the default model mapping, but not the data length? Seems inconsistent?

markushofmeister: Happy to change it, but the frequency of the data series is independent of the forecast model to use. So, whether one forecasts with `Prophet` or `TFT`, both forecast hourly time series; however, `Prophet` does not use scaled data and, hence, works with shorter data lengths than TFT

sm453: That confuses me even more, as it sounds like the frequency has more reasons to be hard-coded than the data length...

Anyway, the bottom line is: As a rule, hard-coding things is bad. I just don't see why, amongst a group of quantities that are all being read from some model default (which looks like good design to me), one is being singled out and hard-coded to an arbitrary, certainly use-case-dependent value.

markushofmeister: Firstly, I believe you are misinterpreting something here: The model default is defined (using hard coded values) just a few lines above that location in the same file. So, it does not make a big difference (to me at least) whether I reference those values or define them explicitely. However, a default value of 8760h simply makes more sense for the transformer model which relies on scaled data (i.e., a year of data available for scaling yields  better results as just 1000 time steps which is sufficient for Prophet though)

Secondly, this hotfix was intended to solve the model loading issue on CPU-only machines. I only made this change as this would improve the default prediction quantitiy form the transformer model. If that's an issue happy to revert this line

sm453 (approved):

PR #807, opened 2023-06-26
hotfix-geoserver-client: Bumped the stack version number.

gpeb2:
Fixes issue with the ```GeoServerClient``` class being used from an agent at least sometimes.


kok-foong (approved):
Tested

PR #824, opened 2023-07-09
Dev 1505 proximity sensor for lab fh agent

Michael-TeguhLaksana:
As FHAgent seems to work properly, it should be ready to merge with main.


Abyssus1010:
Hi Michael, we implemented the data bridge functionality into the fh agent as a temporary measure previously, it might be good to convert it to an agent that is compatible with both standalone and stack deployment once you have the time.

Abyssus1010 (changes requested):

Abyssus1010:
Found a minor bug:
/instantiate is ran against an authenticated blazegraph and no sparql credentials were provided -> returns error ->
sparql credentials are filled in the client.properties file and /instantiate is ran again -> no error is returned -> checking blazegraph reviews that no triples has been instantiated

Abyssus1010:
> Found a minor bug: /instantiate is ran against an authenticated blazegraph and no sparql credentials were provided -> returns error -> sparql credentials are filled in the client.properties file and /instantiate is ran again -> no error is returned -> checking blazegraph reviews that no triples has been instantiated

bug can be temporarily fixed by doing the following:
Delete iriMapping.txt that was generated under the config folder after the first /instantiate -> filled in sparql credentials in the client.properties file -> run /instantiate again

Agents/FHAgent/FHAgent/src/main/java/uk/ac/cam/cares/jps/agent/fh/FHAgentDerivation.java
```diff
@@ -0,0 +1,278 @@
...
+    /*
+     * Create agent and derivatiopn instances for all derivation pair
+     * @param agenURL the URL from which the agent is called
+     */
+    public void instantiateAgent(String agentURL) {
+        if (agentURL.equals(null)){
+            agentURL = derivationInstanceBaseURL;
+        }
+        try{
+            for(String rawVar : RawToDerivedMap.keySet()){
```
Abyssus1010, line 237: Found cause of minor bug:

RawToDerivedMap is empty if IRIMapper.txt already exist
RawToDerivedMap is only populated under CreateIRI (Line 140)
CreateIRI is only executed (Line 218) when IRIMapper.txt does not exist

Abyssus1010: Agent does not instantiate rdf:type om:Length and rdf:type ontodevice:OccupiedState for the variables. Might make more sense for the FH-Agent to do this? 

Agent's supposed workflow requires the agent to be used together with the DevInstAgent but this information was omitted from the README previously. Will need to work out what is the workflow of the FH-Agent with the DevInstAgent and include more information in the README.

Abyssus1010:
Agent does not instantiate rdf:type om:Length and rdf:type ontodevice:OccupiedState for the variables. Might make more sense for the FH-Agent to do this?

Agent's supposed workflow requires the agent to be used together with the DevInstAgent but this information was omitted from the README previously. Will need to work out what is the workflow of the FH-Agent with the DevInstAgent and include more information in the README.

Abyssus1010 (approved):

PR #825, opened 2023-07-09
Dev 1505 proximity sensor for lab dev inst agent _ FAILED

Michael-TeguhLaksana:
The DevInstAgent seems to work properly for now.


mdhillman:
This branch needs a little investigation, do not approve/merge until this is done.

gpeb2:
I think the cause of the Git issues was commit 06fe80b where the feature branch was merged into a local copy of the ```main``` branch. Commit 4fae184 is also a red flag as merges from remote branches into their local equivalent should rarely require an actual commit as they should be done using fast-forward.

Michael-TeguhLaksana:
The suggested commit [4fae184](https://github.com/cambridge-cares/TheWorldAvatar/commit/4fae184ea3e8359fd7e8c5b774c0c856f9601a31) seems to be the issue. I tried reverting the commit, but the issue of handling the large number of conflicts on codes I did not develop seems a bit risky (a good lesson on why a PR should be done quickly after development...). I think its best if I recreate the feature branch again. Since the agent itself is not the issue, I can copy paste the agent's folder to a new branch.

gpeb2:
Sorry, I hadn't made it clear that I'd fixed the issue, nothing else needed doing.

PR #827, opened 2023-07-11
Dev update timeseries ontology

markushofmeister:
Corresponding MIRO board: https://miro.com/app/board/uXjVPFaO5As=/


JPS_Ontology/ontology/ontotimeseries/OntoTimeSeries.csv
```diff
@@ -19,21 +19,20 @@ http://www.w3.org/2006/time#unitType ,Object Property,,,http://www.w3.org/2006/t
...
 http://www.w3.org/2006/time#TimePosition,Class,http://www.w3.org/2006/time#TemporalPosition,IS-A,,,,"A temporal position described using either a (nominal) value from an ordinal reference system, or a (numeric) value in a temporal coordinate system",http://www.w3.org/2006/time#,Time position
 http://www.w3.org/2006/time#TRS,Class,,,,,,"A temporal reference system, such as a temporal coordinate system (with an origin, direction, and scale), a calendar-clock combination, or a (possibly hierarchical) ordinal system",http://www.w3.org/2006/time#,Temporal coordinate system
-http://www.w3.org/2006/time#numericPosition,Data Property,,,http://www.w3.org/2006/time#TimePosition,Integer,,The (numeric) value indicating position within a temporal coordinate system,http://www.w3.org/2006/time#,Numeric position
+http://www.w3.org/2006/time#numericPosition,Data Property,,,http://www.w3.org/2006/time#TimePosition,decimal,,The (numeric) value indicating position within a temporal coordinate system,http://www.w3.org/2006/time#,Numeric position
```
sm453, line 36: What is the motivation/need for this change?

markushofmeister: I believe the previous range  with `xsd:integer` is not compliant with the original definition by the time ontology (https://www.w3.org/TR/owl-time/#time:numericPosition). So, just wanted to fix this

sm453: Ah, ok, I can see now that this is correct.

But this brings me to the next issue: Why are we repeating so many definitions from the time ontology here?

sm453: While I can see these practical advantages, they are rather short-sighted. Redundant repeats of other people's definitions must be avoided as a matter of principle. Imagine what happens if/when these people decide to change their definitions - assuming they are eternally static isn't realistic. Sure, you might argue that if they change their definitions everything breaks anyway, but I'd rather not waste my time trying to find all the repeats... Things should be defined in one place, and in one place only. It's the same thing in software engineering: codes with lots of repetition are unmaintainable...

msff2: Though we indicate the source of a concept or property in the defined-by annotation property, copying a definition completely from another ontology should be avoided as it might raise some issues. One can easily follow the defined-by link to check the definition. I would define a reused concept or property if I had to modify it to make it more specific or generic. I would leave the corresponding column empty if there is nothing new to communicate via the definition.

markushofmeister: Addressed by commit [47df945](https://github.com/cambridge-cares/TheWorldAvatar/pull/827/commits/47df9450fa6fc273a6c9e8d62787deffeaebda49)

markushofmeister:
I removed all concepts and relationships, which were purely included for clarification/reference. The ontology now contains new definitions only. The corresponding [MIRO board](https://miro.com/app/board/uXjVPFaO5As=/) still contains more elaborate representation about the "intended" usage/linkage of the external concepts in a dediacted color to ease usage. I believe that's a good compromise

msff2 (approved):

PR #831, opened 2023-07-12
Dev stackerise ifc agents

sm453:
Added files and documentation to facilitate running the three IFC agents in the stack. Also, made a few minor changes in the tileset agent, including a fix/work-around for a permission issue that prevented it from working in a native Linux environment.


qhouyee (approved):
Added some instructions for agent-specific bind mounts

PR #834, opened 2023-07-13
Dev district heating trigger agent

markushofmeister:
This District Heating Optimisation Trigger Agent triggers recurring optimisations of the district heating system in Pirmasens (including the simulation of corresponding emission dispersion) according to the latest use case design using [chained derivations](https://lucid.app/documents/view/9fabc350-143b-4ca3-be52-b174c9f82c07).

The agent is still WIP, but looping you in early to ensure alignment of the derivation inputs / chaining (e.g., `SimulationTime` concept, etc.)


Agents/DistrictHeatingOptimisationTriggerAgent/README.md
```diff
@@ -0,0 +1,131 @@
...
+        # Dispersion Interactor HTTP URL to establish dispersion derivation markup
+        "DISPERSION_INTERACTOR_URL=<agent http url>"
+    ],
+```
+
+**Please note:** 1) The specified namespace needs to exist/be created in Blazegraph beforehand to avoid agent execution issues. 2) The entire [resources] folder is mounted into the agent container; hence, the absolute path in the `Source` node of the config file likely needs to be adjusted.
+
+If you want to spin up this agent as part of a stack, do the following:
+1) Pull the Docker image `docker pull ghcr.io/cambridge-cares/dh-optimisation-trigger-agent:1.1.0` (alternatively, build with the commands above, but do not spin up the image)
+2) Copy the `dh-optimisation-trigger-agent.json` file from the [stack-manager-input-config] folder into the `inputs/config/services` folder of the stack manager
```
sm453, line 74: Here, I believe it needs to be mentioned that both the path of the bind mount and stack-internal URLs may need to be adjusted.

Agents/DistrictHeatingOptimisationTriggerAgent/README.md
```diff
@@ -0,0 +1,131 @@
...
+3) Add the service to a corresponding stack configuration json in `inputs/config`
+4) Start the stack manager as usual (i.e. `bash ./stack.sh start <STACK_NAME>` from the stack-manager repo). This should start the container. Please use a bash terminal to avoid potential issues with inconsistent path separators.
```
sm453, line 76: This whole readme file seems to suggest that the stack name is arbitrary. Currently, that isn't the case, because there is a particular one hard-coded into `delayed_start.sh`. Please mention somewhere in this readme that the stack name may need to be adjusted in that script prior to building the image. Alternatively, find a way to make the script generic so that it works for any stack name.

sm453 (changes requested):

kok-foong (approved):
tested while working on the district heating stack

Agents/DistrictHeatingOptimisationTriggerAgent/README.md
markushofmeister: Addressed by [bd85807](https://github.com/cambridge-cares/TheWorldAvatar/pull/834/commits/bd8580714ac7648b4e29277bc89d6aef1b1e7ceb)

Agents/DistrictHeatingOptimisationTriggerAgent/README.md
markushofmeister: Addressed by [bd85807](https://github.com/cambridge-cares/TheWorldAvatar/pull/834/commits/bd8580714ac7648b4e29277bc89d6aef1b1e7ceb)

sm453 (approved):
Tested on production server.

As a general comment, stack names should not _ever_ be hard-coded anywhere (because this essentially nullifies the years of effort that we have invested into making sure URLs, namespaces, etc. aren't hard-coded...).

PR #836, opened 2023-07-13
1614-dev-extend-ontobms: Add new concepts into ontobms

Abyssus1010:


Abyssus1010:
Added ExhaustVAV and SupplyVAV as subclass of VAVSystem.
Added ExtractionArm as subclass of s4bldg: FlowTreatmentDevice.
Added property hasOpening with ExtractionArm as the domain and om:Percentage as the range

sm453 (approved):

Issue #832, opened 2023-07-12
IFC tileset agent correctly produces assets but does not refer to them in tileset json


sm453:
This has been observed when converting the PS measurement station IFC file into a tileset. All asset files appear to be generated correctly, but the main json file mentions only one of them, which is then the only one that can be seen in a DTVF visualisation.

PR #842, opened 2023-07-18
Dev ifc2tileset issue 832 fix

qhouyee:
This pull request is a bug fix for Issue #832. Furniture assets should now be properly referenced within the generated tileset


sm453 (changes requested):
Tested on PS measurement station. The furniture is now correctly included into the tileset, and is displayed correctly in the DTVF.

I note however that, while the asset has `name` and `iri` properties, the furniture has neither. In order for it to work in the DTVF and the Feature Info Agent in particular, we need both those fields. Please also note that it has to be _exactly_ those fields (`buildingIri` won't work). Would it be possible to add/change that?

qhouyee:
> I note however that, while the asset has `name` and `iri` properties, the furniture has neither. In order for it to work in the DTVF and the Feature Info Agent in particular, we need both those fields. Please also note that it has to be _exactly_ those fields (`buildingIri` won't work). Would it be possible to add/change that?

Hello, yes, I have made these changes so that furniture and other root content will include name and iri properties. And they are now captured within the TWA-VF. 

The key changes for this feature is as follows:
1) If there is a `building.glb` content, the building iri and name will be queried from the knowledge graph and attached to all root content including furniture (if any) in the `bim_tileset`
2) If there is no `building.glb` content, root contents including furniture in the `bim_tileset`  will not have these properties by default. However, users can pass in optional parameters rootIri and rootName that will be attached instead
3) Removed the IFC uid field from assets, as they can be queried from the Feature Info Agent instead

sm453:
Tested the optional root name/IRI feature on the PS measurement station - works.

In another case, when there is a main building tileset, but in addition a sewage and a solar panel one, the latter two neither have a name nor an IRI. Is there any (reasonable) way to add such things? (I guess at least the sewage tileset probably shouldn't be lumped together with the building, so one could argue that the problem lies with the input...)

qhouyee:
> Tested the optional root name/IRI feature on the PS measurement station - works.

That's great to hear!

> In another case, when there is a main building tileset, but in addition a sewage and a solar panel one, the latter two neither have a name nor an IRI. Is there any (reasonable) way to add such things? (I guess at least the sewage tileset probably shouldn't be lumped together with the building, so one could argue that the problem lies with the input...)

Hmm, I am able to add the name and IRI as additional optional parameters or through the properties file. This should be a minor modification to the code. 

But could I clarify if we should combine these tilesets? Initially, separating these tilesets offers an easy way to turn off the extra sewage and solar panel when unnecessary. But I am unsure if CesiumJS's native styling can detect the features metadata properties and turn them off? I am experimenting with it at the moment, but it does not seem to work when I attempt to target ${feature['content']['metadata']['properties']['name']}, ${feature['properties']['name']} , ${feature['name']} and ${name}. In the case of the b3dm files, using feature['name'] works

sm453:
> But could I clarify if we should combine these tilesets?

I don't think we should. As I tried to explain above, e.g. for house 45, three separate tilesets are generated from a single Ifc input file, one for the building, one for the solar panels, and one for the sewage system. Only the first of these three has metadata. It is great to have separate tilesets, but they need to have (name and IRI) metadata (otherwise there is not much point including them in a visualisation). The point I was trying to make is that perhaps this problem could be circumvented with the current code by simply putting the house, solar panels, and sewage system each into separate Ifc input files? This would be a work-around at best, though. It would be much preferable to fix the code, in my view.

qhouyee:
Hello, I have added these optional IRI and names into the properties.yaml and updated the instructions for them. Note that the optional request parameters for root iri and name has been refactored into this YAML file for convenience.

sm453 (commented):
I have tested the name and IRI metadata functionality. I'm happy with it.

However, I have discovered another issue: The generated sewage tileset for house 45 only seems to contain the manhole covers, but not the pipes. At least that's what the visualisation shows.

sm453:
Tested the latest version - the sewage pipes are now appearing as expected.

Another problem, I'm afraid: For some reason, the tilesets for the two schools do not have a name in their metadata. They do have an IRI, though. Everything else has correct name and IRI metadata.

qhouyee:
Hello, this issue arise from the source IFC file, which I have not added the building name previously. I have added documentation to the readme for how to do this and logging to warn the user if there is a missing name. But I allow the agent to continue to run as the name may not always be crucial.

sm453:
> Hello, this issue arise from the source IFC file, which I have not added the building name previously. I have added documentation to the readme for how to do this and logging to warn the user if there is a missing name. But I allow the agent to continue to run as the name may not always be crucial.

Why is the `bim_tileset_name` from `properties.yaml` not used (if there isn't a name in the IFC file)?

qhouyee:
As per the discussion, the `bim_tileset_name` field is not used as the name will not correspond to the building name in the knowledge graph. 

Rather, I have refactor the code so that the building name will not be instantiated if there is no name in the source IFC file. And if the tileset agent does not detect this name or is an empty string, it will not generate a name for that tileset (in the situation that there is a building instance for that IFC model).

sm453 (approved):
Tested - all good now.

Issue #601, opened 2023-02-13
Stack asking for Blazegraph credentials despite absent secrets file


markushofmeister:
I use the stack with a `postgres` and `geoserver` password as required; however, without a Blazegraph password. Sometimes when starting the stack (especially after crashing or becoming unresponsive), Blazegraph nevertheless starts up as password protected. This vanishes again mostly after running `bash .\stack.sh remove <stackname>` and a subsequent restart

sm453:
I have also observed this, occurring rather randomly. I have not found a way to consistently reproduce it.

gpeb2:
I've found the cause of the issue it's that if the Blazegraph password hasn't been specified by the user the `ContainerService::ensureOptionalSecret` method creates a dummy docker secret to keep the containers that are expecting one happy. The issue comes when the Blazegraph container is stopped without the dummy secret being removed as when it is started again it now thinks that there is a password set.

The short term solution is to always use the `./stack.sh rm` command to remove a stack as this will also remove the secrets.

The long term solution is to add a label to that secret marking it as being a dummy. This will require some changes to the `PodmanClient` as there have been issues adding labels to secrets in Podman in the past. I think those can be over come as we are now using the Podman API rather than their basic implementation of the Docker one to create secrets when running with Podman.

gpeb2:
The other solution is to just set a password.

PR #846, opened 2023-07-21
Spin up a separate Ontop container for each Postgresql database

gpeb2:
Ontop can only map to one database but it is often benificial to put different datasets in their own databases. These changes cause the stack-data-uploader to spin up a new Ontop container if a dataset includes an Ontop maping file.


cmclscourtney (changes requested):
Reviewed code and tested. All looks good on that front. 

Would be good to add something to the documentation on this. These few issues would be worth noting.
- [ ] New UI endpoints 
- [ ] New internal endpoints that are created.
- [ ] ```postgres``` database uses default ontop.

gpeb2:
I've updated the FeatureInfoAgent so that it is now possible to use placeholders like ```[ONTOP-<dataset name>]``` (case-insensitive) after the ```SERVICE``` keyword to direct subqueries to a particular Ontop container. For example ```SERVICE [ontop-pirmasens]```.
@cmclscourtney and @YiKaiTsai, please can you test this works for you?

YiKaiTsai:
Was not able to successfully query from Ontop endpoint when spinning up FeatureInfoAgent (FIA) or querying from Blazegraph. Details are as follows:

1) FIA did not work when using placeholder ```[ONTOP-datasetName]```. When checking the log, the SPARQL that FIA tried was
```
select distinct ?class {
         SERVICE <http://stackName-ontop:8080/sparql/> { <iri> a ?x . }
         ?x rdfs:subClassOf* ?class .
         ?class rdf:type owl:Class .
}
```

It seems that the placeholder did not get properly converted to SERVICE IRI ```<stackName-ontop-datasetName:8080/sparql>```.

2) Using SERVICE IRI ```<stackName-ontop-datasetName:8080/sparql>``` inside Blazegraph did not work. The error was ```java.util.concurrent.ExecutionException: java.net.UnknownHostException: stackName-ontop-datasetName: Name or service not known```.

gpeb2:
This should also close issue #601.

cmclscourtney (approved):

PR #812, opened 2023-06-28
808 add pgrouting dataset type to stack data uploader

cmclscourtney:
Adds a `OSMRouting` data-subset type to the `stack-data-uploader`. An simple example would be the following with `routing` directory containing a `.osm` and `.xml` config file.

```json
{
    "database": "postgres",
    "workspace": "the_world_avatar",
    "datasetDirectory": "routing",
    "dataSubsets": [
        {
            "type": "OSMRouting",
            "skip": false,
            "schema": "public",
            "name": "my_routing",
            "osm2PGRoutingOptions": {
                "flags": [
                    "--attributes"
                ],
                "options": {
                    "--chunk": "40000"
                }
            }
        }
    ]
}
```

PR #864, opened 2023-08-03
Update README.md

Ushcode:
data uploader threw an error at unrecognised keyword `parameter` but does recognise `parameters`


cmclscourtney:
This is resolved by d2feff3 as part of #812 which will shortly be merged

PR #866, opened 2023-08-04
Dev forecast derivation agent

markushofmeister:
This PR converts the ForecastingAgent into a derivation agent using the Derived Information Framework


Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
```diff
@@ -0,0 +1,338 @@
...
+                # NOTE: Derivation Framework only supports one output per rdf type
+                self.sparql_client.reconnect_forecast_with_derivation(old_fc_iri, fc_iri)
+
+            # Add output graph to ensure complete derivation markup
+            # --> this part of the code is only relevant when called via 
+            # 'createSyncDerivationForNewInfo' and its only purpose is to ensure
+            #  that forecast instance is marked up as "belongsTo" the derivation
+            g = Graph()
+            g.add((URIRef(fc_iri), URIRef(RDF_TYPE), URIRef(TS_FORECAST)))   
+            derivation_outputs.addGraph(g)
```
jb2197, line 241: For the derivation created when calling `createSyncDerivationForNewInfo`, the standard practice is to add all triples to `derivation_outputs` so that no duplicate information is added to the triple store when the agent handling concurrent update requests. See https://github.com/cambridge-cares/TheWorldAvatar/issues/184 for more information. So I would suggest moving triples in `self.sparql_client.instantiate_forecast` to part of `derivation_outputs`. That said, I suppose the separation here is to support both creating new derivations and updating existing derivations (when not overwriting time series data) - but this goes back to the question of why such a design choice is made.

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
```diff
@@ -0,0 +1,338 @@
...
+    def process_request_parameters(self, derivation_inputs: DerivationInputs, 
+                                   derivation_outputs: DerivationOutputs):
+        """
+        This method takes 
+            1 IRI of OntoTimeSeries:ForecastingModel (model to be used for forecasting)
+            1 IRI of OntoTimeSeries:Frequency (frequency of forecasted time series)
+            1 IRI of Time:TimeInterval (time interval to be forecasted)
+            1 IRI of Time:TimeDuration (data length prior to forecast interval
+                                        to be used for training/data scaling)
+            1 IRI of OM:Quantity or owl:Thing (the concept associated with the
```
jb2197, line 171: I still think it's less cumbersome to define `om:Quantity` as the domain of `OntoTimeSeries:hasTimeSeries` as opposed to `owl:Thing` and it will work for most of the use cases. This might be a question to @sm453.

Agents/ForecastingAgent/docker-compose-test_dockerised.yml
```diff
@@ -0,0 +1,48 @@
...
+  forecasting_agent:
+    image:  ghcr.io/cambridge-cares/forecasting_agent:2.0.0
+    build:
+      context: .
+      target: prod
+    env_file:
+      # Specify environment variables for testing
+      - agent_test.env
+    environment:
+      - OVERWRITE_FORECAST=true
```
jb2197, line 13: Is it possible to make `OVERWRITE_FORECAST` something that can be queried from the triple store so it can be decided per derivation? It seems a massive resource requirement to have two agents spun up for both cases - considering each docker container is at least 6.5 GB!

Agents/ForecastingAgent/README.md
```diff
@@ -1,234 +1,316 @@
...
 
-2) A missing `STACK_NAME` variable will result in an error; however, when deploying using the stack-manager start up script, the `STACK_NAME` variable will be set automatically for all services. Hence, this could be left blank here; however, if provided, it needs to match the `STACK_NAME` used by the stack-manager!
+Please note: 
+- The `DB_URL` is optional as the agent will first try to retrieve the database endpoint instantiated for the time series IRI to forecast from the KG (and only use the specified environment variable as fallback)
+- When deployed via the Stack manager, the `STACK_NAME` environment variable is set automatically and can be omitted in the config file
+- `SPARQL_QUERY_ENDPOINT` and `SPARQL_UPDATE_ENDPOINT` are both required inputs even for stack deployment (due to default derivation agent config); however, as the actual values for initialising the Forecasting Agent are retrieved via the Stack clientsm they can be left blank.
```
jb2197, line 59: "Stack clientsm" --> "Stack clients"

Agents/ForecastingAgent/forecastingagent/kgutils/kgclient.py
```diff
@@ -0,0 +1,486 @@
...
+    def instantiate_new_forecast(self, cfg:dict):
+        """
+        Returns SPARQL INSERT DATA query to instantiate a new forecast
+        """
+        
+        # Create IRIs for new instances
+        outputTimeInterval_iri = KB + 'Interval_' + str(uuid.uuid4())
+        inputTimeInterval_iri = KB + 'Interval_' + str(uuid.uuid4())
+
+        # Initialise SPARQL update body witth forecast related triples
```
jb2197, line 375: "witth" --> "with"

jb2197 (changes requested):
All integration tests work fine on my machine (Win11, WSL2).

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
sm453: Well, `om:Quantity` is much more restrictive than `owl:Thing`. Not everyone uses OM. Unless it causes _substantial_ trouble to use `owl:Thing`, I'd go for that. Having said that, if the agent can _only_ forecast our own time series class, and that is restricted to a certain type, then there is no benefit in being more general here. Another question: You say "most" of the use cases. Do we already have use cases where `om:Quantity` is not general enough?

sm453: The thing is (pardon the pun), as long as the derived information framework matches inputs based on `rdf:type`, it makes no sense to use `owl:Thing` as an input type, because _everything_ matches that (_any_ `owl:Class` is a subclass of `owl:Thing`!). For that reason, we must not use `owl:Thing` here (at least until issue #896 is resolved).

So, I would suggest sticking to `om:Quantity` for the time being. Let's hope that no-one is using time series of e.g. OntoCAPE quantities...

Agents/ForecastingAgent/docker-compose-test_dockerised.yml
markushofmeister: Of course, one could do that; however, as it has been decided that overwriting forecasts shall be the default behaviour anyway (as not doing so partially contradicts the DIF), I do not see the point of doing so. We should rather make not overwriting less prominent. Ok?

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
```diff
@@ -0,0 +1,338 @@
...
+        return [TS_FORECAST]
+
+
+    def validate_inputs(self, http_request) -> bool:
+        # Validate completeness of received HTTP request (i.e. non-empty HTTP request, 
+        # contains derivationIRI, etc.) -> only relevant for synchronous derivation
+        return super().validate_inputs(http_request)
+    
+
+    def validate_input_values(self, inputs, derivationIRI=None):
```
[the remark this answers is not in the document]
markushofmeister: Because it's just used for logging purposes

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
```diff
@@ -0,0 +1,338 @@
...
+            if len(inp) == 1:
+                input_iris['iri_to_forecast'] = inp[0]
+            elif len(inp) == 0:
+                self.logger.error(f"Derivation {derivationIRI}: Neither 'om:Quantity' nor 'owl:Thing' IRI provided to forecast.")
+                raise TypeError(f"Derivation {derivationIRI}: Neither 'om:Quantity' nor 'owl:Thing' IRI provided to forecast.")
+            else:
+                self.logger.error(f"Derivation {derivationIRI}: No unique 'owl:Thing' IRI provided to forecast.")
+                raise TypeError(f"Derivation {derivationIRI}: No unique 'owl:Thing' IRI provided to forecast.")
+        
+        # 2) For existing derivations
```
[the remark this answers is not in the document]
markushofmeister: I thought the same initially, but that's not the case (if I remember correctly). If I want to forecast `dh:HeatDemand` which is a subclass of `om:Quantity`, then
1) for createNewInfo, the IRI will be returned as part of the `om:Quantity` type
2) for derivation update, the IRI will be returned as part of `dh:HeatDemand` type
Hence, both cases need to be handled differently

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
markushofmeister: It does not cause *substantial* trouble to use `owl:Thing`:
1) Since only one time series is forecasted (and all other derivation input types are clearly defined), it is easily possible to extract the IRI to forecast unambiguously
2) The agent does not take our own time series concept instances directly, but rather the associated dataIRIs to derive the forecasts; hence, the time series design does not restrict the possible classes to forecast
3) For energy demand modelling, one might potentially want to forecast (room) occupancy, or the availability of boilers for the district heating. So, there are cases where target concepts are not defined as `om:Quantity`. Furthermore, `owl:Thing` is more in line with the time series ontology

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
```diff
@@ -0,0 +1,338 @@
...
+                         db_url=rdb_url, time_format=time_format)
+        # Extract times and values from darts TimeSeries
+        times = [str(x) for x in fc_ts.time_index.tz_localize('UTC').strftime(time_format)]
+        values = fc_ts.values().squeeze().tolist()
+        if self.fc_round:
+            # Round data values if set in environment variables
+            values = [round(x, self.fc_round) for x in values]
+
+        # Instantiate new forecast in KG and RDB (if requested or not yet existing)
+        if not self.fc_overwrite or not bool(cfg.get('fc_iri')):
```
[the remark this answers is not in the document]
markushofmeister: That stems from an earlier discussion with Seb and Jethro, where we discussed potential scenarios for creating forecasts: 1) a forecast update either completely replaces the previous forecast or 2) we keep the previous forecast as is and create a completely new one (i.e., not appending to create sort of mixed outputs)
As of our last discussion, 1) shall be the default as mostly in line with DIF design; hence, I removed most references to use the agent without overwriting (but kept the functionality in for reference). This was agreed with Seb

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
markushofmeister: Not all triples from `self.sparql_client.instantiate_forecast` can be added to `derivation_outputs`, as the TimeSeriesClient interacts with the KG directly when writing time series data; however, as overwriting forecasts is now the default, does this "issue" actually still exist?

Agents/ForecastingAgent/README.md
```diff
@@ -1,234 +1,316 @@
...
+# Create new forecast derivation
+agent_iri = <ONTOAGENT_SERVICE_IRI of target Forecasting Agent instance>
+derivation_input_set = [
+    <IRI of om:Quantity or owl:Thing instance>, 
+    <IRI of ts:ForecastingModel instance>, 
+    <IRI of ts:Frequency instance>, 
+    <IRI of time:Interval instance>, 
+    <IRI of time:Duration instance>
+]
+derivation_type = <https://www.theworldavatar.com/kg/ontoderivation/DerivationWithTimeSeries>
```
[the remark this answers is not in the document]
markushofmeister: Addressed by [3c17cdb](https://github.com/cambridge-cares/TheWorldAvatar/pull/866/commits/3c17cdb5136b19cdebecb4552c140640ec2d13e9)

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
```diff
@@ -0,0 +1,338 @@
...
+            # Initialise forecast in KG
+            fc_iri = self.sparql_client.instantiate_forecast(forecast=fc_ts, config=cfg,
+                                                             create_new=True)
+            # Create new forecast instance and add ts data in RDB
+            ts_client.init_timeseries(dataIRI=fc_iri, times=times, values=values, 
+                                      ts_type=cfg['ts_data_type'], time_format=time_format)
+            if old_fc_iri:
+                # In case new forecast has been created for existing derivation,
+                # disconnect old forecast instance from derivation and connect with new one
+                # NOTE: Derivation Framework only supports one output per rdf type
```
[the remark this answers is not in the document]
markushofmeister: Addressed by [9139a31](https://github.com/cambridge-cares/TheWorldAvatar/pull/866/commits/9139a31bc1c8c81181e80746085d4bc6b714914e)

Agents/ForecastingAgent/README.md
markushofmeister: Addressed by [6db7785](https://github.com/cambridge-cares/TheWorldAvatar/pull/866/commits/6db7785cb80e5a6a11c7765d76b6734fdea5f8b0)

Agents/ForecastingAgent/forecastingagent/kgutils/kgclient.py
markushofmeister: Addressed by [6db7785](https://github.com/cambridge-cares/TheWorldAvatar/pull/866/commits/6db7785cb80e5a6a11c7765d76b6734fdea5f8b0)

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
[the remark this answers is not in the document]
jb2197: Could you please add this as comments to the code? This will be helpful for others.

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
jb2197: I agree the issue is resolved by itself if "overwriting forecasts" is made as default. But I would recommend documenting this somewhere in the code as the functionality (not overwriting) is still kept.

Agents/ForecastingAgent/docker-compose-test_dockerised.yml
jb2197: Same as the last comment, I would recommend documenting this somewhere so others know this is only kept in docker-compose for testing purposes.

Agents/ForecastingAgent/forecastingagent/agent/forecasting.py
jb2197: What about defining "(room) occupancy" and "availability of boilers for the district heating" as subclasses of `om:Quantity`?

Agents/ForecastingAgent/stack-manager-input-config/forecasting-agent.json
jb2197, line 1: I would recommend unifying the name of this json file and the name under `ServiceSpec` (hyphen vs. underscore). This will help avoid debugging issues like `No service config loaded with name 'forecasting_agent'.` when using stack.

sm453: I agree. Please use hyphens throughout (like most agents).

markushofmeister:
I addressed most of your comments and further elaborated the documentation to ensure that overwriting existing forecasts is the default behaviour and mostly in line with DIF.

Regarding the support for `om:Quantity` vs `owl:Thing`, this is my point of view:
1. The current implementation supporting `owl:Thing` is in line with the current time series ontology. As pointed out by Seb, not everybody uses OM; hence, this gives us more flexibility
2. Using `owl:Thing` does not cause issues when retrieving derivation inputs. As the agent handles only one instance per rdf type, it is possible to unambiguously derive the target IRI to forecast: even if any other derivation input also has an `owl:Thing` relationship attached, this particular IRI will also be retrieved for the more specific input type (e.g. ForecastingModel). So, the only IRI which only appears in the input list for `owl:Thing` is the IRI to forecast. Please see the `validate_input_values` method for details.

Hence, I believe we should keep the current implementation with `om:Quantity` as default, but also more general support for `owl:Thing`.

Lastly, @sm453 have you been able to spin up the agent as part of the stack-manager by now? I just tried again on my end and it works without issues.

jb2197:
> Hence, I believe we should keep the current implementation with `om:Quantity` as default, but also more general support for `owl:Thing`.

I suppose this is fine for the moment, could you please also add the discussion to the documentation of the codes?

jb2197 (approved):

PR #867, opened 2023-08-07
fix-dtvf-action

mdhillman:
Fixes the incorrect syntax in GitHub workflows relating to the DTVF.


mdhillman:
No functional changes are taking place here, just a syntax change in the automated GitHub workflows (to ignore PRs when only markdown files are changed). Due to the workflow triggers, it is also un-testable until merged to main.

As such, I'm making an executive decision and approving this merge without a separate reviewer (don't do this at home kids).

PR #875, opened 2023-08-15
1615 dev ontoassetmanagement

Abyssus1010:


markushofmeister (changes requested):
I had a look at the ontology using WebVowl and have the following comments:

1) Is there a reason why some properties are restricted using "only" and others are not? I could not identify a clear distinction

2) I believe the TBox needs to be accompanied by an ABox containing the 'DeliveryMethod's and 'Courier's or how did you intend to represent the corresponding instances/subclasses?

3) I find the price section still somewhat confusing: I believe some of the labels can be further refined and also some of the properties seem not fully applicable/ambiguous, e.g., a 'DiscountedPrice' can be 'subjectedTo' a 'Discount'. What would that mean? Is this the price after applying a previous discount? Or is this the discount which was considered to derive the discounted price? I guess this should be resolved

image

Abyssus1010:
> 

1. The reason for including "only" is because for "references", it's either "Item references Element" or "Document references Document". If I don't include "only", the TBoxGenerator prints out the domain and range as shown below without any restrictions applied:
image
This will mean that a Document can also reference an element and an item can reference a document.

PR #879, opened 2023-08-22
dev-rename-dtvf

mdhillman:
This PR brings in a change that renames the `DTVF` to `TWA-VF`. This rename affects the code, documentation, docker configurations, and associated release scripts.


cmclscourtney (approved):
Looks good, need to remove snapshot flag from stack before merge

PR #883, opened 2023-08-23
Dev smell agent

jlai2021:


ChungTingLao:
Whilst the actual input files do not have to be included, the README should contain information for users to know how to create the required input files and use the agent themselves.

PR #893, opened 2023-08-31
Fix issue 892: Postgres runs out of free connections

gpeb2:
Updated the naming of GeoServer "stores" to use the database name rather than the layer name. This means that layers in the same workspace being served from the same database now share a store. This can greatly reduce the number of connections between GeoServer and Postgres.


Ushcode (approved):
LGTM, stack spins up with two stores and all layers as expected. No overload error as before

cmclscourtney (approved):
Checked code, makes sense

PR #905, opened 2023-09-08
903 the stack data uploader doesnt recognise the geoserver creationoptions settings

gpeb2:
I've added the missing annotation.


kok-foong (approved):
The fix is fine but does not solve my underlying issue which is a separate problem

Issue #878, opened 2023-08-16
Feedback on Mapbox Visualisation Tutorial


ao559x:
Dear Sean and Michael:
I have been following the instructions on the tutorial for mapbox (https://github.com/cambridge-cares/TheWorldAvatar/blob/main/web/digital-twin-vis-framework/docs/tutorial-mapbox.md). The task has been successful, but I needed a bit extra help. The following comments are meant to improve the tutorial.

1)	First, I downloaded the datasets. However, for the GP Practices file, I needed to work on the dataset to get it in the correct format. The instructions were very useful; however, I think it might be much easier if the README file would show the first lines of the final result, or a limited version could be downloaded. I worked first with ten lines and manually fixed them, until I could upload the file correctly. The file for the practitioners was useful to learn what the final output looks like.  
When my 10 rows file worked, I updated the whole file. Also note, that some rows of the datasets have extra commas in the address, which produces that longitude and latitude are moved to the right. I created a quick excel algorithm to fix this, at the expense of losing a bit of information on the addresses.
Example of amendment would be (LINE 44): The final GPPractices file should look like this:
"OrganisationID","OrganisationCode","OrganisationType","SubType","OrganisationName","OrganisationStatus","IsPimsManaged","Address1","Address2","Address3","City","County","Postcode","Latitude","Longitude"
"46946","C83036","GP practice","UNKNOWN","Gosberton Medical Centre","Visible","FALSE","Lowgate","Gosberton","Spalding","Lincs","","PE11 4NL","52.86989212","-0.159257933"
"46947","A83023","GP practice","UNKNOWN","Stanley Medical Group","Visible","FALSE","Stanley Primary Care Centre","Clifford Road","Stanley","Stanley","Durham","DH9 0AB","54.87003326","-1.69803834"
"46948","A82006","GP practice","UNKNOWN","Appleby Medical Practice","Visible","FALSE","The Riverside Building","Chapel Street","","Appleby in Westmorland","Cumbria","CA16 6QR","54.57808685","-2.492977619"
2)	 It might be worth noting that CSV files can be updated in format using Excel. The comment might be added to the section “3. Fix funky characters:” in LINE 47 of README file.
![image](https://github.com/cambridge-cares/TheWorldAvatar/assets/136823167/9271e126-3806-4786-b064-9891302554a7)
 
3)	Another issue that would make things super-easy would be to include a visual or text or the data organization.
[...]

Issue #770, opened 2023-05-31
DTVF: changing "Imagery" does not preserve selected visible layers


cmclscourtney:
When changing the background map ("Imagery") the selections the selected visible layers return to their defaults. The check boxes however do not change and when any of these are altered the map reflects these check boxes.

mdhillman:
In progress on the `dev-twa-fix` branch, watch this space for a PR soon!

Issue #534, opened 2022-12-21
DTVF: Pop-up window handling


sm453:
Currently, the pop-up window when hovering over features is hard-coded to a fixed size. It would be great if this was customisable and/or auto-adjusted to the display string.

mdhillman:
I've adjusted the CSS to fix this; long site names and long descriptions now display better.

![Updated popup](https://i.postimg.cc/j5JCLqs7/long-site-name.png)

mdhillman:
The DTVF currently has two implementations for this popup system, one using the Mapbox provided classes, and a generic HTML/JS one for Cesium. This issue is being re-opened with the aim to unify these systems.

mdhillman:
Progress on this has stalled as other features have taken priority. Note that this is still an active, albiet low priority, issue.

mdhillman:
In progress on the `dev-twa-fix` branch, watch this space for a PR soon!

Issue #301, opened 2022-08-23
DTVF: Changing to the same terrain removes site locations


mdhillman:
Selecting the same "terrain" type multiple times currently causes site data to be removed.

mdhillman:
In progress on the `dev-twa-fix` branch, watch this space for a PR soon!

PR #919, opened 2023-09-15
Fixes for TWA-VF

mdhillman:
This PR brings in bug fixes for issues #534, #770, #301, and #878. The former three will need to be tested within Mapbox and Cesium visualisations (using the example ones are fine).

Update: this PR has been updated to include the following changes:
* Updated mouse handling for feature popups.
* Automated action now pushes image with more tags.
* Added null check for possible missing `layers` object when toggling Cesium layers.
* Minor update to build script for local development.

> [!WARNING]
> Note that just before this PR is merged, the version of the TWA-VF will need the "-SNAPSHOT" token removed. In fact, the automated checks *should* fail until this is done.

> [!WARNING]
> Depending on when this PR is approved, and when related PRs are approved, the specific version number may need adjusting.


mdhillman:
Converting to draft whilst I look into some further fixes.

cmclscourtney (changes requested):
- [ ] #534 Might be worth doing multiple wraps on title text for such an edge case as below.  
- [x] #534 Also should popups should go above (or to the side of) icon if would be blocked by screen
Capture

The following are some quality of life things that would be useful for ongoing development
- [x] Would be useful if each SNAPSHOT image version was pushed with a unique branch identifier to make testing easier and make sure there isn't overlap in branch names.
- [ ] A branch per issue fix would be make things easier I think but am aware you might be unwilling to spam minor versions.

mdhillman:
Updates following feedback:

* Image for this branch now includes branch name (ghcr.io/cambridge-cares/twa-vf:4.0.1-fix-twa-vf-4.0.1-SNAPSHOT).
* Scripts to run examples now default to use current version (i.e. that defined in library/VERSION file).
* Popup placement now adjusts at edges of maps.
* Popup content now gets multiple line wraps (if needed).
* Popup descriptions now cut off after 250 characters.

mdhillman:
I am unable to reproduce Seb's popup issue using the Cesium example. 

I can try to run Seb's use case (if it's simple to spin up locally), or ask Chung (who's also working with Cesium at the moment) to try it on his data.

mdhillman:
> I am unable to reproduce Seb's popup issue using the Cesium example.
> 
> I can try to run Seb's use case (if it's simple to spin up locally), or ask Chung (who's also working with Cesium at the moment) to try it on his data.

We have tried this on Chung's machine, with his Cesium data and cannot reproduce it. Later today I will try on a monitor with lower resolution, and perhaps with different Windows scaling options.

mdhillman:
This PR now closes #928.

mdhillman:
> > I am unable to reproduce Seb's popup issue using the Cesium example.
> > I can try to run Seb's use case (if it's simple to spin up locally), or ask Chung (who's also working with Cesium at the moment) to try it on his data.
> 
> We have tried this on Chung's machine, with his Cesium data and cannot reproduce it. Later today I will try on a monitor with lower resolution, and perhaps with different Windows scaling options.

Have now tried on HD and 4K screens, with 100%, 125%, and 150% windows scaling. Can still not reproduce the issue :(

mdhillman:
I have updated the feature-picking routine that is used to determine the currently hovered WMS feature in Cesium visualisations. This appears to resolve the popup placement issue experienced by @sm453.

Note that Cesium WMS feature interaction is asynchronous and can be a little slow. In testing this only causes issues if you rapidly move the mouse (so I suggest not doing that).

Updated docker image is `ghcr.io/cambridge-cares/twa-vf:4.0.1-fix-twa-vf-4.0.1-SNAPSHOT`

mdhillman:
> You're right, it's slow (and not very accurate in determining the contents), but at least the placement is correct. Good enough for now.

This is unfortunately another effect of Cesium not supporting vector data via WMS. With raster data, to get _any_ information, another request has to be made to the server. This causes a delay, which in turn causes issues if you've moved the mouse during that delay.

I have some thoughts on improving in in future, but I struggle to see how we could make it as seamless as it is with the vector data on Mapbox.

cmclscourtney (approved):

PR #920, opened 2023-09-15
dev-stack-uploader-static-file-readme-addition: Added a hint and a wa…

sm453:
…rning about GeoServer static file uploads to the stack data uploader readme file.

### DIFF
--- a/Deploy/stacks/dynamic/stack-data-uploader/README.md
+++ b/Deploy/stacks/dynamic/stack-data-uploader/README.md
@@ -274,7 +274,7 @@ The icons can be found at `GEOSERVER_URL/www/icons` and the "other files" (being
 
 Note: If you want to reference icons uploaded in this way from a GeoServer `.sld` file, you will need to use stack-internal URLs, e.g.:
 ```
-<OnlineResource xlink:type="simple" xlink:href="http://<STACK_NAME>-geoserver:8080/geoserver/www/icons/myicon.png" />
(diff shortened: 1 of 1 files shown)


Deploy/stacks/dynamic/stack-data-uploader/README.md
```diff
@@ -272,6 +272,14 @@ The icons can be found at `GEOSERVER_URL/www/icons` and the "other files" (being
 }
 ```
 
+Note: If you want to reference icons uploaded in this way from a GeoServer `.sld` file, you will need to use stack-internal URLs, e.g.:
+```
+<OnlineResource xlink:type="simple" xlink:href="http://<STACK_NAME>-geoserver:8080/geoserver/www/icons/myicon.png" />
```
cmclscourtney, line 277: Might be worth using ```localhost``` here so new style is not needed when stack name changes
```suggestion
<OnlineResource xlink:type="simple" xlink:href="http://localhost:8080/geoserver/www/icons/myicon.png" />
```

cmclscourtney (changes requested):
See change suggested

Deploy/stacks/dynamic/stack-data-uploader/README.md
sm453: Thanks for the hint - a good suggestion. I did previously try a localhost address, but I guess I must have used the wrong port number... Anyway, I have committed this.

cmclscourtney (approved):

PR #927, opened 2023-09-21
904 reduce number of time series tables

kok-foong:
A new class to interact with the database - TimeSeriesRDBClientWithReducedTables, is introduced. Usage guidelines and how it works is detailed in https://github.com/cambridge-cares/TheWorldAvatar/blob/363d76e4de38fa6f56f1d578e8ccbbc76df7fe4a/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/timeseries/README.md

Existing agents using the TimeSeriesClient should work without any changes.

@mdhillman These changes will impact the FIA because the same TimeSeriesRDBClient class needs to be used to create and query the time series data. Perhaps you can add a user input to specify which class to use, or maybe I could provide something in TimeSeriesClient for you to determine which version to use.


mdhillman:
I am going to push back on this and suggest not making this update until we review the design. I believe that the FIA configuration is already too complex for a lot of people; you already have to have a-priori knowledge of what datetime class was used when the data was input (something I am not happy with), adding knowledge of what client class was used may be to much.

What happens when the person that added the data is no longer with us and has left no documentation? What happens when we start to provide APIs for direct access to our data?

I would suggest that all metadata needed to retrieve the timeseries data should be captured as triples within the KG. We already do this with some fields (measurement name, unit etc), why are we not following the existing approach here?

mdhillman (dismissed):
I am going to push back on this and suggest not making this update until we review the design. I believe that the FIA configuration is already too complex for a lot of people; you already have to have a-priori knowledge of what datetime class was used when the data was input (something I am not happy with), adding knowledge of what client class was used may be to much.

What happens when the person that added the data is no longer with us and has left no documentation? What happens when we start to provide APIs for direct access to our data?

I would suggest that all metadata needed to retrieve the timeseries data should be captured as triples within the KG. We already do this with some fields (measurement name, unit etc), why are we not following the existing approach here?

kok-foong:
Marking up the time class is a trivial thing to do, but I don't see any way around the time class issue due to the way Java works and having the flexibility of adding any type of data you like, at most we can support a fixed set of classes and hardcode them in the FIA.

mdhillman:
My belief is that if we have to hardcode anything, then our design is incomplete. We may be able to get away with it in the FIA, but what happens when we add more functionality? Do we just endless hardcode stuff?

What happens when an new agent needs to get time series data for some calculation? What happens when we need to provide a generic API to our data (this is already being discussed with external partners at CMCL).

We have already had discussions in a number of projects about (potentially) making our data meet [FAIR Principles](https://www.go-fair.org/fair-principles/). To me, this seems to me actively moving in the other direction.

I will try to book in a discussion next week to go over this.

kok-foong:
Actually depends on how you want to query it, to get the entire time series, you can get away with using Object.class and the wildcard, it will be an issue if you want to specify the bounds

kok-foong:
A new TimeSeriesClientFactory class has been added, it queries the time class and rdb client class and then creates the appropriate TimeSeriesClient object. The variables still need to be casted if you want to perform any operation on them.

The TimeSeriesClientFactory relies on a couple of new triples not added by previous versions, so if you want to use this, the time series data need to be reinstantiated, or the triples can also be added manually.

In this revision the class of the values is still ambiguous, though for visualisation purposes, the getValuesAsString may work for a variety of types if we want a generic implementation.

ChungTingLao (approved):
The new time series client backend has been tested with virtual sensor in dev-virtual-sensor-ship-timeseries. Furthermore, it has been tested that changes of this pull request do not affect agents/stacks using the original times series client backend.

PR #933, opened 2023-09-22
Dev 1524 integration data 2

yjytt59:
Merge data integration agent to main branch


web/twa-vis-framework/example-cesium-vis/webspace/index.html
ChungTingLao, line 1: Committing API keys should be avoided unless there is a very good reason for it.

JPS_VIRTUALSENSOR/startup.sh.main
ChungTingLao, line 1: Is the data integration agent part of the virtual sensor? If not, this can be removed.

ChungTingLao (changes requested):
There is no documentation about its purpose or how to use the Building Linking Agent, so I have not tested that yet.

For the Data Integration Agent, three functions are tested.

General feedback:

Unused or commented-out codes (e.g. extractFootprint method) should be removed.

thematicParams should be (converted to) boolean if it is only going to be true or false.

DIA currently returns an empty JSON message; may be it should return some useful information.

DIA processes one building at a time. Performance may be improved if buildings are processed in batches; this would be a good compromise between processing 1 building at a time and processing all buildings at once.

"attribute"

There is an error in the SQL querying for 2D object during testing, please fix that so I can continue testing.

"height"

the code requires thematicParams to be defined, this should be checked earlier or the code to be re-written to assume false.

minZsql is not robust as it assumes all buildings have footprints defined. 0 is assumed for buildings without footprint, which is erroneous for buildings at high altitude (for example, Pirmasens buildings can be >400 m tall which is clearly wrong).

maxZsql is not robust enough, it only considers LoD2 data with thematic surfaces or LoD3 data without thematic surfaces.

The code should throw an error or a warning when the calculated height is non-positive; currently it does not do anything.

"footprint"

for data with thematic surfaces, the method extractPrint_thematic is called. It fails to detect a building with lod0_footprint_id = NULL as a building that needs to be processed. Testing cannot be continued until this is fixed.

for data without thematic surfaces, DIA will try to identify ground surfaces, but it doesn't create footprints after identification. Identification is not implemented for roofprints.

The SQL statement that identifies ground surfaces is not robust enough, it assumes the data is in LoD3. In addition, buildingid and cityobjid returned by this query is going to be exactly the same (that of the building) so it is redundant to have both of them.

The method updateSurface will need major revision. Many entries are missing for inserted objects, such as gmlid, envelope etc. DIA seems to add all ground surface geometries to 1 ground surface of the building, and then creates multiple ground surfaces with no geometry children for the building, which is unlikely to be the intended behaviour. The augmented data is no longer valid CityGML and cannot be exported back to CityGML format. Please check the building_bavaria example to see how CityGML data with thematic information should look like.

sm453:
I'm afraid neither `docker compose build` nor `./stack.sh build` work out of the box. Can you please fix this?

Agents/BuildingLinkingAgent/docker-compose.yml
sm453, line 1: This file isn't going to work when running `docker compose build`, because the `STACK_NAME` environment variable is not going to be defined.

Apart from that, is all this content really needed? The container should be spun up through the stack manager anyway, which means a simple compose file (for building) like the following does the job:
```
version: "3.8"

services:
  data-integration-agent:
    image: ghcr.io/cambridge-cares/data-integration-agent:1.0.1-SNAPSHOT
    build: .
    ports:
      - "8086:8080"
    volumes:
      - logs:/root/.jps
volumes:
  logs:
    name: "logs"
```

Agents/BuildingLinkingAgent/docker/entrypoint.sh
sm453, line 1: This file needs executable permissions for this to work on native Linux  (WSL seems to be more relaxed about these things).

Agents/DataIntegrationAgent/stack.sh
sm453, line 1: This file is not needed. I recommend removing it altogether.

Agents/utils/parent-pom/pom.xml
sm453, line 1: There are a lot of (dubious) changes part of this pull request that are outside the `Agents/DataIntegrationAgent/` folder. Please revert all such changes. They should not be part of this pull request.

JPS_VIRTUALSENSOR/startup.sh.main
yjytt59: I removed all the unused pull, includes weatheragent, virtual sensor and so on.

Agents/DataIntegrationAgent/stack.sh
yjytt59: I need them for build.

Agents/utils/parent-pom/pom.xml
yjytt59: This branch includes BuildingLinkingAgent too, which aims to link building by IRI between KG and postgresql. But during the work, Jethro and George took over this part to redesign it at top level, so the work of this agent has been paused. So far I don't want to merge this agent with main branch. I will separate BuildingLinkingAgent and DataIntegrationAgent and only pull request to merge DataIntegrationAgent only.

PR #935, opened 2023-09-25
1656 dev add threshold and meter related concepts to ontodevice and modify ontoubemmp

Abyssus1010:


sm453 (changes requested):
Apart from one minor point, looks good to me.

sm453 (approved):

PR #943, opened 2023-09-28
Dev-stack-building-thematic-vis

ChungTingLao:
New functionalities have been added for the representation of building data in the stack, including:

- Publishing a geoserver layer for CityDB data for the purpose of visualisation
- Capability to create footprints and heights of buildings during upload
- Thematic surface discovery of building data
- Export vector building data to compressed CityGML

Existing building data examples have been updated.


Deploy/stacks/dynamic/stack-data-uploader/README.md
```diff
@@ -410,18 +413,28 @@ The format is the same as the one used for [ogr2ogr otheroptions](#otheroptions)
 The list of avaliable options can be found [here][3dcitydb-importer-cli].
 The "Database connection options" are set automatically by the `stack-data-uploader` so can be ignored.
 
+#### `"augmentData"`
+
+This boolean flag controls whether footprint and height would be added to the existing data. It is assumed to be `true` if omitted.
```
gpeb2, line 418:
```suggestion
This boolean flag controls whether footprint and height will be calculated and added to the existing data, if not already present. It is assumed to be `true` if omitted.
```

Deploy/stacks/dynamic/stack-data-uploader/README.md
```diff
@@ -410,18 +413,28 @@ The format is the same as the one used for [ogr2ogr otheroptions](#otheroptions)
...
 The full list of file formats that `ogr2ogr` supports is given [here][vector-drivers] although some of these might not be available depending on the exact GDAL Docker image being used, see [here][gdal-docker] for details.
 2. It uses the 3DCityDB Importer [`impexp import`][3dcitydb-importer] tool to initialise the 3DCityDB schema in the PostgreSQL database in the stack.
-3. It uses two SQL scripts to convert the building data uploaded by GDAL and populate the 3DCityDB tables sequentially.
-4. It uses the [`py3dtiler`][py3dtiler] tool to create 3DTile sets that can be used to visualise the newly uploaded geometries (in Cesium).
-5. It uses the GeoServer REST API to create a new layer in GeoServer that can be used to visualise the newly uploaded geometries (in Mapbox).
+3. It uses two SQL scripts to convert the building data uploaded by GDAL and populate the 3DCityDB tables sequentially. The building data will be instantiated as LoD2 buildings with thematic surfaces.
+4. Building footprints and heights are added to the uploaded data.
+5. It writes the processed data in PostgreSQL to a compressed CityGML file using the 3DCityDB Importer [`impexp import`][3dcitydb-importer] tool.
```
gpeb2, line 433:
```suggestion
5. It writes the processed data that has been uploaded to PostgreSQL out to a compressed CityGML file using the 3DCityDB Importer [`impexp import`][3dcitydb-importer] tool.
```

Deploy/stacks/dynamic/stack-data-uploader/README.md
```diff
@@ -436,23 +449,23 @@ Two SQL scripts are executed to convert the building data to conform with the 3D
...
 
 1. `"building_gmlid"`: the GML ID of the parent building of a surface
 2. `"gmlid"`: a randomly generated GML ID of each surface of each building
 3. `"class"`: an integer indicating whether a surface is a roof, wall or ground
 4. `"geom"`: the 3D polygon geometry of each surface
 
-In some cases, the original data require more sophisticated processing. User can supply a custom SQL script with `"preprocessSql"` keyword using [File by Value Name](#value-by-file-name). It should create the `"raw_building"` and `"raw_surface"` tables in the public schema from the uploaded original data. In this case, only `"IDname"` and `"IDval"` will take effect and needed to be specified in `"columnMap`.
+In some cases, the original data require more sophisticated processing. User can supply a custom SQL script with `"preprocessSql"` keyword using [File by Value Name](#value-by-file-name). It have to create the `"raw_building_XtoCityDB"` and `"raw_surface_XtoCityDB"` tables in the public schema from the uploaded original data with the column names stated above. In this case, only `"IDname"` and `"IDval"` will take effect and needed to be specified in `"columnMap`.
```
gpeb2, line 466:
```suggestion
In some cases, the original data require more sophisticated processing. Users can supply a custom SQL script with the `"preprocessSql"` keyword using [File by Value Name](#value-by-file-name). The query must create the `"raw_building_XtoCityDB"` and `"raw_surface_XtoCityDB"` tables in the public schema from the uploaded original data with the column names stated above. In this case, only `"IDname"` and `"IDval"` will take effect and needed to be specified in `"columnMap"`.
```

Deploy/stacks/dynamic/examples/datasets/inputs/config/building-cambridge.json
```diff
@@ -15,21 +15,21 @@
...
+				"IDname": "os_topo_toid",
+				"footprint": "polygon",
+				"elevation": "abshmin",
+				"height": "relh2"
+			},
+			"importOptions": {
+				"sridIn": "27700"
+			},
+			"name": "building-cambridge_gsl",
+			"geoServerSettings": {
```
sm453, line 29: Please configure whichever IDE you're using to replace tabs with spaces.

Deploy/stacks/dynamic/examples/datasets/inputs/config/building-cambridge.sql
```diff
@@ -1,31 +0,0 @@
-WITH "toid_table" AS (
-    SELECT "strval" AS "toid",
-        "cityobject_id"
-    FROM "cityobject_genericattrib"
-    WHERE "attrname" = 'os_topo_toid'
-),
```
sm453, line 6: I note that the SQL file you replaced this one with no longer makes reference to TOIDs. I'm guessing this is for good reasons?

Deploy/stacks/dynamic/examples/datasets/visualisation/building-bavaria/README.md
sm453, line 1: You may want to update the two references to the TWA-VF here (both the wording and the links/paths).

Deploy/stacks/dynamic/examples/datasets/inputs/config/building-cambridge.sql
ChungTingLao: It was removed so building-bavaria and building-cambridge can share the same SQL file for geoserver layer publication.

Deploy/stacks/dynamic/examples/datasets/inputs/config/building-cambridge.json
ChungTingLao: I have now re-formatted changed files to have spaces instead of tabs.

ChungTingLao:
I have updated the code to address raised issues, please check if you are happy with the current version.

Deploy/stacks/dynamic/examples/datasets/visualisation/building-hongkong/README.md
```diff
@@ -6,8 +6,8 @@ After uploading the Hong Kong building data to the stack, three sets of b3dm fil
 - LOD2 Features: Roof, ground and wall surfaces of LoD2 buildings are visualised.
 - LOD1-LOD2 Buildings: Buildings are shown as LoD1 when zoomed out and as LoD2 when zoomed in.
 
-They can be visualised by the [Digital Twin Visualisation Framework (DTVF) with Cesium](../../../../../../../web/digital-twin-vis-framework/example-cesium-vis).
+They can be visualised by the [Digital Twin Visualisation Framework (DTVF) with Cesium](../../../../../../../web/twa-vis-framework/example-cesium-vis).
```
sm453, line 9: Please note it is now called the TWA Visualisation Framework (TWA-VF). Two occurrences need to be updated in this file, and there are three similar files in total (one for each building set).

Deploy/stacks/dynamic/examples/datasets/visualisation/building-hongkong/README.md
```diff
@@ -6,8 +6,8 @@ After uploading the Hong Kong building data to the stack, three sets of b3dm fil
...
 - LOD1-LOD2 Buildings: Buildings are shown as LoD1 when zoomed out and as LoD2 when zoomed in.
 
-They can be visualised by the [Digital Twin Visualisation Framework (DTVF) with Cesium](../../../../../../../web/digital-twin-vis-framework/example-cesium-vis).
+They can be visualised by the [TWA Visualisation Framework (TWA-VF) with Cesium](../../../../../../../web/twa-vis-framework/example-cesium-vis).
 
-In addition, a Geoserver layer is published by the stack for visualisation in [Mapbox](../../../../../../../web/digital-twin-vis-framework/example-mapbox-vis).
+In addition, a Geoserver layer is published by the stack for visualisation in the [TWA Visualisation Framework (TWA-VF) with Mapbox](../../../../../../../web/twa-vis-framework/example-mapbox-vis).
 
 Example data.json and settings.json are provided in this folder (User will need to register with [MapTiler](https://cloud.maptiler.com/) to get an API key for settings.json; otherwise no terrain elevation in the visualisation, buildings would appear to be floating in Cesium). Please read the README.md of the DTVF for instructions of spinning up the visualisation.
```
sm453, line 13: Don't forget this occurrence of `DTVF`. ;-)

sm453:
Is it intentional that, when a CityDB dataset is included for the data uploader to upload, then _everything_ is being associated with the `citydb` schema in PostGIS (even the non-CityDB datasets that would normally have gone into the `public` schema)?

kok-foong:
This is something subtle, under the gdal options, the srid requires the full EPSG code, and the citydb data does not require it, so there is a bit of inconsistency here. 

As far as I understand "srid" should be an integer without the EPSG bit, also requiring an integer to be a string in the JSON seems a bit weird to me.

Deploy/stacks/dynamic/examples/datasets/inputs/config/building-cambridge.json
```diff
@@ -25,11 +25,11 @@
             "importOptions": {
                 "sridIn": "27700"
             },
-            "name": "cambridge_gsl",
+            "name": "building-cambridge_gsl",
             "geoServerSettings": {
                 "virtualTable": {
                     "name": "mapbox_layer",
-                    "sql": "@/inputs/config/building-cambridge.sql",
+                    "sql": "@/inputs/config/mapbox-footprint.sql",
```
kok-foong, line 32: it's not clear at first that this file needs be copied. Also a general comment that the example folder is getting crowded with the growing number of examples

kok-foong:
Looking at this method: com.cmclinnovations.stack.clients.citydb.CityDBClient.updateDatabase(String database, String sridIn).

I haven't tested the behaviour, can the database contain other geometries with different SRIDs?

kok-foong:
I am also a bit wary of the hardcoding of the citydb schema everywhere in the code + SQL scripts

kok-foong:
Not critical, but are we sure we want to use the "latest" tag for 3dcitydb/impexp, an external image? In case something breaks down the line

Deploy/stacks/dynamic/stack-clients/src/main/resources/com/cmclinnovations/stack/clients/citydb/citydb_add_building_height.sql
```diff
@@ -0,0 +1,42 @@
+SET
+    search_path TO public,
+    citydb;
+
+UPDATE
+    "building"
+SET
+    "measured_height" = "h"
```
kok-foong, line 8: is "measured_height" something custom for us? probably worth documenting, at least myself be looking for this to configure the configuration file for visualisation

Deploy/stacks/dynamic/examples/datasets/inputs/config/building-cambridge.json
ChungTingLao: It is explained in the README
https://github.com/cambridge-cares/TheWorldAvatar/blob/main/Deploy/stacks/dynamic/stack-data-uploader/README.md#value-by-file-name

ChungTingLao:
> > I will double check the behaviour of the current code and modify it to always change it back to public if that is the desirable behaviour.
> 
> Well, it should use whatever the schema of the current dataset is meant to be, i.e., if not given, it should use the default (which happens to be the public one).
> 
> As an aside, I note that the `schema` field is not currently documented in the stack data uploader documentation. It would be great if you could add this (even though this isn't the present pull request's job), mentioning its default value.

I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.

ChungTingLao:
> Looking at this method: com.cmclinnovations.stack.clients.citydb.CityDBClient.updateDatabase(String database, String sridIn).
> 
> I haven't tested the behaviour, can the database contain other geometries with different SRIDs?

No, a single SRID is global to the entire citydb schema.

sm453:
> I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.

Thanks. I have tested this and can confirm that it works as expected (i.e. the CityDB data goes into the `citydb` schema and everything else goes into `public`).

kok-foong:
> > > I will double check the behaviour of the current code and modify it to always change it back to public if that is the desirable behaviour.
> > 
> > 
> > Well, it should use whatever the schema of the current dataset is meant to be, i.e., if not given, it should use the default (which happens to be the public one).
> > As an aside, I note that the `schema` field is not currently documented in the stack data uploader documentation. It would be great if you could add this (even though this isn't the present pull request's job), mentioning its default value.
> 
> I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.

I noticed this issue still persists when I stop the uploading process of citydb halfway because it is taking too long, and start another instace of data uploader. Might be better to run this at the beginning of the upload process?

YiKaiTsai:
`geof:sfIntersects` gives SRIDs do not match error even though the SRID of the two arguments do match.

The queried tried, which returned error `it.unibz.inf.ontop.exception.OntopReformulationException: java.lang.IllegalArgumentException: SRIDs do not match: <http://www.opengis.net/def/crs/OGC/1.3/CRS84> and <http://www.opengis.net/def/crs/EPSG/0/4326>`:
```
PREFIX  geo:  <http://www.opengis.net/ont/geosparql#>
PREFIX geof: <http://www.opengis.net/def/function/geosparql/>

SELECT  ?building ?wkt
WHERE
{
BIND("<http://www.opengis.net/def/crs/EPSG/0/4326> POLYGON ((7.61078950633498 49.191732300152346, 7.61078950633498 49.19292683773855, 7.612609478547899 49.19292683773855, 7.612609478547899 49.191732300152346, 7.61078950633498 49.191732300152346))"^^geo:wktLiteral AS ?box)
    ?building  <http://www.opengis.net/citygml/building/2.0/lod0FootPrint>  ?Lod0FootPrint .
    ?geometry  <http://www.opengis.net/citygml/cityobjectgroup/2.0/parent>  ?Lod0FootPrint ;
              geo:asWKT ?wkt .
  FILTER (geof:sfIntersects(?wkt, ?box))
}
```

The following query returns both `?bc` and `?crs` as `http://www.opengis.net/def/crs/EPSG/0/4326` .
```
PREFIX  geo:  <http://www.opengis.net/ont/geosparql#>
PREFIX geof: <http://www.opengis.net/def/function/geosparql/>

SELECT  distinct (geof:getSRID(?box) as ?bc) (geof:getSRID(?wkt) as ?crs)
WHERE
{ BIND("<http://www.opengis.net/def/crs/EPSG/0/4326> POLYGON ((7.61078950633498 49.191732300152346, 7.61078950633498 49.19292683773855, 7.612609478547899 49.19292683773855, 7.612609478547899 49.191732300152346, 7.61078950633498 49.191732300152346))"^^geo:wktLiteral AS ?box)
    ?building  <http://www.opengis.net/citygml/building/2.0/lod0FootPrint>  ?Lod0FootPrint .
    ?geometry  <http://www.opengis.net/citygml/cityobjectgroup/2.0/parent>  ?Lod0FootPrint ;
               geo:asWKT ?wkt .
}
```

The relevant mappings for `?geometry` and `?wkt` inside the OBDA are
```
mappingId	Building - LoDFootprint
target		:cityobject/{"id"} bldg:lod0FootPrint :geometry/{"lod0_footprint_id"} . 
source		SELECT "id", "lod0_footprint_id"
			FROM "building"
```
and
```
mappingId	Non-Solid Geometry
target		:geometry/{"id"} a sf:Polygon ; geo:asWKT "<http://www.opengis.net/def/crs/EPSG/0/4326> {geometry_wkt}"^^geo:wktLiteral ; geo:hasMetricArea {"area_sqm"}^^www:XMLSchema#double . 
source		SELECT "id", ST_ASTEXT(ST_TRANSFORM("geometry", 4326)) AS "geometry_wkt", ST_AREA("geometry") AS "area_sqm"
			FROM "surface_geometry"
			WHERE "geometry" IS NOT NULL
```

YiKaiTsai:
> `geof:sfIntersects` gives SRIDs do not match error even though the SRID of the two arguments do match.

Thanks to the advice of @kok-foong, error is resolved by adding `<http://www.opengis.net/def/crs/EPSG/0/4326>` in front of all wktLiterals in the OBDA and not just the surface geometry.

sm453:
> > I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.
> 
> I noticed this issue still persists when I stop the uploading process of citydb halfway because it is taking too long, and start another instance of data uploader. Might be better to run this at the beginning of the upload process? Also what if something fails before that line of code?

Sounds like, in order to be robust, the uploader should reset the default schema to `public` on multiple occasions: Once at the very beginning (before uploading anything), and once after every CityDB dataset upload (ideally even if that upload has failed).

ChungTingLao:
> > > I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.
> > 
> > 
> > I noticed this issue still persists when I stop the uploading process of citydb halfway because it is taking too long, and start another instance of data uploader. Might be better to run this at the beginning of the upload process? Also what if something fails before that line of code?
> 
> Sounds like, in order to be robust, the uploader should reset the default schema to `public` on multiple occasions: Once at the very beginning (before uploading anything), and once after every CityDB dataset upload (ideally even if that upload has failed).

Resetting the schema to public at the beginning of the uploading process of any PostGIS dataset should be sufficient. If you upload once to CityDB (which changes default to citydb) and then upload another layer to anywhere else other than PostGIS, then the schema change doesn't matter anyway. If you try to upload anything to PostGIS, it is always reset to public regardless of what happened in the past.

sm453 (approved):
Tested - fine from my side, but it is probably best if all reviewers approve before merging.

kok-foong:
> > > > I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.
> > > 
> > > 
> > > I noticed this issue still persists when I stop the uploading process of citydb halfway because it is taking too long, and start another instance of data uploader. Might be better to run this at the beginning of the upload process? Also what if something fails before that line of code?
> > 
> > 
> > Sounds like, in order to be robust, the uploader should reset the default schema to `public` on multiple occasions: Once at the very beginning (before uploading anything), and once after every CityDB dataset upload (ideally even if that upload has failed).
> 
> Resetting the schema to public at the beginning of the uploading process of any PostGIS dataset should be sufficient. If you upload once to CityDB (which changes default to citydb) and then upload another layer to anywhere else other than PostGIS, then the schema change doesn't matter anyway. If you try to upload anything to PostGIS, it is always reset to public regardless of what happened in the past.

I've just tested this, I'm afraid my agent is writing to the citydb schema after running the uploader, I think you'll have to reset it to the default schema after upload completes, and if the upload fails for whatever reason.

kok-foong:
I think it might also be worth resetting the schema before adding the OBDA file, because if the mappings don't contain the schema, it succeeds at uploading it, but the queries might fail after the schema gets reset

ChungTingLao:
> > > > > I have modified the code such that the default schema is switched back to public after uploading a CityDB dataset. I checked the code again, the `schema` field is read in but not actually used (the one within `layerCreationOptions` is used and that is part of GDAL), so I will update the example ( instead.
> > > > 
> > > > 
> > > > I noticed this issue still persists when I stop the uploading process of citydb halfway because it is taking too long, and start another instance of data uploader. Might be better to run this at the beginning of the upload process? Also what if something fails before that line of code?
> > > 
> > > 
> > > Sounds like, in order to be robust, the uploader should reset the default schema to `public` on multiple occasions: Once at the very beginning (before uploading anything), and once after every CityDB dataset upload (ideally even if that upload has failed).
> > 
> > 
> > Resetting the schema to public at the beginning of the uploading process of any PostGIS dataset should be sufficient. If you upload once to CityDB (which changes default to citydb) and then upload another layer to anywhere else other than PostGIS, then the schema change doesn't matter anyway. If you try to upload anything to PostGIS, it is always reset to public regardless of what happened in the past.
> 
> I've just tested this, I'm afraid my agent is writing to the citydb schema after running the uploader, I think you'll have to reset it to the default schema after the upload completes, and if the upload fails for whatever reason.


> I think it might also be worth resetting the schema before adding the OBDA file, because if the mappings don't contain the schema, it succeeds at uploading it, but the queries might fail after the schema gets reset

Now the uploader resets the schema before and after uploading to Postgres. The only case where it may get stuck in citydb would be when you upload CityDB but fail, then run an agent without re-uploading anything else.

Deploy/stacks/dynamic/examples/datasets/inputs/config/mapbox-footprint.sql
```diff
@@ -0,0 +1,31 @@
...
+        "urival" AS "iri",
+        "cityobject_id"
+    FROM
+        "cityobject_genericattrib"
+    WHERE
+        "attrname" = 'iri'
+)
+SELECT
+    "building"."id" AS "building_id",
+    COALESCE("measured_height", 100.0) AS "building_height",
```
kok-foong, line 21: building_height is non-standard? this is the string that needs to be configured in the visualisation file

kok-foong:
In terms of uploading data this works fine.

I also noticed one more issue, not sure if you want to address this in this PR or a separate PR.  The IRI in mapbox-footprint.sql should match the IRI in the ontop mappings and I don't think it is currently the case.

Deploy/stacks/dynamic/examples/datasets/inputs/config/mapbox-footprint.sql
ChungTingLao: Correct, measured_height would be CityGML standard (although we calculate it instead of measuring it). Since users need to provide their own SQL file for geoserver publication anyway, I don't think this is a problem - they could rename it to anything they want.

ChungTingLao:
> In terms of uploading data this works fine.
> 
> I also noticed one more issue, not sure if you want to address this in this PR or a separate PR. The IRI in mapbox-footprint.sql should match the IRI in the ontop mappings and I don't think it is currently the case.

Please submit an issue for this (assign me to it?), and I will update the examples once we have a stable version of ontop mapping for CityDB data.

YiKaiTsai (approved):
Tested, seems fine. Just one comment.


Might be a good idea to add the OBDA that is going to be used for the Pirmasens stack as one of the examples. Please correct me if the OBDA already exists in the examples.

ChungTingLao:
> Tested, seems fine. Just one comment.
> 
> Might be a good idea to add the OBDA that is going to be used for the Pirmasens stack as one of the examples. Please correct me if the OBDA already exists in the examples.

It is not in any example yet, and it is a deliberate choice for now. I suggest adding it only after we have confirmed that the OBDA can work with all agents for the Pirmasens stack.

gpeb2 (approved):

PR #966, opened 2023-10-16
1659 dev isochroneagent

ShinZert:
- [x] Unable to carry out population mapping due to raster error in new version of stack-manager 
Issue is coming out from stack-data-uploader's GDALClient - George is aware and in progress of fixing it. 
- [ ] Make updates to SPARQL query pending OSMAgent update


sm453 (approved):
Tested. Was able to build and spin up the image, and the curl request for one of the examples seemed to complete successfully.

PR #967, opened 2023-10-16
CityDB to LOD1 improvements

gpeb2:
Fix issues with and optimise the citydb_add_footprint.sql script.


YiKaiTsai (approved):
Everthing works for me, and the data uploaded worked with the CEA agent. Just one comment, it was not immediately clear where to find `mapbox-footprint.sql`. Since `mapbox-footprint.sql` would be used whenever we upload new building data, it might be a good idea to highlight where to find it and/or move it to a more noticeable location.

yjytt59:
I tried to upload SG LoD3 data, but it stuck without error and crash. I have to stop data uploader, the data uploading unsuccessful.

yjytt59:
I did 2nd test on SG Lod3 data, it successful. All buildings (96,140) are imported to DB with footprint and height. It got geoserver layer and 3d tiles too.

yjytt59 (approved):
All goods for SG LoD3 data.

PR #968, opened 2023-10-17
dev-Vision: develop CV for IoT application

jlai2021:


jlai2021:
> I can build and spin up the container locally, basic functions are working.
> 
> Blocking issues:
> 
> * Cannot substitute image easily as it is part of the property. Changing the property file inside the container does not work, have to be re-started. Consider making the image source part of the HTTP request.
> * Cannot handle image source via an URL.
> * Should have a docker image on Github that can be pulled and used with different settings without re-building the docker image.
> 
> Nice to have:
> 
> * Sometimes the green box and the label is outside the output image (invisible). It would be nice to be able to always see the label and the box.
> * Currently /update only reports the best match. It would be helpful to also output other close matches to help the training process.
> * Currently there is a hard-coded confidence cut-off of 0.5. It could be made into a user-specified parameter. In addition, the output is empty if all matches have confidence below 0.5. It is probably better to always report at least the best match regardless of how low the confidence is.
> * Currently only images are supported, not videos. Explicit exception messages would be helpful in case users try to apply the agent on videos.

Thank you for your feedback on the Vision Agent. I appreciate the time you've taken to build, test, and review the Docker implementation. Your insights are crucial to enhancing functionality and usability. I have taken note of the blocking issues and agree to proceed to improve the overall experience. Regarding the "nice to have" features you suggested, I find they are highly constructive. I will also try to implement it, if possible, in this pull request or future update.

ChungTingLao:
I can pull the docker image and spin it up with my local images and config files. This can be merged once the following issues are addressed:

- add a docker-compose-build.yml for building the docker image locally. Update README accordingly.
- update the README with (1) instructions to run the agent locally without docker (2) existing online resources for the YOLO model files (if there is any).

ChungTingLao (approved):

PR #973, opened 2023-10-20
Dev aermod vis

kok-foong:
See updated instructions in JPS_VIRTUALSENSOR/README.md on spinning things up. 

Key changes:
1) Ability to select dispersion simulation, timestep, pollutant, and height in the visualisation
2) Querying for vis data is completely separated out from AERMOD agent
3) data.json for vis is dynamically queried
4) Elevation preprocessor is removed (aermap), elevation data is queried directly from PostGIS and the critical height scale is set to be equal to the elevation value (no observable differences from using aermap)
5) Query buildings from ontop (new citydb workflow)


kok-foong:
Visualisation and FIA queries are moved to stack manager inputs

kok-foong:
I can add more details on the requirements for the instantiation of static point sources. 

The main difficulty with making a working example with static point source is that the input agent needs to instantiate a static point source instance that points to a building IRI (that can change without the previous file).
- The "proper" setup for the Pirmasens case is available in the private pirmasens repo
- The Jurong Island case is not using the new buildings workflow and relies on connecting to an endpoint in CARES, so may not be reliable sometimes

ChungTingLao (approved):
I have checked that the virtual sensor in visualisation works. The static point source example can be sorted out in a later branch. Functionality-wise it is working as expected.

PR #988, opened 2023-10-26
987 add language tag support to the tbox csv generator

cmclscourtney:
Small changes to TBox Generator
- Added handling of language tags on strings in comments and labels. In excel can input `"my string"@en` but due to csv standards will have to put `"""my string"@en""` if open with text editor
- Fixed bug that `rdfs:isDefinedBy` points to a string of a URL rather than the URL itself
- Minor refactoring of private methods to make maintenance a little easier

### DIFF
--- a/JPS_BASE_LIB/pom.xml
+++ b/JPS_BASE_LIB/pom.xml
@@ -8,7 +8,7 @@
 
     <!-- Please refer to the Versioning page on TheWorldAvatar wiki for
     details on how version numbers should be selected -->
-    <version>1.40.0</version>
+    <version>1.40.1-987-SNAPSHOT</version>
 
     <!-- Project Properties -->
     <properties>
--- a/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/converter/TBoxManagement.java
+++ b/JPS_BASE_LIB/src/main/java/uk/ac/cam/cares/jps/base/converter/TBoxManagement.java
@@ -18,6 +18,7 @@ import org.semanticweb.owlapi.model.AddOntologyAnnotation;
 import org.semanticweb.owlapi.model.IRI;
 import org.semanticweb.owlapi.model.OWLAnnotation;
 import org.semanticweb.owlapi.model.OWLAnnotationProperty;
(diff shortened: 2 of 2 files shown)


cmclscourtney:
Need to change tests to align with these changes

msff2:
> Need to change tests to align with these changes

Current changes produce an extended ontological model, so the previous tests fail. That means we have to update the reference ontological model.

cmclscourtney:
> > Need to change tests to align with these changes
> 
> Current changes produce an extended ontological model, so the previous tests fail. That means we have to update the reference ontological model.

Made the necessary changes and now tests are running as intended

sm453 (approved):

msff2:
Refine the list of parameters in the comments in all the methods where you have removed the range. For example, the following comment still has a range in it.

/**
	 * Decides the type of logical formula needs to be created.</br>
	 * Currently, it supports formulae with the following two quantifiers:<br>
	 * - for all (only)<br>
	 * - for some (some)<br>
	 * - exactly 1<br>
	 * - maximum 1<br>
	 * - minimum 1<br>
	 * 
	 * @param objectProperty
	 * @param classExpression
	 * @param quantifier
	 * @param singleDomain
	 * @param range
	 * @throws JPSRuntimeException
	 */
	private void decideToAddTypeOfLogicalFormula(OWLObjectProperty objectProperty, OWLClassExpression classExpression,
			String quantifier, String singleDomain) throws JPSRuntimeException {

cmclscourtney:
> Refine the list of parameters in the comments in all the methods where you have removed the range. For example, the following comment still has a range in it.
> 
> /** * Decides the type of logical formula needs to be created. * Currently, it supports formulae with the following two quantifiers: * - for all (only) * - for some (some) * - exactly 1 * - maximum 1 * - minimum 1 * * @param objectProperty * @param classExpression * @param quantifier * @param singleDomain * @param range * @throws JPSRuntimeException */ private void decideToAddTypeOfLogicalFormula(OWLObjectProperty objectProperty, OWLClassExpression classExpression, String quantifier, String singleDomain) throws JPSRuntimeException {

Good catch! Addressed in 7c35ea2

PR #989, opened 2023-10-27
Fixing the jps_base_lib tests

gpeb2:
There are many issues with the jps_base_lib tests. This PR is intended to fix the following:

- The tests that use JUnit5 don't seem to get detected when the test are run directly through mvn.
- Some of the JUnit5 based tests don't work on WSL/Linux systems.
- The Maven dependency directory isn't cached in the GitHub test action, meaning that all of the dependencies need to be redownloaded even if they haven't changed.

Other improvements include:

- Disabled some tests that are not good and cause issues.
- Improved/standerdised the handling of Blazegraph containers used for tests.
- Re-enabled some previously disabled tests after shortening their runtime.
- Moved tests out of ```test``` package directories.
- Updated some of the dependencies to versions that don't have known vulnerabilities.
- Completely excluded the ```fed``` code as it is just a proof of concept and requires a dependency that has known vulnerabilities.


mdhillman (approved):
With the understanding that we are at the beginning of the long process to improve our tests, I can't see any major issues here. However, we do need to capture this learning (i.e. why we've made these changes, how to write better tests) in our training documentation.

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/derivation/DerivationClientIntegrationTest.java
```diff
@@ -9,265 +9,258 @@
...
 
 /**
- * NOTE: all tests provided in this class are stress tests. It is recommended to run one test at a time.
- * Or you can run testCleanUpFinishedDerivationUpdateStressTest by itself first, then comment out it and
- * run the rest altogether.
+ * NOTE: all tests provided in this class are stress tests. It is recommended to
+ * run one test at a time.
+ * Or you can run testCleanUpFinishedDerivationUpdateStressTest by itself first,
+ * then comment out it and run the rest altogether.
  */
```
cmclscourtney, line 29: Likely need to clarify this and align with disabling below

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/query/StoreRouterTest.java
```diff
@@ -13,158 +16,161 @@
...
-		assertFalse(StoreRouter.isFileBasedTargetResourceID("kb/sgp:singapore/SGTemperatureSensor-001.owl"));
-		assertFalse(StoreRouter.isFileBasedTargetResourceID("kb/sgp?singapore/SGTemperatureSensor-001.owl"));
-		assertFalse(StoreRouter.isFileBasedTargetResourceID("kb/sgp*singapore/SGTemperatureSensor-001.owl"));
+
+		// Invalid file paths
+		// These are legal file paths in Linux
+		// assertFalse(StoreRouter.isFileBasedTargetResourceID("kb/sgp:singapore/SGTemperatureSensor-001.owl"));
+		// assertFalse(StoreRouter.isFileBasedTargetResourceID("kb/sgp?singapore/SGTemperatureSensor-001.owl"));
+		// assertFalse(StoreRouter.isFileBasedTargetResourceID("kb/sgp*singapore/SGTemperatureSensor-001.owl"));
 	}
```
cmclscourtney, line 172: Should these just be deleted?

cmclscourtney (commented):
Agree with Michael above. Perhaps should create issue to do a similar thing with the python tests.

gpeb2:
> With the understanding that we are at the beginning of the long process to improve our tests, I can't see any major issues here. However, we do need to capture this learning (i.e. why we've made these changes, how to write better tests) in our training documentation.

I agree, this PR was specifically intended to just get all of the tests working. I'll create another one regarding training material and link to this one so that it can be used to generate examples.

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/derivation/DerivationClientIntegrationTest.java
gpeb2: All of these comment were pre-existing so I don't feel able to clarify them.

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/query/StoreRouterTest.java
gpeb2: I kept them as a sort of documentation of the method being tested. I considered just changing the expected value based on the operating system but decided that that might not be reliable.

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/derivation/DerivationClientIntegrationTest.java
sm453: Perhaps @jb2197 can comment.

gpeb2:
> Agree with Michael above. Perhaps should create issue to do a similar thing with the python tests.

I have now run the python tests and they run fine.

JPS_BASE_LIB/src/test/java/uk/ac/cam/cares/jps/base/derivation/DerivationClientIntegrationTest.java
gpeb2: I'm going to leave this as it is for the PR as I haven't changed which tests from this class are run.

cmclscourtney (approved):

PR #990, opened 2023-10-30
1677-dev-add-concepts-to-ontoems: Added the following concepts to ont…

Abyssus1010:
…oems

Wind chill, heat index and precipitation rate


JPS_Ontology/ontology/ontoems/OntoEnvironmentalMeasurementStation.csv
```diff
@@ -37,6 +39,7 @@ Visibility,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Dista
 UVIndex,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Number,IS-A,,,,The strength of the sun's ultraviolet (UV) radiation expressed as index as developed by the World Health Organization.,https://www.theworldavatar.com/kg/ontoems/,UV index
 PrecipitationProbability,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Number,IS-A,,,,The likelihood of occurrence of precipitation.,https://www.theworldavatar.com/kg/ontoems/,Precipitation probability
 PrecipitationProbability,Class,http://semanticscience.org/resource/SIO_000638,IS-A,,,,The likelihood of occurrence of precipitation.,https://www.theworldavatar.com/kg/ontoems/,Precipitation probability
+PrecipitationRate,Class,http://www.ontology-of-units-of-measure.org/resource/om-2/Speed,IS-A,,,,The amount of rain that has fallen per unit time (usually given in units of mm/hr or in/hr (millimeters per hour or inches per hour)) .,https://www.theworldavatar.com/kg/ontoems/,Precipitation rate
```
markushofmeister, line 42: Is there any better match for this? Speed seems correct wrt the given units; however, isn't it actually a volume per time unit and only "reduced" to length per time unit for standardised tippping bucket etc. measurement devices?

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -1054,7 +1071,6 @@
 
 
     <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/RelativeHumidity">
-        <owl:equivalentClass rdf:resource="https://www.auto.tuwien.ac.at/downloads/thinkhome/ontology/WeatherOntology.owl#Humidity"/>
```
markushofmeister, line 1057: Where does this change come from? Doesn't seem to stem from the csv file...

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -792,7 +786,6 @@
 
 
     <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/DewPoint">
-        <owl:equivalentClass rdf:resource="https://www.auto.tuwien.ac.at/downloads/thinkhome/ontology/WeatherOntology.owl#DewPointTemperature"/>
```
markushofmeister, line 795: Why is this one disappearing?

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -743,13 +738,12 @@
 
 
     <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/AtmosphericPressure">
-        <owl:equivalentClass rdf:resource="https://www.auto.tuwien.ac.at/downloads/thinkhome/ontology/WeatherOntology.owl#AtmosphericPressure"/>
```
markushofmeister, line 746: Why is this one disappearing?

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -730,7 +726,6 @@
 
 
     <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/AirTemperature">
-        <owl:equivalentClass rdf:resource="https://www.auto.tuwien.ac.at/downloads/thinkhome/ontology/WeatherOntology.owl#Temperature"/>
```
markushofmeister, line 733: Why is this one disappearing?

markushofmeister (changes requested):
There seem to be some changes to the owl file which are not directly reflected in the updated csv file. It would be good to understand where they stem from and to correct it.

JPS_Ontology/ontology/ontoems/OntoEnvironmentalMeasurementStation.csv
Abyssus1010: Hmmm the reason for adding precipitation rate as a subclass of om:Speed is because I need to represent this weather station that reports precipitation rate values with mm/hr as the unit. And since mm/hr is a speed unit, it is only applicable to om:Speed and it's subclasses.

JPS_Ontology/ontology/ontoems/OntoEMS.owl
Abyssus1010: I think this is because I was using a newer version of the converter.

Abyssus1010:
I pulled the latest version of the converter and used to generate the owl file again. Only rdfs:isDefinedBy is affected.

JPS_Ontology/ontology/ontoems/OntoEnvironmentalMeasurementStation.csv
Abyssus1010: Or will it be better to rename the concept as rainfall rate instead of precipitation rate?

markushofmeister: If there is nothing more specific available in om, I believe `PrecipitationRate` and Speed is fine

markushofmeister (changes requested):
The new converter seems to change the `DefinedBy` content; however, that still seems alright to me (just different type declaration). Moreover, it introduces some odd characters for PM2.5 and PM10.

Maybe it would be easiest to fix those encoding issues as well as the equivalent class definitions manually (or get in touch with Feroz to understand their root cause)

JPS_Ontology/ontology/ontoems/OntoEMS.owl
```diff
@@ -998,8 +998,8 @@
 
     <owl:Class rdf:about="https://www.theworldavatar.com/kg/ontoems/PM2.5Concentration">
         <rdfs:subClassOf rdf:resource="https://www.theworldavatar.com/kg/ontoems/ParticulateMatterConcentration"/>
-        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The concentration of particulate matter &lt; 2.5 µm in the ambient air.</rdfs:comment>
-        <rdfs:isDefinedBy rdf:datatype="http://www.w3.org/2001/XMLSchema#string">https://www.theworldavatar.com/kg/ontoems/</rdfs:isDefinedBy>
+        <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">The concentration of particulate matter &lt; 2.5 Âµm in the ambient air.</rdfs:comment>
```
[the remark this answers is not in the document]
Abyssus1010: Oh no, I did not noticed that. Thanks for pointing it out. I think it will be faster if I manually fix them.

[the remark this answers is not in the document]
markushofmeister: > Oh no, I did not noticed that. Thanks for pointing it out. I think it will be faster if I manually fix them.

Sounds good. Although I am not convinced that all equivalent classes are added back by now, e.g. `PrecipitationProbability` seems still missing (and potentially more). Please have a double check before re-requesting final review. Thanks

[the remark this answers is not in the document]
Abyssus1010: Oh but I compared the edited file with the one in main and there's no equivalent classes for PrecipitationProbability or any others. Let me compare with the csv file. They were probably not added in when the owl file was last modified.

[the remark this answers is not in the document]
Abyssus1010: I have added the missing equivalent classes definitions which were also missing from main. I compared the csv file with the owl file and all equivalent classes definitions should be accounted for.

markushofmeister (approved):

PR #993, opened 2023-11-01
1676 dev fix isochroneagent input srid

ShinZert:
1) Solved input SRID 0 problem by making 4326 assumption when querying from SPARQL 
2) Removed prefix from iri and add it as prefix within isochrone.obda


kok-foong (approved):

sm453 (approved):
Looks ok, although I have not tested it.

PR #995, opened 2023-11-01
dev-change-cea-properties-path: read properties file from resources

kok-foong:
This should simplify the Pirmasens stack startup process.
Please check if the code still works. In the future, it will be much easier to have the endpoints specified in environment variables rather than reading them from a properties file.


kok-foong:
Once checked, you can increment the version number and push a new image

YiKaiTsai (approved):

YiKaiTsai:
Builds and runs. I'm fine with the changes. Would just like to check with @sm453 that he is fine with this, since he will be the one spinning up the Pirmasens stack.

kok-foong:
I can see this is useful for things like visualisation javascript files, but in most cases the properties file is only read once and the container still needs restarting to read in the new properties. For now we'll rely on bind mounts.

PR #1003, opened 2023-11-07
1687 hotfix fix isochroneagent obda

ShinZert:
1) Fix obda
2) Add name to isochrone, building geoserver


kok-foong (approved):
This is fine. By Java convention, constants should be written in screaming snake case, recommend to install sonarlint that will clean up the code.

PR #1009, opened 2023-11-09
1689 dev update fhsashandoccupancyagent to get the latest sash opening

sandradeng20:
Add /latestsash endpoint to retrieve the latest sash opening of the provided fumehood iri.

- No test case added for this endpoint, because it is built with existing functions and the functions are tested with previous test cases.
- Pass manual test with local stack.


Abyssus1010:
Unit tests ran successfully and tested agent in stack. Original and new functions works properly. README looks good as well.

Abyssus1010 (approved):

Issue #1012, opened 2023-11-10
FIA caches enforcedEndpoint


kok-foong:
In the QueryManager class, the value of enforcedEndpoint should be not be stored in memory. This is because when it first detects an enforced endpoint and saves the value, subsequent calls to determineClassMatches uses the previously cached value here:

`return classHandler.determineClassMatches(iri, this.enforcedEndpoint);`

PR #1015, opened 2023-11-13
1012 fia caches enforcedendpoint

cmclscourtney:
See #1012 for bug fixed.

In effort to work through the code and play around also added some small bits of Jackson serialization. I think the `Optional` is useful for the endpoint specification and will make maintenance in the future easier.


mdhillman (approved):
Changes all look good to me.

cmclscourtney:
@kok-foong just wondering if this resolves #1012?

kok-foong:
> @kok-foong just wondering if this resolves #1012?

is 3.0.1-SNAPSHOT the version to test? It seems that it's still caching the value to me

kok-foong:
> > @kok-foong just wondering if this resolves #1012?
> 
> is 3.0.1-SNAPSHOT the version to test? It seems that it's still caching the value to me

I suspect the latest image does not correspond to the latest code

cmclscourtney:
> > > @kok-foong just wondering if this resolves #1012?
> > 
> > 
> > is 3.0.1-SNAPSHOT the version to test? It seems that it's still caching the value to me
> 
> I suspect the latest image does not correspond to the latest code

The version is `3.0.1-1012-SNAPSHOT`

kok-foong (approved):
Tested and can confirm FIA no longer caches the endpoint

PR #1016, opened 2023-11-14
1688 dev isochroneagent upgrade kingslynn flood

ShinZert:
- Create isochrone for kingslynn, fix segmentization route problem
- Create unreachable area population


Ushcode (commented):
Isochrone agent runs to produce isochrones. Worked with your osm file after ~10 mins but using the pbf I downloaded which covers more area, the query did not complete after about 20 minutes

Ushcode (approved):

Ushcode:
Due to the length of time taken for the query to run, I am worried that this will not be scalable

PR #1019, opened 2023-11-17
1691 dev travelling salesman agent

ShinZert:
1) Query points of interest from the knowledge graph 
2) Create layer for kingslynn travelling salesman agent automatically
3) Take marker coordinate node as inputs

This agent developed is use case specific with a layer of abstraction that allows the flexibility of easy modification for other scenarios.


Agents/TravellingSalesmanAgent/travellingsalesmanagent/src/main/java/uk/ac/cam/cares/jps/agent/travellingsalesmanagent/TSPRouteGenerator.java
```diff
@@ -0,0 +1,120 @@
...
+                "JOIN pgr_dijkstra(\n" +
+                "    '"+cost_table+"',\n" +
+                "    n1.node,\n" +
+                "    n2.node, false\n" +
+                ") AS di ON true\n" +
+                "JOIN routing_ways ON di.edge = routing_ways.gid\n" +
+                "ORDER BY n1.seq\n" +
+                ")\n" +
+                "SELECT tsp_route.the_geom as geom\n" +
+                "FROM tsp_route ";
```
Ushcode, line 57: Think in general it would be better to put the SQL queries in a different file in the resources folder but probably no need to change it here

Ushcode (approved):

Ushcode:
Tested and working as expected

Agents/TravellingSalesmanAgent/travellingsalesmanagent/src/main/java/uk/ac/cam/cares/jps/agent/travellingsalesmanagent/TSPRouteGenerator.java
ShinZert: That is a great suggestion and good point. The SQL takes input -"cost_table" as parameter, maybe it would be good to use a SQL builder for it to look cleaner instead of using resources folder. Will make a note and refractor it on the next update! Thanks!

PR #1039, opened 2023-12-05
Dashboard Agent

qhouyee:
This agent interacts with the Grafana service in the stack to set up a dashboard interface, populated with time series measures from the knowledge graph. When called again, the agent will overwrite the existing dashboard with the latest data.

At the moment, the agent can set up dashboards for devices, rooms, and systems within facilities.


Agents/DashboardAgent/agent/src/main/java/uk/ac/cam/cares/jps/agent/dashboard/DashboardAgent.java
```diff
@@ -0,0 +1,189 @@
...
+            // Ensure that the agent is running on a stack by initialising the services
+            SERVICES = new StackClient();
+        } catch (ServletException exception) {
+            // This error only occurs when super.init() fails
+            DashboardAgent.VALID = false;
+            LOGGER.error("Could not initialise an agent instance!", exception);
+        } catch (Exception exception) {
+            // This error will only occur if the agent is not running on a stack, which causes an error in Stack Client
+            DashboardAgent.VALID = false;
+            LOGGER.error("Agent is not running on a stack and will not be initialised!", exception);
```
gpeb2, line 51: See comment above regarding the `VALID` field.
The exceptions should be rethrown or new exceptions that wrap them should be thrown as the agent is in an invalid state.
```suggestion
            DashboardAgent.valid = true;
        } catch (ServletException exception) {
            // This error only occurs when super.init() fails
            LOGGER.error("Could not initialise an agent instance!", exception);
            // Add code to rethrow existing exception or throw a new one that wraps the current one
        } catch (Exception exception) {
            // This error will only occur if the agent is not running on a stack, which causes an error in Stack Client
            LOGGER.error("Agent is not running on a stack and will not be initialised!", exception);
            // Add code to rethrow existing exception or throw a new one that wraps the current one
```

Agents/DashboardAgent/agent/src/main/java/uk/ac/cam/cares/jps/agent/dashboard/DashboardAgent.java
```diff
@@ -0,0 +1,189 @@
...
+    protected JSONObject resetRoute() {
+        JSONObject response = new JSONObject();
+        if (DashboardAgent.VALID) {
+            SERVICES = new StackClient();
+            response.put("Result", "Agent has been successfully reset!");
+        } else {
+            response.put("Error", "Agent could not be initialised! Please check logs for more information...");
+        }
+        return response;
+    }
```
gpeb2, line 188: What is this useful for? As far as I can tell if `VALID` is `true` then `SERVICES` should also be initialised.

Agents/DashboardAgent/agent/src/main/java/uk/ac/cam/cares/jps/agent/dashboard/DashboardAgent.java
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * This class acts as the entry point of the compiled war.
+ *
+ * @author qhouyee
+ */
+@WebServlet(urlPatterns = {"/reset", "/status", "/setup"})
+public class DashboardAgent extends JPSAgent {
+    private static final Logger LOGGER = LogManager.getLogger(DashboardAgent.class);
+    // Agent starts off in valid state, and will be invalid when running into exceptions
+    private static boolean VALID = true;
```
gpeb2, line 24: The Java naming convention is that only `static final` variables (global constants) are all caps.
It is more correct to initialise `valid` to `false` as it isn't valid until `init` has been run.
```suggestion
    // Agent starts off in invalid state, and will become valid when initialised without exceptions
    private static boolean valid = false;
```

gpeb2 (commented):
I've made some explicit comments, mainly in the Agent class code.
There are other minor issues that it would be easier to discuss in a call.

gpeb2 (approved):
Changes look good

PR #1040, opened 2023-12-05
Fixes to stack code for running with Podman

gpeb2:
Cherry-picked changes that fix issues with running stacks using Podman.


sm453 (approved):
Tested with (rootless) podman only - works. The stack now spins up without errors.

PR #1042, opened 2023-12-11
Dev add parallelisation option

kok-foong:
Key changes:
1) Made it optional to parallelise emission sources derivation update via the parameter PARALLELISE_EMISSIONS_UPDATE
2) Corrected error in calculating diameter of static point source when srid is not in meters
3) Changed the way stack exit velocity is calculated, it is assumed that the total emissions mass flow rate is a fraction of the total air flowrate, and the fraction is set via TARGET_EMISSION_VOLUME_FRACTION


markushofmeister (approved):
Tested, works. Results are now much more plausible than before for district heating

PR #1046, opened 2023-12-13
Dev fix virtualsensor height

kok-foong:
After implementing multiple height simulations, the virtual sensor was randomly picking the layer based on the sequence of the SPARQL query results.

Now it will always pick the lowest layer (e.g. ground level, 0 m)


markushofmeister (approved):
Looks good to me. I have not tested this version yet, but tested version `1.4-dev-fix-virtualsensor-height-SNAPSHOT` successfully yesterday.

PR #1047, opened 2023-12-14
Dev hotfix dashboard agent

qhouyee:
At the moment, the agent is unable to detect the postgreSQL driver required and cannot be run. A solution is to refactor the PostGisClient to use RemoteRDBStoreClient, so that this connection is abstracted away to the JPS base lib


gpeb2 (approved):
I haven't tried to run this but it looks ok so if you're happy it runs then feel free to merge. Just remember to change the version numbers to a release ones (remove "-SNAPSHOT").

Issue #566, opened 2023-01-24
Add dimensions to geoserver in stack data uploader


cmclscourtney:
Add ability to specify and configure dimensions in stack data uploader

PR #1048, opened 2023-12-14
Dev-upload-timeseries-heat

Ushcode:
*This PR provides functionality to upload netCDF natively into postGIS and GeoServer. This is done by using `gdal` to check the filetype, and if it is `netCDF` then to copy straight into geotiffs volume. netCDF plugin is now added to GeoServer by default

* Separately, this PR adds functionality to, if a projection is unrecognised by `gdal`, add it as a custom projection using the value of sridOut in the config file as the SRID and Authority (e.g. "sridOut": "TWA:101916") 

* This PR now provides all functionality requested in #566 by @cmclscourtney


gpeb2 (changes requested):
This is just a first look

gpeb2 (approved):
@Ushcode, please check that it still runs as expected after the changes I've made.
Once that is done it should be ready to merge.
Please remember to update the version numbers before merging though.

PR #1072, opened 2024-01-15
Mapbox Control elements

Ushcode:
This PR should close issues 1054, 914, and 911

# TWA-VF 4.5.0
Three control buttons are now included by default. These are:

* Scale Control (bottom right indicator of current map scale in metres and kilometres)
* Geolocator (a button in the bottom right to display current location of the user)
* Zoom, pitch and compass controls in the top right
  
  Ability to turn off with flags in `settings.json`.

Documentation and changelog.md updated


mdhillman (approved):
Reviewed with @Ushcode, changes look good.

PR #1086, opened 2024-01-25
dev-marie-combustion-symposium: Develop KGQA over OntoSpecies, OntoKin, and OntoCompChem

picas9dan:


sm453 (approved):
Ok, fine. Unfortunately, the way you've done it doesn't allow one to trace the revision history of a file in the git log. It would have been nicer to filter-branch the original repositories to relocate the files like I suggested before merging. Never mind...

PR #1088, opened 2024-01-26
Dev live ship data

kok-foong:
Added functionality to use live data from aistream.io, instructions are in the main README


jlai2021:
The README.md in virtual sensor should be modified. 

It should begin with an overall of this agent does, and key features.

ChungTingLao (changes requested):
I have checked that the virtual sensor is working as expected for both live ship data and static ship data. However, the virtual sensors created in the visualisation does not report pollutant concentrations.

As a side note, it would be nice to remove obsolete variables that is no longer used from ALL files e.g. GEOSERVER_URL in AermodAgent/docker-compose.yml.

ChungTingLao (approved):

PR #1091, opened 2024-01-30
No pull Podman/Docker

gpeb2:
Fixed the coded that checks for existing images so that it doesn't try to pull an image if it finds an appropriate copy locally.


sm453 (commented):
What's the relationship of this to #735? Is #735 now obsolete?

gpeb2:
> What's the relationship of this to #735? Is #735 now obsolete?

It covers the bug but not the ability to specify credentials.

sm453 (approved):

PR #1092, opened 2024-02-01
Add anonymised Pirmasens district heating stack instructions to TWA

markushofmeister:
Moved instructions to spin up district heating stack from private Pirmasens repo to TWA. Replaced references to confidential data with placeholders/examples and removed unnecessary agents


kok-foong:
The stack name is an environment variable that should be visible in each container started up by the stack manager, I can modify my containers in a separate PR and then only merge this in once that is done unless this is urgent

markushofmeister:
I believe this entire discussion is unnecessary work creation. My point of view:
1) If we actually want to refine the stack name settings, we should do it in the `Pirmasens` repository and not here
2) The purpose of this PR is to make relevant code, required to inspect/reproduce the work of the paper, publically available.

Hence, I believe a hard coded stack name does not do any harm here

kok-foong (approved):
All uploader configs should be needed as they are referred to by the generated data.json by the dispersion interactor

sm453:
> I believe this entire discussion is unnecessary work creation. My point of view:

Well, now that this can of worms is wide open, I would have to say that I _strongly_ disagree:

> 1. If we actually want to refine the stack name settings, we should do it in the `Pirmasens` repository and not here
> 2. The purpose of this PR is to make relevant code, required to inspect/reproduce the work of the paper, publically available.

No, the purpose of this PR is not to dump some (bad) code for a paper, but to move the code where it should be. The `pirmasens` repository shouldn't even exist. Any future development should be done within TWA repository.

> Hence, I believe a hard coded stack name does not do any harm here

This practice is appalling, sets a very bad example, and people _will_ copy/imitate it. I would rather contain this cancer before it spreads.

Having said all that, I do recognise the need to get on with it, which is why I approved the PR with my initial comment. I have added issue #1093 for the record.

PR #1104, opened 2024-02-12
Dev-op-example-viz-hotfix

Ushcode:
This PR fixes issue where some sidebar contents do not stack vertically by changing the display to a flex column.

To test this functionality, pull the `ghcr.io/cambridge-cares/twa-vf:4.5.2-sidebarcontents-SNAPSHOT` image and include it in the viz config.

also run the example stack in `TheWorldAvatar/web/twa-vis-framework/example` by running `/.start.sh PASSWORD=<your password to set for geoserver and postgis> but be aware that this will remove and any inputs in the manager and uploader data directories, so save them somewhere else or perform this step in a fresh clone


sm453 (approved):
Tested, works. Now the side panel looks as expected (on a 4K laptop screen).

PR #1109, opened 2024-02-14
dev heat data vrt addition

Ushcode:
- add VRT raster driver support and create VRT for NETCDF files
- snapshot version number


cmclscourtney (approved):
Tested in CReDo and worked well. Not sure if @gpeb2 wants to have a look at the code...

Ushcode (commented):
Further two commits address the unholy 360 day year data by skipping the cast of time dimension in the `.nc` file to an ISO time and instead leave it as a string in the `<layer>_times` table in postgis (still as field `time`, along with the only other field `bands`). 

GeoServer now parses time data as a time if it is a time but as integer band number if it is not cast

Ushcode:
@gpeb2 are you happy for me to merge?

gpeb2 (approved):

PR #1115, opened 2024-02-19
Allow datasets to specify a schema.

gpeb2:
Allow the user to upload a dataset to a PostgreSQL schema other than "public".


cmclscourtney (approved):
Reviewed changes and tested. Works well. Only concern for future is if there are two tables of the same name in different schemas, what should the geoserver layer names be...

PR #1122, opened 2024-02-26
Update and fix JPS base lib dependencies

gpeb2:
There was a mismatch in dependencies that was preventing FedX based federated queries working in some situations.
Also tidied up the dependencies of the base lib and stack projects so that all 3rd -party dependency versions come from the parent pom.


cmclscourtney (approved):
Tested with latest CReDo agents and worked well

PR #1126, opened 2024-02-28
1767 dev add concepts to ontochemplant and ontodevice

Abyssus1010:


JPS_Ontology/ontology/ontodevice/OntoDevice.csv
```diff
@@ -172,4 +172,5 @@ hasMaxThreshold,Object Property,,,https://w3id.org/bot#Zone,Threshold,,A relatio
 hasCumulativeEnergyConsumption,Object Property,hasQuantity,IS-A,http://www.theworldavatar.com/ontology/ontocape/chemical_process_system/CPS_function/process.owl#TemperatureChange UNION Device,http://www.ontology-of-units-of-measure.org/resource/om-2/Energy,,A relationship that describes the amount of cumulative energy used or consumed during a process or by a device.,https://www.theworldavatar.com/kg/ontodevice/,has cumulative energy consumption
 hasMode,Object Property,,,Device,Mode,,A relationship that describes the mode that a device is operating in.,https://www.theworldavatar.com/kg/ontodevice/,has mode
 hasDefaultPosition,Object Property,,,Valve,DefaultValvePosition,,"A relationship that describes a valve and it's default position (NC, NO etc).",https://www.theworldavatar.com/kg/ontodevice/,has default position
+hasCorrespondingRange,Object Property,,,https://saref.etsi.org/core/State,http://www.w3.org/ns/ssn/systems/OperatingRange,,A relationship that describes a state to it's corresponding range (e.g. a state of 0 to 100% corresponds to a range of 0 to 1000 rpm).,https://www.theworldavatar.com/kg/ontodevice/,has corresponding range
```
sm453, line 175: It would be more conventional to call this relationship `hasOperatingRange`, given that its range is `OperatingRange`.

Also, please change the description: "relates" instead of "describes", and "its" instead of "it's".

sm453 (changes requested):

JPS_Ontology/ontology/ontodevice/OntoDevice.csv
Abyssus1010: will it be correct if I use the relationship correlatesTo?

e.g. I have a gear pump that can be cofigured to operate from 0 to 100%. And I know that the gear pump has  a operating range of 0 to 5000 rpm. Will it be correct for me to say that this percentage value correlatesTo the operating range?

sm453: Doesn't sound right to me. The word "correlation" is usually reserved for a statistical relationship (in a well-defined sense), so, using to relate a percentage value to an operating range seems wrong to me.

sm453 (approved):

PR #1137, opened 2024-03-13
1777 dev update osm agent matching iteration

YiKaiTsai:


kok-foong:
Have you tested this?

YiKaiTsai:
> Have you tested this?

Yes. Tested on the Pirmasens dataset, and compared it to the results from 1.5.0, same number of OSM geometries matched.

kok-foong:
what about speed and performance?

YiKaiTsai:
> what about speed and performance?

I thought this update was so that the OSM agent could run in the background while not taking up most of the memory, and that the speed was not being considered.

kok-foong:
yes, have you observed that the memory usage is lower?

YiKaiTsai:
> yes, have you observed that the memory usage is lower?

Yes.

kok-foong (approved):

PR #1138, opened 2024-03-13
Format the time String put into the times table in Postgres

Ushcode:
When netCDF time values are not cast to ISO standard, the string is recorded into a table in the database.

This PR would add slight reformatting of these strings to appear cleaner for when they are surfaced in a visualisation


Ushcode:
image


Quick example of how different timestrings are formatted. These will be those specified in the 

```json
"mdimSettings": {
        "layerArrayName": "array",
        "timeOptions": {
          "arrayName": "_this piece_",
          "regex": ".*([0-9]{10}).*",
          "timeZone": "GMT"
        }
      }
```

part of the uploader config

gpeb2 (approved):
I think this is a good improvement for now, although as discussed in issue #1148 there is a more robust solution available to parse and format the 30-day-month datetimes.

PR #1145, opened 2024-03-18
Only run custom CRS to postgis if `SridIn` is not specified

Ushcode:


Ushcode:
@kok-foong could you pull the stack manager and data uploader images with tags `1.25.2-sridin-SNAPSHOT` and try upload your .xyz raster data again? The behaviour has now changed so that if an input SRID is specified in the `gdaloptions` the custom CRS method will not be called

kok-foong (approved):
tested

PR #1162, opened 2024-04-16
Dev composite materials ontology

mhs62:


JPS_Ontology/KBTemplates/TBox/ontocompmat.csv
sm453, line 1: Please move this file to `JPS_Ontology/ontology/ontocompmat/`.

PR #1163, opened 2024-04-17
Dev ontobuildenv update

YiKaiTsai:


JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
```diff
@@ -118,3 +118,10 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
...
 Bank,Class,Non-Domestic,IS-A,,,,A property where customers can deposit and withdraw money.,https://www.theworldavatar.com/kg/ontobuiltenv/,Bank
 Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are prepared or sold.,https://www.theworldavatar.com/kg/ontobuiltenv/,Pharmacy
+BuildingFloors,Class,,,Property,,,A concept to describe amount of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Building floors
+hasBuildingFloors,Object Property,,,Property,BuildingFloors,exactly 1,A relationship between a property and its amount floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has building floors
+hasFloorsValue,Data Property,,,BuildingFloors,integer,exactly 1,A relationship between a building floors and its value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has floors value
+hasAccuracy,Data Property,,,BuildingFloors,string,exactly 1,A relationship between a building floors and its accuracy level.,https://www.theworldavatar.com/kg/ontobuiltenv/,has accuracy
+hasGFA,Data Property,,,Property,float,exactly 1,A relationship between a building and its GFA value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has GFA
```
ShinZert, line 125: Might be good to use full name Gross Floor Area instead of acronym to provide clarity to reader

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
```diff
@@ -118,3 +118,10 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
 representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,exactly 1,,https://www.theworldavatar.com/kg/ontobuiltenv/,representative for
 Bank,Class,Non-Domestic,IS-A,,,,A property where customers can deposit and withdraw money.,https://www.theworldavatar.com/kg/ontobuiltenv/,Bank
 Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are prepared or sold.,https://www.theworldavatar.com/kg/ontobuiltenv/,Pharmacy
+BuildingFloors,Class,,,Property,,,A concept to describe amount of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Building floors
+hasBuildingFloors,Object Property,,,Property,BuildingFloors,exactly 1,A relationship between a property and its amount floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has building floors
+hasFloorsValue,Data Property,,,BuildingFloors,integer,exactly 1,A relationship between a building floors and its value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has floors value
+hasAccuracy,Data Property,,,BuildingFloors,string,exactly 1,A relationship between a building floors and its accuracy level.,https://www.theworldavatar.com/kg/ontobuiltenv/,has accuracy
```
ShinZert, line 124: While I understand it's good to mark up the provenance of building floors (i.e., generated from either HDBPropertyInformation.csv, OpenStreetMap, BuildingHeightEstimation). BuildingFloors can't really have accuracy (i.e., high, medium, low) because there is no indication why one dataset is more accurate than the other.

Would something along the line of <BuildingFloors> <derivedFrom> <HDBPropertyInformation/OpenStreetMap/BuildingHeightEstimation> be better than high, medium, low, accuracy?

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
```diff
@@ -118,3 +118,10 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
...
+hasGFA,Data Property,,,Property,float,exactly 1,A relationship between a building and its GFA value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has GFA
+hasCountry,Data Property,,,http://ontology.eil.utoronto.ca/icontact.owl#Address,string,exactly 1,The country where the address is located.,https://www.theworldavatar.com/kg/ontobuiltenv/,has country
```
ShinZert, line 126: Might be good to keep consistency with previous way of writing the relationship in address 

Example: 
A relationship between an address and the respective country.
[hasStreet](https://github.com/cambridge-cares/TheWorldAvatar/blob/45776dd293a74ac55b8a3ff679cfaa37c43aa2e2/JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv#L27)

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
```diff
@@ -118,3 +118,10 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
...
+hasCity,Data Property,,,http://ontology.eil.utoronto.ca/icontact.owl#Address,string,exactly 1,The city where the address is located.,https://www.theworldavatar.com/kg/ontobuiltenv/,has city
```
ShinZert, line 127: Same comment on keeping the consistency in description of hasCity with other "has" relationships of address.

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
```diff
@@ -118,3 +118,10 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
 representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,exactly 1,,https://www.theworldavatar.com/kg/ontobuiltenv/,representative for
 Bank,Class,Non-Domestic,IS-A,,,,A property where customers can deposit and withdraw money.,https://www.theworldavatar.com/kg/ontobuiltenv/,Bank
 Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are prepared or sold.,https://www.theworldavatar.com/kg/ontobuiltenv/,Pharmacy
+BuildingFloors,Class,,,Property,,,A concept to describe amount of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Building floors
+hasBuildingFloors,Object Property,,,Property,BuildingFloors,exactly 1,A relationship between a property and its amount floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has building floors
+hasFloorsValue,Data Property,,,BuildingFloors,integer,exactly 1,A relationship between a building floors and its value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has floors value
```
ShinZert, line 123: "a building floors " does not seem grammatically correct to me

ShinZert (changes requested):

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
YiKaiTsai: Was following how the comments were labelled inside the icontact ontology, where the Address and has... concepts originally came from. Since we are extending inside OntoBuiltEnv instead of icontact, I guess it would be better to follow the convention inside OntoBuiltEnv instead of icontact. Will update accordingly.

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
yjytt59: too long, it can be described in comments, but it's not good for such a long name

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
yjytt59: it's class name 'BuildingFloors'

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
yjytt59: too long, it can be described in comments, but it's not good for such a long name

ShinZert: Yep agree, describing in comment sounds good

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
ShinZert: Unsure how that is not easy for further query and identification, can get @YiKaiTsai and @sganguly-cares to weigh in on this

YiKaiTsai: @sm453, would appreciate your advice on this.

yjytt59: 1. It doesn't make sense to store such a long and irregular name of data sources in database. 2. The objective of this concept is to help us know the reliability of data, then we can know which data should be kept. When querying get a data source name, how could the agent know it is a reliable data or not? how to make a decision to update data or not? Even in the future if some new data is coming, how could agent know the data is reliable or not? 

So far, data accuracy level: 1. high: data from reliable departments (e.g. government agencies) 2. mid: open source data without any guarantees (e.g. OSM) 3. low: estimate by height. I think it's very clear and easy for further work.

sm453: A few comments/suggestions:
1) Using string literals to represent categorical data ("high", "medium", "low") is really the worst way to represent this. It's slow, storage-inefficient, error-prone, and completely non-semantic. I would strongly recommend to change this (pretty much any alternative will be preferable).
2) Using the word "accuracy" leads me to expect some kind of quantified uncertainty, e.g. a percentage error, whereas what you have in mind, I guess, is much less quantitative and more akin to a "level of confidence" that you have in the data (or the entity providing them). I would find it much clearer if you use the latter wording.
3) I agree with @ShinZert's point that rather than associating a subjective assessment of quality (that may offend people!) to a dataset, it would be much preferable to simply store the provenance (where the data came from). Then any "user" can make up their own mind how good the data are, without having to spell it out!
4) If you do decide to go with some kind of categorical solution, I would recommend using subclasses (rather than string datatype properties), because this does not rely on string-processing and is fully semantic. So, for example, if you want to categorise a class `A`, you simply define subclasses `A_cat1`, `A_cat2`, etc. corresponding to each of your categories. As the subclasses inherit all properties of the parent class, this is a really light-weight solution.

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
```diff
@@ -118,3 +118,10 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
 representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,exactly 1,,https://www.theworldavatar.com/kg/ontobuiltenv/,representative for
 Bank,Class,Non-Domestic,IS-A,,,,A property where customers can deposit and withdraw money.,https://www.theworldavatar.com/kg/ontobuiltenv/,Bank
 Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are prepared or sold.,https://www.theworldavatar.com/kg/ontobuiltenv/,Pharmacy
+BuildingFloors,Class,,,Property,,,A concept to describe amount of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Building floors
+hasBuildingFloors,Object Property,,,Property,BuildingFloors,exactly 1,A relationship between a property and its amount floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has building floors
```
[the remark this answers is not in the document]
sm453: Using "BuildingFloors" as a class name, when you mean really just their number, is highly confusing. Please do clarify in all class/property names that you mean "NumberOf".

JPS_Ontology/ontology/ontobuiltenv/OntoBuildingEnvironment.csv
sm453, line 1: Please note that "the built environment" is a fixed expression - see e.g. [Wikipedia](https://en.wikipedia.org/wiki/Built_environment). Thus, I believe the name of the ontology (`OntoBuiltEnv`) is correct, but the name of the CSV file, and the description need changing.

sm453 (changes requested):

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
```diff
@@ -121,7 +121,9 @@ Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are p
 BuildingFloors,Class,,,Property,,,A concept to describe amount of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Building floors
```
sm453, line 121: As commented previously, I strongly recommend renaming `BuildingFloors` to `NumberOfFloors` (or similar, making it clear that this class only refers to a number, and not the floors themselves). In addition, in the comment, "amount" should be replaced with "number".

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
```diff
@@ -121,7 +121,9 @@ Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are p
...
 hasBuildingFloors,Object Property,,,Property,BuildingFloors,exactly 1,A relationship between a property and its amount of floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has building floors
 hasFloorsValue,Data Property,,,BuildingFloors,integer,exactly 1,A relationship between the floors and its value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has floors value
```
sm453, line 123: What is the "value" of a (set of) floors? This is completely unclear, and the comment doesn't explain anything. Please clarify.

sm453 (changes requested):
In addition to the comments, please change the description text for the ontology (in line 4 of the csv file). I would replace "The Building Environment ontology (OntoBuiltEnv)" with "The Ontology for the Built Environment (OntoBuiltEnv)".

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
yjytt59: ok changed in the latest revision

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
yjytt59: it's a data property to describe the numerical value of NumberOfFloors, as NumberOfFloors is a class.

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
```diff
@@ -118,12 +118,12 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
...
-FloorsCatA,Class,BuildingFloors,IS-A,,,,Category A of floors: data from reliable departments (e.g. government agencies),https://www.theworldavatar.com/kg/ontobuiltenv/,floors category A
-FloorsCatB,Class,BuildingFloors,IS-A,,,,Category B of floors: open source data without any guarantees (e.g. OSM),https://www.theworldavatar.com/kg/ontobuiltenv/,floors category B
-FloorsCatC,Class,BuildingFloors,IS-A,,,,Category C of floors: estimate,https://www.theworldavatar.com/kg/ontobuiltenv/,floors category C
+hasNumberOfFloors,Object Property,,,Property,NumberOfFloors,exactly 1,A relationship between a property and its number of floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has number of floors
+NumberOfFloors,Class,,,Property,,,A concept to describe number of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Number of floors
+hasNumberOfFloorsValue,Data Property,,,NumberOfFloors,integer,exactly 1,A relationship between the number of floors and its numerical value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has number of floors value
```
sm453, line 123: Given that the domain class is now called `NumberOfFloors`, it would be customary to name this data property simply `hasValue`. (This is analogous to any quantity that e.g. has a value and a unit. The meaning is now obvious, given the new name of the domain class.)

sm453 (changes requested):

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
yjytt59: ok I will change it to `hasValue`. I was thinking to use `hasValue`, just afraid it's too simply and confuse with the other property definition. As you said, it seems no issue.

yjytt59:
I add a new object property `hasEstimatedConstructionCost` to describe the construction cost of the building. Because we haven't accurate construction cost data, it is calculated by the latest construction cost standard, not real construction cost of building, I call it as estimated. `hasEstimatedConstructionCost` link to `om:Cost`, which is subClass of `om:AmountOfCost`.

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
```diff
@@ -45,7 +45,7 @@ hasTotalRoofArea,Object Property,,,http://www.purl.org/oema/infrastructure/Build
...
 http://www.ontology-of-units-of-measure.org/resource/om-2/hasValue,Object Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Area UNION http://www.ontology-of-units-of-measure.org/resource/om-2/Height UNION http://www.ontology-of-units-of-measure.org/resource/om-2/AmountOfMoney UNION AveragePricePerSqm,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,,,http://www.ontology-of-units-of-measure.org/resource/om-2/,has value
-http://www.ontology-of-units-of-measure.org/resource/om-2/hasNumericalValue,Data Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure,float,,,http://www.ontology-of-units-of-measure.org/resource/om-2/,has numerical value
+http://www.ontology-of-units-of-measure.org/resource/om-2/hasNumericalValue,Data Property,,,http://www.ontology-of-units-of-measure.org/resource/om-2/Measure UNION http://www.ontology-of-units-of-measure.org/resource/om-2/Cost,float,,,http://www.ontology-of-units-of-measure.org/resource/om-2/,has numerical value
```
sm453, line 48: Redefining someone else's classes/properties is not generally a good idea and should be avoided if at all possible. I would recommend removing such re-definitions. There seem to be several in this file.

Also, I believe `om:hasValue`, `om:hasNumericalValue`, and `om:hasUnit` can be used without having to change any definitions.

sm453 (changes requested):

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
sm453: Having discussed this with @yjytt59, we have concluded that, since this is an existing problem and not one that was introduced in this PR, we leave it as it is.

However, please do add a `README.md` file that explains this problem and notes this as an open TO-DO item.

yjytt59: I already add a `README.md` to explain this issue and the current solution in sea level rise project.

sm453 (approved):

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
```diff
@@ -118,3 +118,13 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
 representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,exactly 1,,https://www.theworldavatar.com/kg/ontobuiltenv/,representative for
 Bank,Class,Non-Domestic,IS-A,,,,A property where customers can deposit and withdraw money.,https://www.theworldavatar.com/kg/ontobuiltenv/,Bank
 Pharmacy,Class,Medical,IS-A,,,,A healthcare facility where medicinal drugs are prepared or sold.,https://www.theworldavatar.com/kg/ontobuiltenv/,Pharmacy
+hasNumberOfFloors,Object Property,,,Property,NumberOfFloors,exactly 1,A relationship between a property and its number of floors.,https://www.theworldavatar.com/kg/ontobuiltenv/,has number of floors
+NumberOfFloors,Class,,,Property,,,A concept to describe number of floors in a building.,https://www.theworldavatar.com/kg/ontobuiltenv/,Number of floors
```
ShinZert, line 122: A concept to describe **the** number of floors in a building

JPS_Ontology/ontology/ontobuiltenv/OntoBuiltEnv.csv
```diff
@@ -118,3 +118,13 @@ representativeFor,Object Property,,,AveragePricePerSqm,PostalCode,,A relationshi
...
+hasValue,Data Property,,,NumberOfFloors,integer,exactly 1,A relationship between the number of floors and its numerical value.,https://www.theworldavatar.com/kg/ontobuiltenv/,has value
+FloorsCatA,Class,NumberOfFloors,IS-A,,,,Category A of floors: data from reliable departments (e.g. government agencies),https://www.theworldavatar.com/kg/ontobuiltenv/,floors category A
```
ShinZert, line 124: The label column for floors catergory is lower-cased instead of usual upper-case for first word (Optional change)

ShinZert (approved):

PR #1169, opened 2024-04-19
Fixes to cities related ontologies

qhouyee:


picas9dan (commented):
Hi Hou Yee, thank you for making the changes.
I have one question. At [this](https://github.com/cambridge-cares/TheWorldAvatar/blob/7393a68848ed5f1df85e7bf79fd7f9f6e964a27a/JPS_Ontology/ontology/ontozoning/OntoZoning.owl#L2480) line, I find the IRI "https://www.theworldavatar.com/kg/ontozoning/Plot". I would like to clarify if it should be there or it should not be present at all in the owl file.

qhouyee:
Yes, it should be removed and thank you for catching this. I have regenerated the owl file from the TBox generator, and it should be working as expected.

PR #1170, opened 2024-04-23
Vis Platform Map Layer Bug Fix

qhouyee:
Fixes the disappearing map layers when switching pages or info panel


jlai2021 (commented):
With these changes, the code should work as expected.

jlai2021 (approved):
Reviewed

PR #1172, opened 2024-04-24
Hotfix update kingslynn credo integration

markushofmeister:
This PR updates the CReDo scenario IDs for the King's Lynn visualisation of the flood warnings. This visualisation sticks to `credo-phase-two` to avoid ongoing changes of scenario IDs.
@Ushcode if there is an easy and dynamic way to get the scenario IDs, this could be explored; however, the `manager.selectedScenario` etc. methods won't work as discussed

### DIFF
--- a/Deploy/stacks/KingsLynn/StackDeployment/inputs/stack-manager/inputs/data/visualisation/data.json
+++ b/Deploy/stacks/KingsLynn/StackDeployment/inputs/stack-manager/inputs/data/visualisation/data.json
@@ -397,6 +397,7 @@
         {
             "name": "Population Data",
             "stack": "http://165.232.172.16:3838",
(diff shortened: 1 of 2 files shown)


markushofmeister:
> LGTM, although these hardcoded scenario IDs will change if the scenario is run again (which is not likely to be any time soon) so at some point in the future if phase2 viz at `kg.cmclinnovations.com` goes down and we've to spin it back up this will break again.
> 
> As discussed we can use the endpoint in the javascript code to edit a local `data.json` but this is a fine fix for the time being to get things back up

Makes sense. Potentially easier would be if the scenario IDs are available via some basic API, so that I could quickly get them in a script to update the data.json. This wouldn't need any scenario selector, but a quick restart of my stack.

PR #1175, opened 2024-04-25
Auto restart stacks when they come down

gpeb2:
Change the restart policy of the containers started by the stack so that they restart if they fail. The main intention being for them to restart after a machine reboot or restarting docker.


Deploy/stacks/dynamic/stack-clients/src/main/java/com/cmclinnovations/stack/services/DockerService.java
```diff
@@ -326,7 +326,9 @@ protected ServiceSpec configureServiceSpec(ContainerService service) {
                 .withName(service.getContainerName())
                 .withLabels(StackClient.getStackNameLabelMap());
         service.getTaskTemplate()
-                .withRestartPolicy(new ServiceRestartPolicy().withCondition(ServiceRestartCondition.NONE))
+                .withRestartPolicy(new ServiceRestartPolicy()
+                        .withCondition(ServiceRestartCondition.ON_FAILURE)
+                        .withMaxAttempts(3l))
```
Ushcode, line 331: 31 seems a random number, was it meant to be 3 or 1?

Ushcode: Confirmed this is `3l` not `31`  🙃

Ushcode (approved):

PR #1177, opened 2024-04-26
Dev cleanup dispersion data

kok-foong:
Main additions:
1) Number of GeoServer layers is kept to 20 by default to keep GeoServer performance reasonable, also specified via NUMBER_OF_LAYERS in AermodAgent.

2) Added NUMBER_OF_LAYERS parameters in DispersionInteractor that specifies the number of timesteps to show in the dispersion dropdown, should ideally match the number specified in AermodAgent

2) Added a route in ShipInputAgent to clear old data, via the /clear-old-data route

3) Added IGNORE_BUILDINGS flag in AermodAgent when the buildings dataset is too large and slows down the simulation

4) Limit number of process in AermodAgent to 1, in case of duplicate requests to updateDerivation when the previous request is still executing

5) Limit number of process in UpdateVirtualSensors route in DispersionInteractor, to prevent duplicate requests from the visualisation


ChungTingLao (changes requested):
The environment variables for dispersion-interactor and dispersion-interactor-debug should be identical.

Please switch to use stack manager 1.27.1, otherwise visualisation does not work.

When ship data are deleted (/clear-old-data), dispersion results are no longer visible in visualisation, but the ship layer still shows up despite all ship data were supposedly deleted.

kok-foong:
> When ship data are deleted (/clear-old-data), the ship layer still shows up in visualisation despite all ship data were supposedly deleted.

This is intentional, the ships layers are created by the Aermod agent and only stored in 1 table, hence they can be easily deleted manually if needed. The ships data is very difficult to clear without resetting the database

ChungTingLao (approved):
Everything works fine on my end, please relabel and rebuild all modified docker images before merging.

PR #1178, opened 2024-04-26
Add escape / close button to scenario selector

Ushcode:
There is now a small x in the top right of the scenario selector

* This will only show up if a scenario has been selected at least once (so you can't skip it and have a map with no data)
* Also works to hit the `esc` key

- **confirm px height of footer**
- **scenario escape**

### DIFF
--- a/web/twa-vis-framework/library/src/ts/ui/scenario_handler.ts
+++ b/web/twa-vis-framework/library/src/ts/ui/scenario_handler.ts
@@ -7,7 +7,7 @@ class ScenarioHandler {
     /**
      * Base URL of the agent to contact about scenarios.
      */
-    private agentBaseURL : string;
+    private agentBaseURL: string;
 
     /**
      * Optional dataset identify to send to the scenario agent.
@@ -17,7 +17,7 @@ class ScenarioHandler {
     /**
      * JSON object holding definitions of possible scenarios.
      */
-    private definitions: JSON; 
+    private definitions: JSON;
 
     /**
      * ID of currently selected scenario.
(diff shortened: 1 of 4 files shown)


web/twa-vis-framework/library/src/ts/ui/scenario_handler.ts
```diff
@@ -74,43 +74,40 @@ class ScenarioHandler {
...
-            // Send of request to get the scenario details
-            let self = this;
+                            </div>
+                            </div>
+                            </div>
+                            `;
+                            document.body.appendChild(container);
+                            
+                            // Send of request to get the scenario details
+                            let self = this;
```
gpeb2, line 102: Has the formatting gone wrong here?

gpeb2 (changes requested):
On my machine the `scenario-header` seems to overlap with the `scenario-close` button, see image. I've also made a minor point about code formatting.
Also would it be possible to either highlight the current scenario and/or make it clickable? I think adding a call to the close function when the scenario ID matches the current one, rather than just `return`.
image

web/twa-vis-framework/library/src/ts/ui/scenario_handler.ts
Ushcode: Yes looks like autoformatter doesn't work on string injected HTML in the .js files. Fixed now!

Ushcode (commented):
> On my machine the `scenario-header` seems to overlap with the `scenario-close` button, see image. I've also made a minor point about code formatting.
Fixed this.

Code should be there to show currently selected scenario in a darker colour but only one scenario currently up on hobbes to test

gpeb2 (approved):

PR #1182, opened 2024-04-29
1181 add ability to specify time in feature info agent

cmclscourtney:
@kok-foong I am not in a position to, but would be great if you could test this with your use case.

Version `3.1.0-1181-SNAPSHOT` is pushed.


kok-foong (approved):
Tested

PR #1187, opened 2024-05-02
Update some package config files

Ushcode:
- npm likes lowercase no space package names
- change module resolution to node instead of bundler
- update conflicting dependencies
- Bundler
- resolve dependencies not found by node
- fix types resolution


qhouyee:
Verified that the updated platform can be deployed on Docker for all environments

qhouyee (approved):

PR #1190, opened 2024-05-07
dev-zaha-icl

picas9dan:


QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/func/sg_carpark.py
```diff
@@ -0,0 +1,103 @@
...
+        timestamp = time.time()
+        query = """PREFIX carpark:	<https://www.theworldavatar.com/kg/ontocarpark/>
+    PREFIX geo: <http://www.opengis.net/ont/geosparql#>
+    PREFIX geof: <http://www.opengis.net/def/function/geosparql/>
+    PREFIX unit: <http://qudt.org/vocab/unit/>
+            
+    SELECT * WHERE {{
+    ?Carpark a carpark:Carpark; rdfs:label ?Label .
+    SERVICE <http://sg-ontop:8080/sparql> {{
+        SELECT (geof:distance(?Coords, "<http://www.opengis.net/def/crs/OGC/1.3/CRS84> POINT({lon} {lat})"^^geo:wktLiteral, <http://www.opengis.net/def/uom/OGC/1.0/metre>) as ?Distance) ?Carpark ?Coords
```
gpeb2, line 59: In case you didn't know `<http://www.opengis.net/def/crs/OGC/1.3/CRS84>` is the default CRS with GeoSPARQL so shouldn't be needed here. Saying that, versions of Ontop before version 5, which the stack now uses, might have needed this.

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_response.py
```diff
@@ -0,0 +1,101 @@
...
+from services.entity_store import EntityStore, get_entity_store
+from services.support_data import DataItem, TableDataItem, WktDataItem
+
+
+class SparqlResponseProcessor:
+    TWA_PREFIXES = [
+        "http://www.theworldavatar.com/kg/",
+        "https://www.theworldavatar.com/kg/",
+    ]
+    WKT_LITERAL_PREFIX = "<http://www.opengis.net/def/crs/OGC/1.3/CRS84> "
```
gpeb2, line 15: Linked to the other comment, but be aware that this is not the only CRS that WKT literals can come in, although I would generally encourage using this one when possible as it is the default and is effectively the same as EPSG:4326.

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_query.py
```diff
@@ -0,0 +1,171 @@
...
+                break
+            idx_end = token_start
+
+            values = []
+            for token in tokens:
+                values.extend(self._link_token(token))
+
+            # TODO: remove custom replacement of VALUES clause with FILTER
+            # once Karthik fixes the bug that causes ontop to throw error
+            # on VALUES of facility IRIs.
```
gpeb2, line 111: What is the error you are seeing here?
I know that Ontop version 4 didn't handle `VALUES` efficiently but the stack now uses version 5 which does.

I also notice that you are working with IRIs of the form `<Type:...>` and I'm not sure that's a good idea. According to the [URI/IRI specifications](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#syntax) the first part of a URL/IRI (before the first ":") should be its schema, generally taken from the official list of schemas.

gpeb2 (commented):
I've added a few comments, feel free to reply here or we can set up a Skype of Teams call if required.

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/func/sg_carpark.py
picas9dan: Hi @kok-foong, could you help confirm if `<http://www.opengis.net/def/crs/OGC/1.3/CRS84>` is needed here for the query that finds the carpark nearest to a give set of coordinates, in view of the current setup and future compatibility on your end?

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_response.py
picas9dan: Is there a specification for WKT literals that I can refer to? So that I can safely parse WKT literals into approprate data objects. Or is there a library that I can use for this?

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_query.py
picas9dan: This is the error that I'm seeing. I only encounter it when I use VALUES clause on IRIs of data centres and factories. I have asked @nagarajankarthik about this. I'm not sure what the timeline for fixing this issue is.
Screenshot 2024-05-07 at 8 49 58 PM

For reference, here is the SPARQL query
```
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX om: <http://www.ontology-of-units-of-measure.org/resource/om-2/>
PREFIX ontocompany: <http://www.theworldavatar.com/kg/ontocompany/>

SELECT DISTINCT ?IRI ?GeneratedHeatValue ?GeneratedHeatUnit WHERE {
  VALUES ?IRI { <https://www.theworldavatar.com/kg/infrastructure/datacenter/c84c1d32-69de-508f-a419-83096735aadd> <https://www.theworldavatar.com/kg/infrastructure/datacenter/7f8a8c48-516b-5757-a332-2a264052bdf8> }
  ?IRI ontocompany:hasGeneratedHeat/om:hasValue ?GeneratedHeat .
  ?GeneratedHeat om:hasNumericalValue ?GeneratedHeatValue .
  OPTIONAL { ?GeneratedHeat om:hasUnit/skos:notation ?GeneratedHeatUnit . }
}
```

picas9dan: Regarding my use of `<Type:"surface form">`. The purpose of this is to serve as a marker for entities that need to be linked to exact IRIs and it is not meant to be a faithful representation of an URI/IRI as per general specifications. 

In literature, I have seen people propose special tokens for this purpose e.g. `<entity><type>Facility</type><label>ExxonMobil</label></entity>`. However, I find this too verbose. 

If you think that it is preferable to use special tokens in this case, I'll make the change.

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_response.py
gpeb2: Unfortunately the GeoSPARQL standard doesn't actually support CRSs and officially assumes that all WKT literals are in CRS84. Ontop extends the official standard by allowing an optional CRS IRI to be appended to the start of the WKT literal. An example of some of these can be seen in [this OBDA file](https://github.com/ontop/ontop/blob/170318de14b1affdeeab43bf3fe2dee5cd0cf9e7/binding/owlapi/src/test/resources/geosparql/geosparql-h2.obda#L10).

I'd probably recommend forcing developers to reproject their non-EPSG:4326 geometries to EPSG:4326 in the `source` query of their OBDA mappings using the PostGIS [`ST_Transform`](https://postgis.net/docs/ST_Transform.html) function and not specifying the CRS IRIs in the WKTLiterals. This will save confusion as Ontop is not able to perform operations on geometries in different projections and you won't need to parse out the CRS IRIs.

picas9dan: Is the projection to EPSG:4326 and removal of CRS IRI from WKTLiteral a to-do item that you would suggest to be done asap, following which Zaha should not check for CRS IRI and accept the WKTLiteral verbatim? Then I'll need to coordinate with Kok Foong to make the change. 

Otherwise, the current code works for both scenarios when `<http://www.opengis.net/def/crs/OGC/1.3/CRS84>` is prepended and when it is not, so no changes are needed (unless the code must also anticipate other CRS IRIs).

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_query.py
gpeb2: > Regarding my use of `<Type:"surface form">`. The purpose of this is to serve as a marker for entities that need to be linked to exact IRIs and it is not meant to be a faithful representation of an URI/IRI as per general specifications.

I don't fully understand the interplay between your code, Redis and the triples but what you say makes enough sense so its probably ok.

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_response.py
gpeb2: Keeping the code as it is is fine for now, I just wanted to make sure you were aware of the potential issues.

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/func/sg_carpark.py
gpeb2: @kok-foong, please check my recommendation in [this comment](https://github.com/cambridge-cares/TheWorldAvatar/pull/1190#discussion_r1592535569).

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/func/sg_carpark.py
```diff
@@ -0,0 +1,103 @@
...
+
+        timestamp = time.time()
+        query = """PREFIX carpark:	<https://www.theworldavatar.com/kg/ontocarpark/>
+    PREFIX geo: <http://www.opengis.net/ont/geosparql#>
+    PREFIX geof: <http://www.opengis.net/def/function/geosparql/>
+    PREFIX unit: <http://qudt.org/vocab/unit/>
+            
+    SELECT * WHERE {{
+    ?Carpark a carpark:Carpark; rdfs:label ?Label .
+    SERVICE <http://sg-ontop:8080/sparql> {{
```
kok-foong, line 58: just a comment, don't think it's a good idea to hardcode this, if we spin up the stack with a different stack name this breaks

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/sparql/process_response.py
picas9dan: @kok-foong pointed me to Jena library for parsing WKT literal, specifically the method [`WKTReader::extract`](https://github.com/apache/jena/blob/cf3a14008b7cbcf5abc8cf485eec4501566f8469/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/parsers/wkt/WKTReader.java#L255). Referencing the Jena implementation, I wrote a simple [method to parse WKT literals](https://github.com/cambridge-cares/TheWorldAvatar/blob/fa0166847fd769940388bec56559619ca7d4b824/QuestionAnswering/QA_Agent/deployment/fastapi_app/services/wkt.py).

If the SRS URI is found to be CRS84, then the WKT will be visualised. Otherwise, it will be presented verbatim in a table. The code for this handling is [here](https://github.com/cambridge-cares/TheWorldAvatar/blob/fa0166847fd769940388bec56559619ca7d4b824/QuestionAnswering/QA_Agent/deployment/fastapi_app/controllers/qa/execute_action/sparql/process_response.py#L83).

QuestionAnswering/QA_Agent/deployment/fastapi_app/services/nlq2action/execute_action/func/sg_carpark.py
picas9dan: Is the endpoint http://sg-ontop:8080/sparql (which I suppose is referencing the local network of the SG stack) the same as http://174.138.23.221:3838/ontop/ui/sparql? I already configure the latter as an environment variable, so if the 2 endpoints are the same, then I'll interpolate the query string with the latter. Otherwise, I'll define a new environment variable for this.

kok-foong: No, they are not the same, one is for accessing within the stack and one is from outside

kok-foong: A possible solution is to have an access agent within the stack which you send queries to, instead of sending it directly to blazegraph which will figure out the correct ontop endpoint

picas9dan: I tested the query with http://sg-ontop:8080/sparql and http://174.138.23.221:3838/ontop/ui/sparql and it seems that they give the same results/behaviours? Does the difference lie only in speed? In other words, is the use of either endpoint correct?

> A possible solution is to have an access agent within the stack which you send queries to, instead of sending it directly to blazegraph which will figure out the correct ontop endpoint

Yes, this will ensure that Zaha doesn't need to know the internal endpoints of the SG stack.

picas9dan: > A possible solution is to have an access agent within the stack which you send queries to, instead of sending it directly to blazegraph which will figure out the correct ontop endpoint

Will this be something that you plan to implement within next week?

kok-foong: > > A possible solution is to have an access agent within the stack which you send queries to, instead of sending it directly to blazegraph which will figure out the correct ontop endpoint
> 
> Will this be something that you plan to implement within next week or so?

It's not straightforward to implement this for the Sg stack/dataset, can explain more to you in person

picas9dan: if this is not in your work plan in the next 1-2 weeks then for immediate action I'll use a environment variable for `http://sg-ontop:8080/sparql`.

PR #1201, opened 2024-05-09
process excel files

Ushcode:


cmclscourtney:
Note that if layer name is declared all sheets will get merged into a single table.

Ushcode:
@cmclscourtney Good spot, did setting the layer to null fix that

cmclscourtney:
> @cmclscourtney Good spot, did setting the layer to null fix that

It did but I took it out since it would be nice to be able to have the option of renaming if you only have one sheet. For instance a lot of UKPN open data has a single sheet call "Data" and that could conflict if that becomes the name of the table

cmclscourtney (approved):

PR #1204, opened 2024-05-14
1794-dev-ontosealevel

YiKaiTsai:
There a few issues/concerns that require attention. @sm453 , would appreicate your inputs, thanks.

1) Currently, there is a `SeaLevel` concept and a `SeaLevelMeasurement` concept. The intention is that a sea level (say global mean sea level or Singapore mean sea level) can have different measurements at different date and time. Instead of instantiating different instances of `SeaLevel` each time we measure the global mean sea level or Singapore mean sea level, we have one instance of `SeaLevel` to represent the concept of the global mean sea level or Singapore mean sea level, then different instances of `SeaLevelMeasurement` to represent the sea level measured at different time. And then `om:Height` is linked to `SeaLevelMeasurement`. @sm453, would appreciate your thoughts on whether this would be the good practice or `om:Height` is to be linked to `SeaLevel` and `SeaLevelMeasurement` to be removed?

2) Currently, the concept of `Damage` due to rising sea level is incorporated into OntoSeaLevel, and is linked directly to `SeaLevelMeasurement` via `hasImpact`. I feel that there is a better way to represent this. 

For example, we could have a concept called `SeaLevelRise`, perhaps in OntoSeaLevel or a new ontology called OntoSeaLevelRise. Then, one could have `SeaLevelRise` link to a `SeaLevel`, and have a starting `SeaLevelMeasurment` and an ending `SeaLevelMeasurement`. Then, the `Damage` concept can be linked to the `SeaLevelRise` concept via `hasImpact` and will no longer link to `SeaLevel`.

Please see the Miro board for a visual illustration. See the frame ‘2nd draft’ for the current OntoSeaLevel draft in this pull request. See the frame ‘potential 3rd draft’ for the illustration of the above comment in 2) regarding a possible `SeaLevelRise` concept.


sm453:
> 1. Currently, there is a `SeaLevel` concept and a `SeaLevelMeasurement` concept. The intention is that a sea level (say global mean sea level or Singapore mean sea level) can have different measurements at different date and time. Instead of instantiating different instances of `SeaLevel` each time we measure the global mean sea level or Singapore mean sea level, we have one instance of `SeaLevel` to represent the concept of the global mean sea level or Singapore mean sea level, then different instances of `SeaLevelMeasurement` to represent the sea level measured at different time. And then `om:Height` is linked to `SeaLevelMeasurement`. @sm453, would appreciate your thoughts on whether this would be the good practice or `om:Height` is to be linked to `SeaLevel` and `SeaLevelMeasurement` to be removed?

Sounds to me like any sea level should be a standard time series, but the way you talk about it suggests that isn't currently your plan? I'm guessing this would solve all the issues you mention. Also, this would have the added advantage that you could use the Forecasting Agent to create forecasts, and you wouldn't need to reinvent any wheels how to represent this ontologically (because this has already been done).

sm453:
> 2. Currently, the concept of `Damage` due to rising sea level is incorporated into OntoSeaLevel, and is linked directly to `SeaLevelMeasurement` via `hasImpact`. I feel that there is a better way to represent this.
> 
> For example, we could have a concept called `SeaLevelRise`, perhaps in OntoSeaLevel or a new ontology called OntoSeaLevelRise. Then, one could have `SeaLevelRise` link to a `SeaLevel`, and have a starting `SeaLevelMeasurment` and an ending `SeaLevelMeasurement`. Then, the `Damage` concept can be linked to the `SeaLevelRise` concept via `hasImpact` and will no longer link to `SeaLevel`.

For starters, I'd call it "change" rather than "rise", for generality. I believe it should simply point to two entries in a sea-level time series (the "start" and "end" of the delta). This would then contain all necessary information (the times over which the change is considered, and the start and end values of the sea level). You could then associate any damage to an instance of such a SeaLevelChange.

Whether such a design is politically a good idea is a different question... I would expect that this will expose a large attack surface, e.g. people will question whether any particular damage is really attributable to a certain change, and over what period this should or should not have been considered...

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
```diff
@@ -0,0 +1,34 @@
...
+measuredIn,Data Property,,,SeaLevelMeasurement,http://www.w3.org/2001/XMLSchema#date,,A relationship between a sea level measurement and the time of measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured in
+hasHeight,Object Property,,,SeaLevelMeasurement,http://www.ontology-of-units-of-measure.org/resource/om-2/Height,,A relationship between a sea level measurement and the sea level height.,https://www.theworldavatar.com/kg/ontosealevel/,has height
+isAffectedBy,Object Property,,,https://www.theworldavatar.com/kg/ontobuiltenv/Property,EconomicDamage,,A relationship between the damage due to rising sea level and the places it affects.,https://www.theworldavatar.com/kg/ontosealevel/,is affected by
+isAffectedBy,Object Property,,,CulturalSite,CulturalDamage,,A relationship between the damage due to rising sea level and the places it affects.,https://www.theworldavatar.com/kg/ontosealevel/,is affected by
+hasMonetaryValu,Object Property,,,EconomicDamage,AmountOfMoney,,A relationship between the economic damage due to rising sea level and its total monetary value.,https://www.theworldavatar.com/kg/ontosealevel/,has monetary value
```
sm453, line 32: typo: `hasMonetaryValue`

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
```diff
@@ -0,0 +1,34 @@
...
+hasInputScenario,Object Property,,,ForecastSeaLevelMeasurement,Scenario,,A relationship between the forecasted sea level measurement and the input scenario the forecast was based on.,https://www.theworldavatar.com/kg/ontosealevel/,has input scenario
+hasMeasurment,Object Property,,,SeaLevel,SeaLevelMeasurement,,A relationship between a sea level and its measurement.,https://www.theworldavatar.com/kg/ontosealevel/,has measurement
```
sm453, line 34: typo: `hasMeasurement`

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
```diff
@@ -0,0 +1,34 @@
...
+ForecastSeaLevelMeasurement,Class,SeaLevelMeasurement,IS-A,,,,Forecasted sea level measurement based on a forecasting model.,https://www.theworldavatar.com/kg/ontosealevel/,Forecast sea level measurement
+ForecastSeaLevelMeasurement,Class,https://www.theworldavatar.com/kg/ontotimeseries/Forecast,IS-A,,,,Forecasted sea level measurement based on a forecasting model.,https://www.theworldavatar.com/kg/ontosealevel/,
+Tree,Class,CulturalSite,IS-A,,,,Tree,https://www.theworldavatar.com/kg/ontosealevel/,Tree
+HeritageTree,Class,Tree,IS-A,,,,A tree that is the natural heritage of a specific place.,https://www.theworldavatar.com/kg/ontosealevel/,HeritageTree
+https://schema.org/Museum,Class,CulturalSite,IS-A,,,,,https://www.theworldavatar.com/kg/ontosealevel/,Museum
+https://schema.org/TouristAttraction,Class,CulturalSite,IS-A,,,,,https://www.theworldavatar.com/kg/ontosealevel/,Tourist attraction
+https://schema.org/LandmarksOrHistoricalBuildings,Class,CulturalSite,IS-A,,,,,https://www.theworldavatar.com/kg/ontosealevel/,Landmarks or historical building
```
sm453, line 21: Redefining other people's concepts is something that should be avoided, I believe.

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
```diff
@@ -0,0 +1,34 @@
...
+Scenario,Class,,,,,,Scenario that can act as input to a forecasting level for forecasting sea level.,https://www.theworldavatar.com/kg/ontosealevel/,Scenario
+SharedSocioEconomicPathway,Class,Scenario,IS-A,,,,A climate change scenario of projected socioeconomic changes.,https://www.theworldavatar.com/kg/ontosealevel/,Shared Socioeconomic Pathway
+measuredAt,Object Property,,,LocalMeanSeaLevel,http://www.opengis.net/ont/geosparql#Feature,exactly 1,A relationship between the local mean sea level and the reference point for the local mean sea level measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured at
+measuredIn,Data Property,,,SeaLevelMeasurement,http://www.w3.org/2001/XMLSchema#gYear,,A relationship between a sea level measurement and the time of measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured in
+measuredIn,Data Property,,,SeaLevelMeasurement,http://www.w3.org/2001/XMLSchema#date,,A relationship between a sea level measurement and the time of measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured in
```
sm453, line 28: Remember, making two separate statements about an entity is interpreted as a logical _conjunction_. In this case here, that would imply that the range of `measuredIn` has to be both a year and a date (which is impossible). In order to achieve what you want, you need a single definition, with the range being the union of year and date.

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
```diff
@@ -0,0 +1,34 @@
...
+hasHeight,Object Property,,,SeaLevelMeasurement,http://www.ontology-of-units-of-measure.org/resource/om-2/Height,,A relationship between a sea level measurement and the sea level height.,https://www.theworldavatar.com/kg/ontosealevel/,has height
+isAffectedBy,Object Property,,,https://www.theworldavatar.com/kg/ontobuiltenv/Property,EconomicDamage,,A relationship between the damage due to rising sea level and the places it affects.,https://www.theworldavatar.com/kg/ontosealevel/,is affected by
+isAffectedBy,Object Property,,,CulturalSite,CulturalDamage,,A relationship between the damage due to rising sea level and the places it affects.,https://www.theworldavatar.com/kg/ontosealevel/,is affected by
```
sm453, line 31: Similar to the comment on `measuredIn`, I believe that in order to make this correct, you need one statement defining `isAffectedBy` with both the domain and the range being the union of the relevant concepts. In addition, you then need statements that restrict values (using the `only` qualifier), so that one cannot make statements like `CulturalSite3 isAffectedBy EconomicDamage5`.

sm453 (changes requested):

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
YiKaiTsai: Created our own class, and gave `equivalentTo` definiton where needed.

YiKaiTsai:
> Sounds to me like any sea level should be a standard time series

Yeah. I have updated accordingly. 

> believe it should simply point to two entries in a sea-level time series (the "start" and "end" of the delta).

Makes sense. Only thing is that I am not sure how this should be represented in the ontology. As the time series is actually store in PostgreSQL, and OntoTimeSeries does not have any concept that represents an entry in the time series.

sm453:
If sea level is represented as time series, does the `measuredIn` relationship it still make sense? Is this not already captured by the time series?

sm453:
> > believe it should simply point to two entries in a sea-level time series (the "start" and "end" of the delta).
> 
> Makes sense. Only thing is that I am not sure how this should be represented in the ontology. As the time series is actually stored in PostgreSQL, and OntoTimeSeries does not have any concept that represents an entry in the time series.

Use time stamps (with which you can then look up in the time series)?

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
```diff
@@ -28,7 +26,7 @@ Monument,Class,https://raw.githubusercontent.com/beAWARE-project/ontology/master
...
 SharedSocioEconomicPathway,Class,Scenario,IS-A,,,,A climate change scenario of projected socioeconomic changes.,https://www.theworldavatar.com/kg/ontosealevel/,Shared Socioeconomic Pathway
 measuredAt,Object Property,,,LocalMeanSeaLevel,http://www.opengis.net/ont/geosparql#Feature,exactly 1,A relationship between the local mean sea level and the reference point for the local mean sea level measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured at
-measuredIn,Data Property,,,SeaLevelMeasurement,http://www.w3.org/2001/XMLSchema#gYear UNION http://www.w3.org/2001/XMLSchema#date,,A relationship between a sea level measurement and the time of measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured in
+measuredIn,Data Property,,,SeaLevelMeasurement,http://www.w3.org/2001/XMLSchema#gYear UNION http://www.w3.org/2001/XMLSchema#date UNION http://www.w3.org/2001/XMLSchema#dateTime UNION http://www.w3.org/2006/time#Instant,,A relationship between a sea level measurement and the time of measurement.,https://www.theworldavatar.com/kg/ontosealevel/,measured in
```
sm453, line 29: Would you be able to explain in a readme file the design/intended use of this relationship, given that a sea level measurement is a time series (and hence the apparent redundancy)?

sm453 (changes requested):

JPS_Ontology/ontology/ontosealevel/OntoSeaLevel.csv
YiKaiTsai: My initial thinking was that it could be the case that we only have one datapoint for a forecast or actual measurement. And that having this relationship, it would be easier to query for cases with one datapoint.

I've given more thoughts to it. Yeah, this is redundant, as a single data point can be instantiated as a time series with 1 entry.

sm453 (approved):

PR #1210, opened 2024-05-16
TWA-ViP: Customisation Enhancements

qhouyee:
Features:
- Embedding of dashboard using the "dashboard" parameter in settings.json
- Multiple data.json can be ingested using the new data-setting.json
- Improve dark mode design for some features
- Addition of "grouping" parameter within data.json to allow alternate views of the same source layer
- Addition of "hover" parameter within data.json to create hovering effect for the specified layer if enabled
- Added feature selector to manage multiple closely-positioned features
- Bug fixes


jlai2021:
How not create secret folder to store the user name and api key? instead of putting them in the ./

jlai2021 (approved):
Reproduceable

PR #1219, opened 2024-05-22
Hotfix-stack-clients-spring-nested-jars

gpeb2:
Updated the FileUtils::listFilesFromJar method so that it still works when the URL points to a directory inside a nested .jar file. This occurs when building a Spring Boot based project as .jar file rather than .war file.


Ushcode (approved):
Using a filesystem `fs` was the same solution I found when I looked up other people with this problem.

Tested and working with data upload in nested jar files

PR #1221, opened 2024-05-23
Dev scheduler fix

kok-foong:
Key changes:
1) Workaround for an issue encountered in digital ocean where the thread that sends the HTTP request to execute AERMOD disappears when the simulation takes long. "Solved" by destroying the thread after waiting for a predefined time.
2) Added an API to execute a simulation over a predefined time
3) Potentially reduced ram usage in PythonService by closing figures
4) Added null values for timesteps that do not have raster data, should not be an issue anymore with the fix with the updated stack-clients


JPS_VIRTUALSENSOR/VirtualSensorAgent/src/main/java/com/cmclinnovations/virtualsensor/QueryClient.java
```diff
@@ -271,9 +271,8 @@ public void updateStationUsingDispersionRaster(Instant latestTime, Map<String, S
                         double concentration = result.getDouble("val");
                         concentrations.add(concentration);
                     } else {
-                        String errmsg = "Could not obtain raster value";
-                        LOGGER.error(errmsg);
-                        throw new RuntimeException(errmsg);
+                        LOGGER.warn("Could not obtain raster value for {}", rasterFileName);
```
ChungTingLao, line 274: What is the effect of this changed? Instead of non-existent it returns null?

kok-foong: The concentration value at that time point will be null if raster data cannot be retrieved

ChungTingLao (approved):

PR #1228, opened 2024-05-30
Crystal Agent

paul-cares:
This is a subset of the zeolite agent, which covers only the crystal properties. 
I tried to fix all known problems and comments that were found during discussion of the ontozeolite.
Also added XRD spectrum computation and saving to ABox.


sm453 (changes requested):
Changing to `https` in TWA prefixes is certainly the correct thing to do, as a matter of principle. However, @Kondinski tells me that this is inconsistent with currently instantiated data, and that this doesn't help with ongoing projects. Being practical, I would suggest reverting to `http` in this pull request, so that we can complete the ongoing projects in a timely fashion. Please also add a "to do" item in the readme file so that we don't forget to fix this when the time is right.

paul-cares:
This is a different agent. 
The current is CrystalAgent, and it is not in the Zeolite manuscript. This is a stand-alone code.

sm453 (approved):
Thanks for the clarification. Approving this one, then.

PR #1233, opened 2024-06-05
Fix issue with inserting JSON-LD into Blazegraph

gpeb2:
Fixes this issue with inserting JSON-LD files https://github.com/blazegraph/database/issues/170 by replacing the .jar file for the Apache Commons IO library in the Blazegraph image.


qhouyee (approved):
Tested the upload of JSON-LD data into blazegraph, which is working now

PR #1237, opened 2024-06-10
1814 dev update osm agent for sg

YiKaiTsai:


Agents/OSMAgent/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN sed -i "s|REPO_USERNAME|$(cat ../credentials/repo_username.txt)|;s|REPO_PASS
 
 # Build
 WORKDIR /root/osmagent
-RUN --mount=type=cache,target=/root/.m2/repository mvn package
+RUN --mount=type=cache,target=/root/.m2/repository mvn package -DskipTests
```
ShinZert, line 17: The -DskipTests might want to be removed after the results is checked and deemed ok

ShinZert (approved):

Agents/OSMAgent/Dockerfile
YiKaiTsai: This is to skip unit tests, as we are not updating the unit tests in this pull request to save time. Unit tests will be updated in the future and then -DskipTests will be removed.

PR #1239, opened 2024-06-11
1230-py4jps-wrapper-allows-multiple-jvms-to-be-initialised

### LINKED ISSUE #1230: py4jps wrapper allows multiple JVMs to be initialised
Each call to ```jpsBaseLibGW.launchGateway``` starts a new Java Virtual Machine (JVM) and if you create a Java object in one and then try to pass it to the other then bad things will probably happen.
The solution is to make this function create a singleton object that is initialised once and then just returned on subsequent calls.

jb2197:
This PR adopts a singleton pattern for `JPSGateway`. It closes #1230.


gpeb2 (approved):
This looks good.

PR #1242, opened 2024-06-12
Dev 1236 trig stack log4j error and Ontoisochrone.owl

ShinZert:
Fix log4j error and add iso:hasFloodDepth triple to OntoIsochrone.owl


Ushcode (commented):
- Version increments for agents LGTM
- new ontology property LGTM

- The version numbers for log4j are already included in the latest parent pom (these agents depend on a version from about a year ago) so can be removed from the agent poms if parent is updated. The version of parent pom dependency could be updated as well `2.2.0` to `2.3.0`

I'll leave it up to you if you want to update the overall dependency or leave as is (the old `2.2.0` pom has a different version of `log4j`)

Ushcode (approved):

Ushcode (approved):

PR #1274, opened 2024-07-09
Dev derivation ogm

jb2197:
This PR integrates derived information framework with object graph mapper within `twa` Python package, which should make `pyderivationagent` Python package obsolete. The stable release will be made once the PR is approved.


sm453 (approved):
Looks ok, but have not tested.

PR #1291, opened 2024-07-30
Run link checker on Markdown files

### LINKED ISSUE #464: Run link checker on Markdown files
The Markdown files used to document use of the various pieces of code in the TWA often have both internal and external links in them that are easy to mistype. It would be useful to ensure they get checked before merging a branch to "main". A GitHub action such as [my-markdown-link-checker](https://github.com/marketplace/actions/my-markdown-link-checker) could be appropriate for this.

mayomatsuda:
Add github workflow to check markdown links for broken links before pushing.

Resolves #464

### DIFF
--- a/.github/workflows/twa-md-push.yml
+++ b/.github/workflows/twa-md-push.yml
@@ -1,0 +1,23 @@
+# 
+# This workflow contains a job to check for broken links within Markdown files in the repository.
+# 
+name: TWA Markdown Push
+
+# Trigger this workflow during pushes to the 'main' branch if changes to Markdown files
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**.md'
+
+jobs:
+  # Check for broken links within Markdown files
+  markdown-link-check:
+    name: Check markdown files for broken links
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


gpeb2 (changes requested):
Thank you for this contribution and apologies for not reviewing it sooner.
Whilst this pull request currently adds the Markdown link checker GitHub action, as suggested in the issue, merging this as it is will block any future pull requests until all of the broken links that the action finds are fixed.

To get a list of the broken links it is possible to  run the action locally using the [act tool](https://nektosact.com/introduction.html). Fixing them then requires either putting in an exclusion pattern/rule (i.e. for example URLs), or finding the new/correct link and replacing it.

.github/workflows/twa-md-push.json
```diff
@@ -0,0 +1,65 @@
+{
+    "ignorePatterns": [
```
mayomatsuda, line 2: certain links prevent the workflow from working even though they're fine, e.g. dropbox. I've added such to this list. The workflow is buggy with header references (#) and mailto so ignoring those as well.

and ignoring /user, /explore and /images/defaults because it's part of the website code

PR #1304, opened 2024-08-08
Add PID Controller to ontodevice

Abyssus1010:
Add PID Controller to ontodevice


JPS_Ontology/ontology/ontoassetmanagement/OntoAssetManagement.owl
sm453, line 1: I'm guessing in commit efd21c9a89c664217149074b055da89afa1e8834, you forgot to also revert changes to this file?

sm453 (changes requested):

JPS_Ontology/ontology/ontoassetmanagement/OntoAssetManagement.owl
sm453: I can see that. However, it would be better practice to revert _all_ changes to OntoAssetManagement, because this PR is only concerned with OntoDevice.

Abyssus1010: Ah okay. Understood. I have reverted OntoAssetManagement to the same one on the main branch.

sm453 (approved):

PR #1321, opened 2024-08-22
OntoService

qhouyee:
The development of OntoService serves to record the occurrence of each service and describe the terms, conditions, and obligations associated with service delivery. It is designed to be a minimal modular ontology that is reusable for different projects, requiring further extensions to the specific domain.


gpeb2 (approved):
This seems fine

PR #1322, opened 2024-08-22
Dev time series benchmark

ChungTingLao:
Create Benchmarking module for JPS base lib time series client using Java Microbenchmark Harness (JMH).


JPS_BASE_LIB_BENCHMARK/pom.xml
```diff
@@ -0,0 +1,205 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>uk.ac.cam.cares.jps</groupId>
+    <artifactId>jps-base-lib-benchmark</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <name>jps-base-lib-benchmark</name>
+
```
gpeb2, line 42: You should probably set the parent pom as the parent to improve the likelihood of the same versions of dependencies being used in "production" and the benchmarks.

gpeb2 (changes requested):

gpeb2 (approved):

PR #1324, opened 2024-08-26
1318 update feature info agent to for reduced time series tables

kok-foong:
Update FIA to use TimeSeriesClientFactory to generate the correct TimeSeriesClient.  Added a second method in TimeSeriesClientFactory for federated query option


kok-foong:
Before merging, the jps base lib needs to be deployed first. After that, the pom file for FIA needs to be updated with the right base lib version

cmclscourtney (approved):

PR #1329, opened 2024-08-28
Dev virtual sensor ship timeseries

kok-foong:
Updated all components in the dispersion stack to use the new time series client with reduced tables


kok-foong:
https://github.com/cambridge-cares/TheWorldAvatar/pull/1324 should be merged first

JPS_VIRTUALSENSOR/ShipInputAgent/src/main/java/com/cmclinnovations/ship/QueryClient.java
```diff
@@ -416,31 +417,22 @@ void setMeasureIri(List<Ship> ships) {
 
     void updateTimeSeriesData(List<Ship> ships) {
         try (Connection conn = remoteRDBStoreClient.getConnection()) {
-            ships.stream().forEach(ship -> {
+            ships.stream().filter(ship -> ship.hasTimeSeries() && !ship.getTimestampList().isEmpty()).forEach(ship -> {
```
ChungTingLao, line 420: This change breaks the case where ship data is provided by static JSON file. Please change the logic to accommodate both static JSON and live data stream cases.

ChungTingLao (changes requested):
Logic in query client of ship input agent is currently broken for static JSON case based on my testing. Please change the logic to accommodate both static JSON and live data stream cases.

JPS_VIRTUALSENSOR/ShipInputAgent/src/main/java/com/cmclinnovations/ship/QueryClient.java
kok-foong: the relevant commit is reverted, should be fixed

ChungTingLao (approved):
I have tested with ship data from JSON files and CSV files and they both work as intended.

PR #1333, opened 2024-08-30
Add Ontologies and Rules to Datasets

cmclscourtney:
Added support for adding
- Ontologies that are added to Blazegraph and Ontop
-  Ontop Rules files

@kok-foong: This also fixes issue with querying the Ontop endpoint before any mappings have been added.


gpeb2 (approved):
Looks good and doesn't break the existing functionality.

PR #1344, opened 2024-09-11
1834 update sensorloggeragent for compressed data

katrinashonka:
includes updated data flow for sending compressed data to the server and updated permissions


Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -225,46 +252,76 @@ private void onRecordButtonClicked(View view) {
...
+
     private void initPermissions() {
         permissionsMap.put(Permission.PermissionType.LOCATION_FINE, new Permission(Permission.PermissionType.LOCATION_FINE, requestFineLocationPermissionLauncher));
         permissionsMap.put(Permission.PermissionType.AUDIO, new Permission(Permission.PermissionType.AUDIO, requestAudioPermissionLauncher));
         permissionsMap.put(Permission.PermissionType.NOTIFICATION, new Permission(Permission.PermissionType.NOTIFICATION, requestNotificationPermissionLauncher));
     }
 
 
-    private void checkFineLocationPermission(Permission permission) {
+    private void checkFineLocationPermission(Permission permission, Runnable onPermissionGranted) {
```
sandradeng20, line 279: rename this method. It supports the check of audio and location permission

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -127,72 +138,88 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
-        }
-        binding.startRecordTv.setEnabled(hasToggledOnSensor);
-    }
 
     /**
      * Handles the logic for starting or stopping sensor recording when the button is clicked.
      *
      * @param view The view that was clicked.
      */
     private void onRecordButtonClicked(View view) {
```
sandradeng20, line 185: The state management for this function is quite confusing. Please update this method
1. Probably OK: Gather all the selectedSensors from adapter, but will be more ideal to just use selectedSensors in ViewModel to make it the only source of true of state
    - selectedSensors is updated whenever a sensor is toggled
2. OK: Check whether location and audio have been toggled and request for the corresponding permission
3. Not Good: manually call startRecording(...). It should change `isRecording` state only and listen to this state to perform the corresponding UI updates and viewmodel function calls. And selectedSensors should not be passed into ViewModel, it should be set into ViewModel, and the ViewModel is the only source of true of state.
4. Not Good: Should not set listener and onClickeListener in this method unless they response differently every time the button is clicked. Instead they should be defined in onViewCreated

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -127,72 +138,88 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
 
-        // Gather selected sensor types
-        List<SensorItem> selectedSensorTypes = new ArrayList<>();
-        for (SensorItem item : adapter.getSensorItems()) {
-            if (item.isToggled()) {
-                selectedSensorTypes.add(item);
+        Runnable startRecordingRunnable = () -> startRecording(selectedSensorTypes);
+
+        if (locationToggled) {
+            Permission permission = permissionsMap.get(Permission.PermissionType.LOCATION_FINE);
```
sandradeng20, line 212: Since now we are checking the permission only when the button has been toggled, there is no need to use `permissionMap` any more. Can just create the permission instance when needed

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -225,46 +252,76 @@ private void onRecordButtonClicked(View view) {
...
             permission.isGranted = true;
+            onPermissionGranted.run();
         } else if (ActivityCompat.shouldShowRequestPermissionRationale(requireActivity(), permission.permissionString)) {
             new MaterialAlertDialogBuilder(requireContext())
                     .setMessage(permission.explanation)
                     .setPositiveButton(uk.ac.cam.cares.jps.ui.R.string.ok, (dialogInterface, i) -> {
                         // some permissions (eg. notification) may be auto granted in lower sdk version
                         if (permission.permissionString.isEmpty()) {
                             permission.isGranted = true;
+                            onPermissionGranted.run();
```
sandradeng20, line 290: this might not be needed here, cause the `ActivityResultLauncher` defined when creating the `Permission` object should already handle it

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -225,46 +252,76 @@ private void onRecordButtonClicked(View view) {
...
                             return;
                         }
+                        pendingPermissionCallback = onPermissionGranted;
```
sandradeng20, line 293: The idea of using a global runnable object here is a bit weird. If the action for all permissions are the same, should just call fragment member function directly in the `ActivityResultLauncher` instead of create an global object and set it to null after finishing using, which makes the workflow confusing and have the risk of null pointer exception

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -279,7 +336,9 @@ private void checkFineLocationPermission(Permission permission) {
      */
     @Override
     public void onSensorToggle() {
```
sandradeng20, line 338: If not override, can just remove this

Apps/TimelineApp/gradle.properties
```diff
@@ -18,4 +18,5 @@ android.useAndroidX=true
 # Enables namespacing of each library's R class so that its R class includes only the
 # resources declared in the library itself and none from the library's dependencies,
 # thereby reducing the size of the R class for that library
-android.nonTransitiveRClass=true
\ No newline at end of file
+android.nonTransitiveRClass=true
+MAPBOX_DOWNLOADS_TOKEN=sk.eyJ1Ijoia3MyMTgyIiwiYSI6ImNsenJzNjRxcjI0c3QyanFyaDF5a21kZ2gifQ.wTr6vSBv9TEbmT1R3oaMH
```
sandradeng20, line 22: this is the secret key and should not be exposed. Please rotate your secret key, and set it in your PC's local gradle.properties. `settings.gradle` line24 already handle it. 

https://docs.mapbox.com/android/maps/guides/install/#configure-your-secret-token

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/SensorService.java
```diff
@@ -95,59 +110,173 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
-                type
-        );
 
-        // run the sendData task on a separate thread
-        handler.post(sendData);
+
+
+        //  buffer and flush process
+        handler.post(bufferAndFlushData);
+        handler.post(sendDataToNetwork);
```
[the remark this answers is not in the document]
katrinashonka: Room is able to handle concurrency properly so it is thread safe, but I can add transactions if necessary.

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -1,11 +1,14 @@
 package uk.ac.cam.cares.jps.user;
 
+import static uk.ac.cam.cares.jps.user.LoginFragment.LOGGER;
```
sandradeng20, line 3: should not import the LOGGER from LoginFragment, create one for this class instead

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -90,196 +91,236 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+     * the recording state. Updates the UI and RecyclerView based on the observed data.
+     */
+    @SuppressLint("NotifyDataSetChanged")
+    private void setupObservers() {
         sensorViewModel.getSensorItems().observe(getViewLifecycleOwner(), sensorItems -> {
             adapter.updateSensorItems(sensorItems);
             adapter.notifyDataSetChanged();
         });
+        sensorViewModel.getAllToggledOn().observe(getViewLifecycleOwner(), this::updateToggleAllState);
+        sensorViewModel.getIsRecording().observe(getViewLifecycleOwner(), this::updateRecordingUI);
```
sandradeng20, line 135: the control of start/stop recording is still weird. Please recheck logics in `updateRecordingUI()`, `startRecording()`, `stopRecording()` and `SensorViewModel.stopRecording()` and  `SensorViewModel.startRecording()`.

- It seems that `startRecording()` in fragment manages both start and stop recording logic and also some of the UI component control. It overlap with `stopRecording()` and `updateRecordingUI()`
- A more ideal way will be when click start/stop button, the fragment only update the `isRecording` state in ViewModel and the rest is done with this observer, which require modification in SensorViewModel as well

sandradeng20 (approved):

PR #1357, opened 2024-10-01
1842-dev-add-concepts-to-ontolab: add lab equipment concepts to ontolab

Abyssus1010:


jb2197 (approved):
The CSV file looks okay. As discussed, several concepts are present in OntoCAPE but are defined at the scale of chemical plants, therefore they are not reused.

PR #1367, opened 2024-10-07
Visualisation Platform Map Feature Search

qhouyee:
Added the complete map feature search capability, to metadata and time series
Added geocoding UI


Ushcode (approved):
filtering works as expected when testing

PR #1369, opened 2024-10-08
Fix logging in the network analysis agent

gpeb2:
The network analysis agent doesn't always work due to clashes in the logging framework it uses. This PR aims to fix these issues.


ChungTingLao (approved):
Network analysis agent can now accepts HTTP request and run. While it is time consuming to run, it is outside the scope of this pull request.

ShinZert (approved):

PR #1370, opened 2024-10-09
Dev trajectory fia

kok-foong:
Updated FIA to display trajectory information. Descriptions are in the README.


Agents/FeatureInfoAgent/sample/fia/feature_query.sparql
```diff
@@ -0,0 +1,26 @@
...
+               disp:hasTime ?time;
+               disp:hasPollutant/rdf:type <https://www.theworldavatar.com/kg/ontodispersion/NOx>;
+               geo:asWKT ?geom.
+
+      # Subquery to find the maximum time value
+      {
+        SELECT (MAX(?maxTime) AS ?time)
+        WHERE {
+          ?polygon a disp:DispersionPolygon;
+                   disp:hasTime ?maxTime.
```
cmclscourtney, line 19: Would this be needed in case there is another `disp:DispersionPolygon` with a different `disp:hasPollutant/rdf:type` with a later time.
```suggestion
          ?polygon a disp:DispersionPolygon;
                   disp:hasTime ?time;
                   disp:hasPollutant/rdf:type <https://www.theworldavatar.com/kg/ontodispersion/NOx>;
                   geo:asWKT ?geom.
```

Agents/FeatureInfoAgent/sample/fia/trajectory_meta.sparql
```diff
@@ -0,0 +1,12 @@
...
+SELECT ?Property (AVG(?value) AS ?Value) ?Unit
+WHERE {
+  SERVICE [ONTOP] {
+    ?Feature disp:hasValue ?value.
+  }
+  BIND("Average NOx concentration" AS ?Property)
+  BIND("&micro;g/m&sup3;" AS ?Unit)
+}
+GROUP BY ?Property ?Unit
+
```
cmclscourtney, line 12: Not overly relevant to this feature but it should be noted that this isn't a correct calculation of the average concentration.

cmclscourtney (approved):
In futrue, it would be best to do tragectory calculations all in a single sparql query. This requires the time series being kg accessible. This is alright for now.

Agents/FeatureInfoAgent/sample/fia/feature_query.sparql
kok-foong: In this case it is fine because all the pollutants share the same timestamps

Agents/FeatureInfoAgent/sample/fia/trajectory_meta.sparql
kok-foong: I am aware it is not accurate as it does not take into account the duration spent in each polygon

kok-foong:
> In futrue, it would be best to do tragectory calculations all in a single sparql query. This requires the time series being kg accessible. This is alright for now.
The main reason behind the current design is to allow flexibility to use SQL queries for the dataset that is not feasible with GeoSPARQL via Ontop (it is too slow)

PR #1377, opened 2024-10-11
1839 record transport mode

katrinashonka:


Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/handler/SoundLevelHandler.java
```diff
@@ -141,7 +159,7 @@ private void logDBFS(double dBFS) {
             dataPoint.put("time", System.currentTimeMillis() * 1000000);
             dataPoint.put("values", values);
 
-            synchronized (this) {
+            synchronized (sensorData) {
```
sandradeng20, line 162: it sounds like not a good idea to use sensorData as the lock, because it is not final.

And this should be a protection set in SensorHandlerManager as well, where it reads the sensorData. Consider to add this synchronization more properly with the other sensor handlers as well

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/handler/SoundLevelHandler.java
```diff
@@ -97,11 +104,22 @@ public void run() {
      */
     private void processAudioStream() {
         short[] buffer = new short[BUFFER_SIZE / 2];
-        while (isRecording) {
-            int readResult = audioRecord.read(buffer, 0, buffer.length);
-            if (readResult > 0) {
-                double dBFS = calculateDBFS(buffer, readResult);
-                logDBFS(dBFS);
+        while (true) {
+            AudioRecord localAudioRecord = audioRecord; // local reference to prevent NullPointerException
```
sandradeng20, line 108: this won't work, they will be on the same instance and addr

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/network/SensorNetworkSource.java
```diff
@@ -92,13 +88,19 @@ public void sendPostRequest(String deviceId, byte[] compressedData, JSONArray se
...
                 Map<String, JSONArray> sensorDataMap = convertSensorDataToMap(sensorData);
 
-                    UnsentData unsentData = new UnsentData();
-                    String serializedData = serializeMap(sensorDataMap);
-                    unsentData.deviceId = deviceId;
-                    unsentData.data = serializedData;
-                    Log.e("network source", "data =" + serializedData);
-                    unsentData.timestamp = System.currentTimeMillis();
-                    sensorLocalSource.insertUnsentData(unsentData);
+                    String dataHash = sensorLocalSource.computeHash(sensorData.toString());
```
sandradeng20, line 91: if computeHash is used by both source, maybe can consider to put it in a util class as static function. It feels weird to put in either of the source class

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/network/SensorNetworkSource.java
```diff
@@ -92,13 +88,19 @@ public void sendPostRequest(String deviceId, byte[] compressedData, JSONArray se
                     // convert it to the correct format for the writeToDatabase method
                 Map<String, JSONArray> sensorDataMap = convertSensorDataToMap(sensorData);
```
sandradeng20, line 89: remove unused code

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/database/SensorLocalSource.java
```diff
@@ -338,7 +374,55 @@ private List<Long> extractTimes(List<? extends SensorData> sensorDataList) {
...
+        }
+    }
+
+    /**
+     * Checks if the given data already exists in the UnsentData table.
+     *
+     * @param sensorData The sensor data as a JSONArray.
+     * @return True if data exists, false otherwise.
+     */
+    public boolean isDataInUnsentData(JSONArray sensorData) {
```
sandradeng20, line 405: can modify the input to take the hash string, because it is hashed in NetworkSource anyway

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/worker/SensorUploadWorker.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Compresses a given string into a GZIP-compressed byte array.
+     * This method uses the GZIP compression algorithm to reduce the size of the input string.
+     *
+     * @param data The input string to be compressed.
+     * @return A byte array containing the GZIP-compressed data.
+     * @throws IOException If an I/O error occurs during the compression process.
+     */
+    public static byte[] compressData(String data) throws IOException {
```
sandradeng20, line 115: this function has also been defined in SensorService. Move it to some util class as static method

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/SensorService.java
```diff
@@ -272,18 +267,18 @@ public void run() {
      * @throws IOException If an I/O error occurs during the compression process.
      */
     public static byte[] compressData(String data) throws IOException {
```
sandradeng20, line 269: this has been defined in SensorUploadWorker, move to some util class as a static method

Apps/TimelineApp/feature/user/src/main/java/uk/ac/cam/cares/jps/user/SensorSettingFragment.java
```diff
@@ -282,5 +339,16 @@ private void requestAudioPermission(Runnable onGranted) {
         }
     });
 
+    /**
+     * Disables EmojiCompat for the given TextView by directly setting a CharSequence
+     * without emoji handling.
+     *
+     * @param textView The TextView to disable EmojiCompat on.
+     */
+    private void disableEmojiCompat(TextView textView) {
```
sandradeng20, line 348: is this still needed?

sandradeng20 (changes requested):

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/SensorService.java
```diff
@@ -317,30 +312,87 @@ private Notification getNotification() {
...
+                return;
+            }
+            activityRecognitionClient.removeActivityUpdates(pendingIntent);
+        }
+
+
+        LOGGER.info("Stopping sensor service 1");
         try {
-            sensorHandlerManager.stopSensors();
+            if (sensorHandlerManager != null) {
```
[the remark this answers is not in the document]
katrinashonka: i added this when i was debugging the thread stuff because it gave me the error that it was unable to stop the service - i can remove it but if the manager is null can it call stopsensors? won't it crash

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/SensorService.java
```diff
@@ -317,30 +312,87 @@ private Notification getNotification() {
...
             unregisterReceiver(networkChangeReceiver);
             networkChangeReceiver = null;
         }
 
-        LOGGER.info("Stopping sensor service");
+        // Stop activity recognition updates
+        if (activityRecognitionClient != null) {
+            PendingIntent pendingIntent = PendingIntent.getService(this, 0,
+                    new Intent(this, ActivityRecognitionService.class), PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE);
+            if (ActivityCompat.checkSelfPermission(this, Manifest.permission.ACTIVITY_RECOGNITION) != PackageManager.PERMISSION_GRANTED) {
```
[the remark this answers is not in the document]
katrinashonka: this does not request permissions, it just checks whether permission has been granted. if not, it returns without making any permission requests - that's handled in fragment. it errors unless i have this check even though there is already a check earlier on in the code

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/worker/BufferFlushWorker.java
```diff
@@ -0,0 +1,87 @@
...
+            for (Map.Entry<String, JSONArray> entry : localData.entrySet()) {
+                String sensorName = entry.getKey();
+                JSONArray newData = entry.getValue();
+
+                if (!memoryBuffer.containsKey(sensorName)) {
+                    memoryBuffer.put(sensorName, new JSONArray());
+                }
+
+                // Add the new data to the existing buffered data
+                JSONArray bufferedData = memoryBuffer.get(sensorName);
```
sandradeng20, line 57: why create local variable for `memoryBuffer.get(sensorName)`? and if want to check whether `bufferedData` is not null, it looks more proper to put outside of the for loop at line 58. 

but you have already done check and initialize the mapping in line52-54, don't think the assert is necessary.

can consider to use Map.computeIfAbsent() which get the value of key or init the key mapping if absent to make the code neat

Apps/TimelineApp/core/sensor/build.gradle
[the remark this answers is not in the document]
sandradeng20: found some dependency on junit test and mokito as well, better to move these to common gradle file

sandradeng20 (changes requested):

sandradeng20:
since the workflow of sensor recording has been changed, need to update the readme accordingly as well

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/source/handler/SoundLevelHandler.java
```diff
@@ -105,13 +106,14 @@ public void run() {
...
-                break;
+            synchronized (audioRecordLock) {
+                if (audioRecord == null || audioRecord.getRecordingState() != AudioRecord.RECORDSTATE_RECORDING) {
+                    break;
+                }
             }
 
             try {
-                int readResult = localAudioRecord.read(buffer, 0, buffer.length);
+                int readResult = audioRecord.read(buffer, 0, buffer.length);
```
sandradeng20, line 116: better to include this in the synchronized block, otherwise it also risks from null pointer exception when audioRecordLock is released but picked up by stop()

Apps/TimelineApp/core/sensor/src/main/java/uk/ac/cam/cares/jps/sensor/SensorService.java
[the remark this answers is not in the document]
katrinashonka: it should have been removed in the most recently pushed version

sandradeng20 (approved):

PR #1381, opened 2024-10-14
Hotfix podman

gpeb2:
This PR fixes the bugthat causes the following error when runing with Podman:
```ERROR com.github.dockerjava.api.async.ResultCallbackTemplate - Error during callback
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: {"cause":"invalid argument","message":"must provide at least one stream to attach to: invalid argument","response":500}
```


sm453 (approved):
Tested. Works.

PR #1384, opened 2024-10-21
1383 add authorization header to getmap requests

kok-foong:
Overview:
- Adds this header to GetMap requests for MapBox - "Authorization: Bearer ${jwtToken}"

New environment variable
- REACT_APP_USE_GEOSERVER_PROXY

On server.js, a new API is added:
- /geoserver-proxy
  - MapBox will be sending requests here when REACT_APP_USE_GEOSERVER_PROXY is true, it adds the JWT token to the header before sending the requests to the specified source
  
In mapbox-container.tsx
- a transformRequest field is added to the Map object, when REACT_APP_USE_GEOSERVER_PROXY is true, mapbox will send the requests to /geoserver-proxy


web/twa-vis-platform/code/src/ui/map/mapbox/mapbox-container.tsx
```diff
@@ -29,67 +29,84 @@ interface MapProperties {
...
+            pitch: props.defaultPosition.pitch,
+            transformRequest: (url) => {
+                if (process.env.REACT_APP_USE_GEOSERVER_PROXY === 'true') {
+                    try {
+                        const urlObject = new URL(url);
+                        const params = new URLSearchParams(urlObject.search);
+                        if (params.get('request') === 'GetMap') {
+                            // not sure if this will work across all conditions
+                            const serverUrl = `${window.location.protocol}//${window.location.host}`;
+                            const proxyUrl = `${serverUrl}/geoserver-proxy?url=${encodeURIComponent(url)}`;
```
Ushcode, line 73: this needs to change as we do not want to restrict viz and geoserver url to the same host as the viz url

web/twa-vis-platform/code/server.js
```diff
@@ -48,75 +49,109 @@ let store;
...
+                        headers: headers,
+                        responseType: 'stream', // To stream the response back
+                    });
+
+                    // Pipe the response back to the client
+                    response.data.pipe(res);
+                } catch (err) {
+                    // most of these errors can probably be ignored
+                    console.error(err);
+                }
```
Ushcode, line 142: I think we should get rid of this whole path and just save the `req.kauth.grant.access_token` to the already existing /api/userinfo/path or somewhere else as a global variable. Then call that directly from the mapbox container url transformer

web/twa-vis-platform/code/src/ui/map/mapbox/mapbox-container.tsx
```diff
@@ -29,67 +29,84 @@ interface MapProperties {
...
+                            return {
+                                url: proxyUrl
+                            }
+                        }
+                    } catch { }
+                } else {
+                    return { url: url }
+                }
+            }
```
Ushcode, line 82:
```suggestion
            transformRequest: (url, resourceType) => {
        if (resourceType === 'Source' && url.startsWith('http://myHost')) {
            return {
                url: url.replace('http', 'https'),
                headers: {'my-custom-header': true},
                credentials: 'include'  // Include cookies for cross-origin requests
            };
        }
    }
```

comes from [here](https://docs.mapbox.com/mapbox-gl-js/api/map/). Let's try just pass the header in this bit of code and use a variable to get the bearer token from /api/userinfo

web/twa-vis-platform/code/src/ui/map/mapbox/mapbox-container.tsx
kok-foong: Will add an additional environment variable for the server url

kok-foong: agree with this

web/twa-vis-platform/code/server.js
kok-foong: The intention behind this design is that the token needs to be taken directly from the session store/server for each request, otherwise the token gets outdated and becomes invalid.

kok-foong: I think this design guarantees that each request is handled by the middleware

web/twa-vis-platform/code/src/ui/map/mapbox/mapbox-container.tsx
kok-foong: Using resourceType does not seem to be robust enough, when I use MapBox vector tile, its value is 'Tile'

web/twa-vis-platform/code/src/ui/map/mapbox/mapbox-container.tsx
kok-foong: addressed by https://github.com/cambridge-cares/TheWorldAvatar/pull/1384/commits/1250573f4f2596ebae7ea3d0953f83696bb369fe

Ushcode (approved):
As discussed on call earlier, happy to merge this since it works well for the trajectories use case as demonstrated by @kok-foong and does not interfere with CReDo or other mapbox applications when the geoserver proxy setting is not  set.

May revisit the application later with an improved resource server that authorises the token directly with keycloak using the spring adapter, or maybe uses a Requesting Party Token / Token exchange @gpeb2

PR #1386, opened 2024-10-22
Dev geoserver proxy

kok-foong:
Created proxy to read the Authorization header and updated timeline stack config to include it


Agents/utils/GeoServerJwtProxy/stack-manager-input/geoserver-jwt-proxy.json
```diff
@@ -0,0 +1,41 @@
+{
+    "ServiceSpec": {
+        "Name": "geoserver-jwt-proxy",
+        "TaskTemplate": {
+            "ContainerSpec": {
+                "Image": "ghcr.io/cambridge-cares/geoserver-jwt-proxy:1.0-SNAPSHOT",
+                "Env": [
+                    "VIEWPARAM_NAME=device_id"
```
sandradeng20, line 8: this one is it better to make it user_id so it is same as the current geoserver setting?

kok-foong: This an example, there is another one in the timeline stack that uses user_id

sandradeng20 (approved):

PR #1388, opened 2024-10-25
Vip Hotfix V5.8.0

qhouyee:
This hotfix fix the issues in production builds


Ushcode (commented):
builds in prod container, ESLint still complaining about something unclear. Will do a proper fix when I figure it out

PR #1407, opened 2024-11-19
Generic rules

cmclscourtney:
Rules node in stack data uploader that runs SPARQL rules in Blazegraph as well as adding to Ontop.


gpeb2 (changes requested):
Looks fine other than updating the readme.

PR #1413, opened 2024-11-28
give scripts execution permissions

Ushcode:


Ushcode:
Several agents and the stack manager unable to run on after building on linux because scripts had no execution permissions in the repo

ChungTingLao (approved):

PR #1420, opened 2024-12-03
1261 nginx failing to restart sometimes

### LINKED ISSUE #586: Stack manager to cleanly remove agents
Currently, adding agents to an already running stack is straightforward: Just copy a config json file into the `inputs/config` folder and start the stack manager. However, removing such agents with `docker rm -f` and removing the config json files leaves (then no longer valid) routing information in the stack nginx container. Any further agents that are subsequently added are not reachable (404). It appears nginx fails at the invalid entry and does not consider any following entries. The easiest work-around is to kill the stack altogether and start it again from scratch.

It would be nice if there was a way to cleanly remove a running agent from a running stack, so that users can dynamically add and remove containers as they see fit.

### LINKED ISSUE #1261: Nginx failing to restart sometimes
Nginx fails to restart with the rest of the containers sometimes. This happens for two reasons, each with a different fix.

1. The containers are auto-restarting and the other containers are not ready before the last time the Nginx container tries to spin up. This should be fixed by adjusting the restart config for Nginx so that it tries for longer, either by specifying more attempts or longer between them. If possible delaying the first attempt might be a good idea.
2. The containers are being manually restarted via the stack-manager. this could be fixed by either removing the exiting Nginx config files when the stack-manager starts or changing the way the configs are set so that Nginx can be started last.

gpeb2:
Resolves #1261 and resolves #586 by modifying the way Nginx resolves container names to IP addresses so that they can change and still be resolved and so that Nginx doesn't fail if one is missing when it starts up.

Also added the ability to customise the restart policy of a service in it's config file and fixed a issue with removing old services when spinning up a stack.


sm453 (approved):
Tested with podman v4.9.4-rhel with netavark backend on Rocky 8:
- Starting a stack on top of a full running stack.
- Starting a stack on top of a running stack with selected containers manually killed.
- Restarting a stack that crashed.
These all work.

Issue #1416, opened 2024-11-28
Restarting stack overwrites changes in fileserver made by HTTP POST


jb2197:
Currently, the fileserver reloads the folder provided during initialisation whenever the stack restarts. This behaviour effectively discards any changes made to the files via `HTTP POST`, as the files are stored internally within the fileserver container. It would be ideal to implement a mechanism to preserve the file contents in the fileserver container across restarts. One potential solution could involve connecting the container's folder to the initialisation folder to make sure they are in sync.

simonrihm:
This can be rectified by changing the mount type of the related docker service from "volume" to "bind". 
When deploying with the stack-manager, the file-server.json would look something like this then:
```json
{
    "ServiceSpec": {
        "Name": "file-server",
        "TaskTemplate": {
            "ContainerSpec":    {
                                "Image": "ghcr.io/cambridge-cares/fileserver:1.1.0",
                                "Mounts":   [
                                                {
                                                "Type": "bind",
                                                "Source": "vis_files",
                                                "Target": "/app/fs_root"
                                                }
                                            ],
                                "Secrets": [
                                                {
                                                "SecretName": "file_server_password"
                                                }
                                            ]
                                }
                        }
                    }, 
    "endpoints":    {
                    "file-server": {
                                    "url": "http://localhost:8080/FileServer/",
                                    "externalPath": "/file-server/"
                                    }
                    }
}
```

I will add a sentence about this to the fileserver README

PR #1425, opened 2024-12-04
Update fileserver README

simonrihm:
Added a sentence to mount types in response to https://github.com/cambridge-cares/TheWorldAvatar/issues/1416

### DIFF
--- a/Deploy/stacks/db/fileserver/README.md
+++ b/Deploy/stacks/db/fileserver/README.md
@@ -70,6 +70,10 @@ In short, you need to add a service configuration file and include the file serv
 }
 ```
 
+### Mount Type
+If mount type is "volume", changes to the fileserver via POST or DELETE requests will not be reflected in the the source directory. HTTP requests will only affect the target directory which is located inside the Docker container. This means, when the container crashes or stack gets restarted, all changes in the meantime will be effectively discarded.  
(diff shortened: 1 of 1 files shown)


Deploy/stacks/db/fileserver/README.md
```diff
@@ -70,6 +70,10 @@ In short, you need to add a service configuration file and include the file serv
 }
 ```
 
+### Mount Type
+If mount type is "volume", changes to the fileserver via POST or DELETE requests will not be reflected in the the source directory. HTTP requests will only affect the target directory which is located inside the Docker container. This means, when the container crashes or stack gets restarted, all changes in the meantime will be effectively discarded.  
+If mount type is "bind" on the other side, any changes to the target folder inside the container (via POST or DELETE requests) will be synced to the source folder outside the container. If the stack is restarted, all changes will still be there.
```
jb2197, line 75:
```suggestion
If mount type is "bind" on the other side, any changes to the target folder inside the container (via POST or DELETE requests) will be synced to the source folder outside the container. If the stack is restarted, all changes will still be there. 

See https://github.com/cambridge-cares/TheWorldAvatar/issues/1416 for more details.
```

jb2197 (approved):

PR #1430, opened 2024-12-06
move code to Ushcode/viz

Ushcode:
This PR also marks as deprecated the old `twa-vis-platform` code and points to its new home


gpeb2:
It looks like one of the GitHub checks needs to be removed. I assume there are other GitHub workflows, etc. that need to be removed.

PR #1433, opened 2024-12-10
Deprecate cambridge-cares/TheWorldAvatar/web/twa-vis-platform following creation of TheWorldAvatar/viz

Ushcode:


gpeb2 (approved):
Everything that should be seems to have been removed and the new readme seems sufficient.

PR #1448, opened 2025-01-21
Fix chown on copy issue

gpeb2:
This PR is to fix an issue where transferring files between containers using the Docker API currently requires a `chown` call with user `1000`, which is rightly disallowed on secure systems.


sm453 (approved):
Tested on VM with podman, and secure platform. Seems to resolve the problem (assuming sub*id files are correctly in place).

gpeb2:
Merged in [new repo](https://github.com/TheWorldAvatar/stack/pull/1).

PR #1461, opened 2025-01-31
1851 add oura ring api key

sandradeng20:
add oura ring api key to twa account mapping


Agents/UserAgent/UserAgent/src/main/java/uk/ac/cam/cares/jps/agent/useragent/UserAgent.java
```diff
@@ -61,57 +77,123 @@ public JSONObject processRequestParameters(JSONObject requestParams, HttpServlet
...
 
-        } else if (request.getRequestURI().contains("getPhoneIds")) {
-            validateInputGetPhoneIds(requestParams);
-            JSONArray phoneIds = kgQueryClient.getPhoneIds(requestParams.getString("userId"));
-            JSONObject result = new JSONObject();
-            result.put("PhoneIds", phoneIds);
-            result.put("Comment", "Phone ids are retrieved.");
-            return result;
+        TokenIntrospectionResponse tokenIntrospectionResponse = authzClient.protection().introspectRequestingPartyToken(response.getToken());
+        return tokenIntrospectionResponse;
```
kok-foong, line 103: what's the purpose of this method? I think the /userinfo endpoint already validates the token?

sandradeng20: This one is to check whether the token has the permission to perform the operations (eg. `registerPhone`, `getPhoneIds`, `registerOuraRing`), although we now allow all users to perform these three operations.

kok-foong (approved):

Agents/UserAgent/UserAgent/src/main/java/uk/ac/cam/cares/jps/agent/useragent/UserAgent.java
kok-foong: This is fine, just wondering if there is a way to only contact KeyCloak once instead of twice

kok-foong (approved):

PR #1465, opened 2025-01-31
1464-handle-devices-without-time-series: updated function to handle d…

kok-foong:
Each device id is checked whether it has time series before using it to generate the layer

### DIFF
--- a/Agents/TrajectoryQueryAgent/docker-compose.yml
+++ b/Agents/TrajectoryQueryAgent/docker-compose.yml
@@ -1,4 +1,4 @@
 services:
   trajectoryqueryagent:
-    image: ghcr.io/cambridge-cares/trajectoryqueryagent:4.1.0
+    image: ghcr.io/cambridge-cares/trajectoryqueryagent:4.1.1
     build: .
--- a/Agents/TrajectoryQueryAgent/trajectoryqueryagent/src/main/resources/functions.sql
+++ b/Agents/TrajectoryQueryAgent/trajectoryqueryagent/src/main/resources/functions.sql
@@ -100,23 +100,30 @@ RETURNS TABLE (
 ) AS $$
 DECLARE
     query TEXT := '';
+    processed_array TEXT[] := ARRAY[]::TEXT[];
 BEGIN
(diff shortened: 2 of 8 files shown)


sandradeng20:
found that function `get_device_iri(text)` is used in line_layer_device_id.sql, but not in functions.sql. This causes error to create layers in new stack

sandradeng20:
/createLayer will through error when `devices` table hasn't been created. This happens in new stack when no data has been uploaded with sensorloggeragent

kok-foong:
> /createLayer will through error when `devices` table hasn't been created. This happens in new stack when no data has been uploaded with sensorloggeragent

This is expected, it's not possible to create the layer without the data present

kok-foong:
> found that function `get_device_iri(text)` is used in line_layer_device_id.sql, but not in functions.sql. This causes error to create layers in new stack

This function is added by the sensor logger agent, it is necessary to instantiate the data with sensor logger agent first for any of the functions to work

sandradeng20:
This may cause the app to show error message with new stack. App workflow after login: 
1. app automatically registers the logged in user with device id -> call user agent to register smartphone
2. user sees map page -> call trajectoryqueryagent to create layer
3. user may choose to record data in setting page -> upload data

With a new stack, user will always see the error message when open up the app because there is no data uploaded yet.

sandradeng20:
`get_location_table(text)` will still have error when all device id of a user not having the corresponding ts_iri and column in `time_series_quantities`

Error in query (7): ERROR: upper bound of FOR loop cannot be null
CONTEXT: PL/pgSQL function get_location_table(text[]) line 12 at FOR with integer loop variable

kok-foong:
> `get_location_table(text)` will still have error when all device id of a user not having the corresponding ts_iri and column in `time_series_quantities`
> 
> Error in query (7): ERROR: upper bound of FOR loop cannot be null CONTEXT: PL/pgSQL function get_location_table(text[]) line 12 at FOR with integer loop variable

This is expected...

sandradeng20 (approved):

PR #1470, opened 2025-02-12
move-stack-code: move stack code and add readme

Ushcode:


gpeb2 (changes requested):
Most of the bash scripts and some of the other files should probably be left here for now so that it's still possible to build old agents and start a stack from this repo.
I will go through an work out what needs to be kept. Also there are changes that need to be made in the new stack repo before it can be used.

gpeb2 (approved):

PR #1473, opened 2025-02-13
move-fia: move and add note

Ushcode:
https://github.com/TheWorldAvatar/Feature-Info-Agent


Agents/FeatureInfoAgent/MOVED.md
sm453, line 1: Shouldn't this file have some content explaining the situation and where to find the current git rep?

sm453 (changes requested):

Agents/FeatureInfoAgent/MOVED.md
Ushcode: Yes it should... It got lost somehow

sm453 (approved):

PR #1474, opened 2025-02-17
move-parent-pom: move it

Ushcode:


gpeb2 (changes requested):
GitHub shows the content of README.md by default so it is probably a good idea to keep the name of the .md file as README.md. The message is fine though.

gpeb2 (approved):

Issue #1496, opened 2025-04-23
Feature environmental exposure calculation agent


sandradeng20:
Develop an environmental exposure calculation agent to setup a modularised and extendable workflow.

```mermaid v2
graph TD
    A[Sample points] --> B[Create buffer]
    B --> C[Access TWA KG for filtered objects information ]
    C --> D[Feed IRIs to Postgis/Ontop as SQL/Sparql query for geometry calculation]
    D --> E[Output]
```

Sample points
- [ ] Regularly sample within the given rectangular area
- [ ] Postal code
- [ ] Files input

Create buffer
- [ ] Circular buffer with given radius

Calculation
- [ ] Point: count intersected points in the buffer
- [ ] Polygon: area intersected percentage

Support dataset
- [ ] SG Facility data
- [ ] Culture site from sea level data
- [ ] Greenspace

Output
- [ ] Store result in database
- [ ] JSON output format

sm453:
Under "Calculation", please add another item "Polygon: count polygons intersecting the buffer".

PR #1497, opened 2025-04-23
1854 environment exposure calculation

sandradeng20:
#1496 Develop an environmental exposure calculation agent to setup a modularised and extendable workflow.


Agents/EnvironmentExposureCalculationAgent/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+# COPY ./resources/ontop.obda ./resources/ontop.obda
+ADD https://jdbc.postgresql.org/download/postgresql-42.5.0.jar ./resources/
+
+# Install Python package and the required libraries
+RUN python -m pip install --upgrade pip
+RUN pip install -e .
+# Installation of agentlogging currently suspended due to performance issues
+# Install StackClients twa resource
+RUN stack_clients_jar=$(find ../tmp_stack/stack-clients*.jar) && stack_clients_jar=${stack_clients_jar##*/} && \
+    jpsrm install StackClients ../tmp_stack --jar $stack_clients_jar
```
kok-foong, line 33: I found that it is necessary to run `jpsrm uninstall StackClients` when the a new twa package is downloaded. Not critical but probably better to reduce the number of instances where this line is copied over, though redundant once a StackClients package is released properly

PR #1510, opened 2025-06-03
Kok foong/1507/trip agent

kok-foong:
Documentation available in the README

### DIFF
--- a/Agents/TripAgent/README.md
+++ b/Agents/TripAgent/README.md
@@ -40,6 +40,8 @@ Time | Location
 3 | POINT(5,6)
 4 | POINT(7,8)
 
+The raw data must contain at least a visit in order for the trip detection module to work (staying within an area over a certain period).
+
 The agent will check for the existence of trip and visit, if they do not exist, they will be instantiated and share the same time series with the point time series. Existing triples before calculations:
 
 ```sparql
--- a/Agents/TripAgent/README.md
+++ b/Agents/TripAgent/README.md
@@ -16,7 +16,7 @@ API route:
        - This IRI should contain a time series of PostGIS points and should contain the following triple
 
         ```sparql
-        <iri> <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>
+        <http://abc> <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>
         ```
 
     2) upperbound
@@ -43,10 +43,10 @@ Time | Location
 The agent will check for the existence of trip and visit, if they do not exist, they will be instantiated and share the same time series with the point time series. Existing triples before calculations:
 
 ```sparql
-<iri> <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>
+<http://abc> <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>
 ```
 
-New triples:
+New triples showing trip and visit sharing the same time series:
 
 ```sparql
 PREFIX twa: <https://www.theworldavatar.com/kg/>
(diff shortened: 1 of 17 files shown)


Agents/TripAgent/agent/trip/kg_client.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        query_results = self.remote_store_client.executeQuery(query)
+        return query_results.getJSONObject(0).getString('time_series')
+
+    def instantiate_trip_and_visit(self):
+        trip = PREFIX + 'trip/' + str(uuid.uuid4())
+        visit = PREFIX + 'visit/' + str(uuid.uuid4())
+        query = f"""
+        INSERT {{<{trip}> a <{TRIP}>. <{visit}> a <{VISIT}>}}
+        WHERE {{}}
```
sm453, line 54: Use `INSERT DATA` instead of an empty `WHERE` clause.

Agents/TripAgent/docker-compose-debug.yml
```diff
@@ -0,0 +1,7 @@
+services:
+  trip-agent:
+    image: trip-agent-debug:1.0
```
sm453, line 3: I believe the standard convention for image tags would suggest "1.0.0", and I don't see any reason not to follow this here. Please fix. This applies to all the docker compose files.

sm453 (changes requested):
This agent should not be added in this git repository. Please create a new git repository just for this agent in the https://github.com/TheWorldAvatar organisation.

Agents/TripAgent/README.md
```diff
@@ -0,0 +1,103 @@
...
+<http://abc> <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>
+```
+
+New triples showing trip and visit sharing the same time series:
+
+```sparql
+PREFIX twa: <https://www.theworldavatar.com/kg/>
+
+<trip_iri> rdf:type twa:Trip; <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>.
+<visit_iri> rdf:type twa:Visit; <https://www.theworldavatar.com/kg/ontotimeseries/hasTimeSeries> <time_series>.
```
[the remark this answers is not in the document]
kok-foong: The input to this agent is the lowest level point instance, and the point instance might be a part of a trajectory instance, e.g.
```sparql
<trajectory> <hasPoint> <point>. <point> <hasTimeSeries> <time_series>.
```

In the ideal case, the agent should probably attach the visit and trip instances to the trajectory instance,
```sparql
<trajectory> <hasPoint> <point>; <hasTrip> <trip>. 
<point> <hasTimeSeries> <time_series>.
<trip> <hasTimeSeries> <time_series>.
```

But the agent does not know about the higher level trajectory instance.

To query for trips and visit for a particular trajectory, the idea is to query the trip and visit instances that share the same time series as the point instance.

Then to get the underlying data (trip and visit indices), `<trip_iri>` and `<visit_iri>` are the inputs to the TimeSeriesClient.

Agents/TripAgent/agent/trip/kg_client.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        self.remote_store_client.executeUpdate(query)
+        return trip, visit
```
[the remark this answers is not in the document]
kok-foong: I just thought of recording all the results from the trip module as is to know whether it is a trip or a visit in case they are needed later

Agents/TripAgent/docker-compose-debug.yml
kok-foong: Will address in new repo

kok-foong:
New PR - https://github.com/TheWorldAvatar/trip-agent/pull/2

PR #1547, opened 2026-03-16
Update README to reflect project migration

Ushcode:
Removed key features and documentation sections from README.


sm453 (approved):
I'm happy with these changes.

At least as important, though, is to add documentation equivalent to what's been removed here to the new organisation/repos. The organisation does not even have a readme right now.

